PR #181, opened 2012-01-23
Add ddescribe, iit for exclusive suites,specs

vojtajina:
When any ddescribe is registered, only specs withing these exclusive suites will be run

``` javascript
describe('normal', function() {
  ddescribe('exclusive', function() {
    // all specs here will be run
  });

  // nothing here will run
  it('should not run', function() {});
});
```

When any iit is registered, only these exclusive specs will be run (precedence over ddescribe)

``` javascript
it('should not run', function() {});
iit('should run', function() {});
it('should not run 2', function() {});
iit('should run as well', function() {});
```

We've been using this feature for quit a while during development of [AngularJS](http://angularjs.org).

It's part of [Jasmine - JsTD adapter](https://github.com/mhevery/jasmine-jstd-adapter), I think it was originally an idea of @IgorMinar.

I really miss this feature when using [jasmine-node](https://github.com/mhevery/jasmine-node) or when using [SlimJim](https://github.com/vojtajina/slim-jim).

I think, it really should be part of the core, so I did implement it.

### DIFF
--- a/spec/core/RunnerSpec.js
+++ b/spec/core/RunnerSpec.js
@@ -264,4 +264,62 @@ describe('RunnerTest', function() {
       expect(suiteNames(suites)).toEqual([suite1.getFullName(), suite3.getFullName()]);
     });
   });
+
+  describe('ddescribe', function() {
+    it('should run only ddescribe suites when at least one registered', function() {
+      var normal = jasmine.createSpy('normal spec');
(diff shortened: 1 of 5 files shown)


travisperson:
For anyone that found this thread and looking for if this was implemented:

https://github.com/jasmine/jasmine/pull/309#issuecomment-56873934

PR #272, opened 2012-08-17
Custom error message of Matcher

dsuket:
Usage:

```
var failMessage = 'obj.selected is not true. obj.id = ' + obj.id;
expect(obj.selected, failMessage).toBe(true)
```

By default, when obj.selected is false, "Expected false to be true" is printed. if you needs more information when test fail, that can show custom message.
I know get same effect to add message function to matcher object. but that is prolixity.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -516,8 +516,8 @@ if (isCommonJS) exports.xit = xit;
  *
  * @param {Object} actual Actual value to test against and expected value
  */
-var expect = function(actual) {
-  return jasmine.getEnv().currentSpec.expect(actual);
+var expect = function(actual, message, notMessage) {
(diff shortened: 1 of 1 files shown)


gcbw:
work around for anyone that reaches here and despair. or to make an argument for this feature... cuts both ways :)

right out of our code base, enjoy:

```javascript
// generator pattern just so we can have a meaningful error message
// on failures. curse karma-jasmine for not having a reason arg on expect()!
var goodUrls = [ '...' ]; // huge list.
for( url in goodUrls ){
    url = goodUrls[url];
    it('test good url: "'+url+'"', function (url) {
        return function(){
            expect( myClass.isValid(url) ).to.be.true;
        }
    }(url));
}
// now repeat the exact same block for `badUrls` but expecting .to.be.false;
```

HolgerJeromin:
https://github.com/avrelian/jasmine2-custom-message seems to be a solution for that

PR #292, opened 2012-10-30
Support to run specs in parts

acroca:
This feature is usefull to run jasmine in parallel.


brandonros:
Has any kind of parallel/concurrent support been added to any part of Jasmine?

slackersoft:
At this point in time, Jasmine's solution for running in parallel, is to use a [custom boot](https://jasmine.github.io/edge/custom_boot.html) that knows the algorithm to determine which spec ids are run where and pass those ids to the `env.execute()` method.

PR #370, opened 2013-05-23
support for lazy-evaluated variables

kawamanza:
Hi,

I know there were other attempts, such as pull-request #101. I believe this new suggestion is closer to the expected implementation. Please consider the specifications to validate the suggested uses.

Example:

``` javascript
describe('first level tests', function() {
  given('foo', function() { return "bar"; });
  given('subject', function() { return {"foo": given.foo}; });

  it('should call nested lazy-evaluated variables', function() {
    expect(given.subject.foo).toBe(given.foo);
    expect(given.foo).toBe("bar");
  });

  describe('second level tests', function() {
    given('foo', function() { return "rab"; });

    it('should change the value of foo', function() {
      expect(given.subject.foo).toBe(given.foo);
      expect(given.foo).toBe("rab");
    });
  });

});
```

This mechanism is similar to RSpec "let" helper ([here some advantages of this feature](http://www.verious.com/qa/when-to-use-rspec-let/)).

### DIFF
--- a/spec/core/GivenSpec.js
+++ b/spec/core/GivenSpec.js
@@ -11,13 +11,6 @@ describe("Given", function() {
       expect(lazy.values).not.toEqual( {} );
       expect(lazy.values).toEqual( {"myLazyValue":["a"],"testValue":"a"} );
     });
-
-    it("should cache values", function() {
-      var cachedValue = given.myLazyValue;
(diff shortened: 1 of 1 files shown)


maboesanman:
Shameless plug: I've also created a minimal implementation of this behavior called [givens](https://github.com/enova/givens)

PR #455, opened 2013-10-25
Mock date object when installing mock clock.

### LINKED ISSUE #361: jasmine.Clock.useMock doesn't mock Date
It would be useful if `jasmine.Clock.useMock` mocked `Date` such that `jasmine.Clock.tick` incremented the times returned by `Date` after `jasmine.Clock.useMock` is called.  This would make it compatible with `_.throttle`, the implementation of which changed in documentcloud/underscore#820.

jalopez:
Closes #361

This pull adds a new object called MockDate with mimics a date object and can be used to run synchronously tests that depends on current time. It is integrated with mock clock and can be used jointly with setTimeout and setInterval functions (if they use the Date object, it will be the mocked one).

To use it, the spec must install the mock clock but passing an argument (either a date or true). This argument represents the time reference in which tests starts. If its the boolean true, current time will be taken as reference.

Usage:

```
    jasmine.clock.install(true); 
   var now = new Date().getTime()
   jasmine.clock.tick(100);
   expect(Date.now() - now).toEqual(100);
```

### DIFF
--- a/spec/core/ClockSpec.js
+++ b/spec/core/ClockSpec.js
@@ -361,4 +361,73 @@ describe("Clock (acceptance)", function() {
       clock.clearInterval(123)
     }).not.toThrow();
   });
+
+  it("does not mock the Date object when installing without parameters", function() {
+     var delayedFunctionScheduler = new j$.DelayedFunctionScheduler(),
+      global = {Date: Date},
+      mockDate = new j$.MockDate(global),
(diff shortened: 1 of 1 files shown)


fiznool:
Just want to say thanks a bunch for including this. Mocking `new Date()` was always a pain until I found out about this! :purple_heart:

Issue #480, opened 2013-12-21
Data Driven Testing


PeterJCLaw:
Following the principles of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), it would be great if jasmine offered support for writing tests in a data-driven manner.

This is particularly useful for things like validation helpers, where you want to run a battery of different values against it, and assert that they're either valid or not.

It appears that this has already been [blogged about](http://blog.jphpsf.com/2012/08/30/drying-up-your-javascript-jasmine-tests), and even [implemented](https://github.com/jphpsf/jasmine-data-provider) it, but it would be nice to see this in the core.

PR #527, opened 2014-02-22
Added Data Driven Tests

gburghardt:
Resolves Issue #480

Clean up your repetive Jasmine specs with data driven tests using an easy to read
syntax:

```
all("your base are belong to us", [ [2,3], [1,4] ], function(your, base) {
    var longToUs = 5;

    expect(your + base).toBe(longToUs);
});
```

Specify single arguments or multiple arguments:

```
all("specs can have one argument", [1, 2, 3, 4], function(x) {
    expect(x > 0).toBe(true);
});

all("specs can have multiple arguments", [ [1,2], [3,4] ], function(x, y) {
    expect(y > x).toBe(true);
});
```

Under the hood, data driven tests are expanded into multiple `it`'s, which are visible
in the Spec Runner:

```
all("numbers are greater than zero", [ [2,1], [4,3] ], function(a, b) {
    expect(a - b > 0).toBe(true);
});
```

Becomes:

```
numbers are greater than zero (Variant #0 <2, 1>)
numbers are greater than zero (Variant #1 <4, 3>)
```

Since each argument group in the dataset becomes its own spec, you can easily run a
single variant by clicking on the spec name in the browser based spec runner.

Data driven tests can be marked as pending:

```
xall("these are pending", [1, 2, 3], function(x) { });
```

The `all` and `xall` functions return the array of specs that were created:

```
var specs = all("foo", [1, 2], function(x) { ... });
```

Asynchronous data driven tests are executed in the order they are declared, just
like regular async specs defined using `it`:

```
all("repetitive async tests are supported", [1, 2, 3, 4], function(x, done) {
    setTimeout(function() {
        expect(x > 0).toBe(true);

        done();
    }, 500);
});
```

Jasmine will assume your data driven test is asynchronous and provide a `done` callback if
your test function defines `n + 1` arguments, where `n` is the number of arguments in each
dataset argument group. So beware of:

```
all("foo", [1, 2], function(x, y) {
    // "y" is a callback function!
});
```

The dataset is validated to ensure your test function doesn't exhibit wonky behavior, which
should prevent some hair pulling and head banging.

Consider the following test definition:

```
all("test description", [ [1,2], [5] ], function(x, y) {
    ...
});
```

The second argument group only has one argument. When the test is created, this error will
be thrown:

```
Jasmine.ArgumentCountMismatchError: Expected 2 argument(s). Found 1 at index 1 (test description)
```

Now consider:

```
all("test description", [ [1,2], [3,4] ], function(a, b, c, d) {
    ...
});
```

The dataset has two arguments in each group, but the test function itself is expecting 4
arguments. When the test is created, this error will be thrown:

```
Jasmine.ArgumentCountMismatchError: Expecting data driven spec to accept 2 arguments, but 4 arguments are specified in the dataset at index 0 (test description)
```

If you give an invalid or empty dataset:

```
all("test description", [], function(x) {
    ...
});
```

This error is thrown:

```
Jasmine.ArgumentsMissingError: No arguments for a data-driven test were provided (test description)
```

### DIFF
--- a/lib/jasmine-core/boot.js
+++ b/lib/jasmine-core/boot.js
@@ -71,6 +71,14 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
       return env.xit(desc, func);
     },
 
+    all: function(desc, values, func) {
+      return env.all(desc, values, func);
+    },
+
+    xall: function(desc, values, func) {
+      return env.xall(desc, values, func);
+    },
+
     beforeEach: function(beforeEachFunction) {
       return env.beforeEach(beforeEachFunction);
     },
--- a/lib/jasmine-core/boot/boot.js
+++ b/lib/jasmine-core/boot/boot.js
@@ -49,6 +49,14 @@
       return env.xit(desc, func);
     },
 
+    all: function(desc, values, func) {
+      return env.all(desc, values, func);
+    },
+
+    xall: function(desc, values, func) {
+      return env.xall(desc, values, func);
+    },
+
     beforeEach: function(beforeEachFunction) {
       return env.beforeEach(beforeEachFunction);
     },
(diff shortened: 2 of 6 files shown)


gscarva4:
What about jasmine 3? Is there any option?

PR #696, opened 2014-10-24
feat(Promise support): Support using Promises in async tests.

briandipalma:
If a Promise is returned from a test the test result will depend on the
settled state of the returned Promise.

If the Promise fulfills the test will pass if the Promise rejects
the test will fail.

This will allow easy integration of ES7 async functions into Jasmine.


DouglasMeyer:
@slackersoft Was it the quantity of divergent code/added dependencies that makes this a bit too much to support? Or the feature itself? I think this could be an amazing opportunity for test readability.

Consider specs that would look like this:

    it('shows dialog after animation', async function(){
      dialog.show();
      await waitForEl(dialog, { visible: true }); // <- fn returns a promise
      expect(dialog.visible());
    });

 Crummy example, but you know what tests with nested functions looks like :smiley:

briandipalma:
@DouglasMeyer This is the PR https://github.com/jasmine/jasmine/pull/696/files as you can see if was a couple of lines. I'd suggest taking a look at Mocha as converting between the two is straightforward as long as you kept Jasmine specific code out of your tests.

slackersoft:
@DouglasMeyer It is the divergent code and dependencies in addition to the fact that you should be able to do a promise base test with the `done` callback, without any changes to jasmine. Something like:

````javascript
it('promises', function(done) {
  aThingThatPromises().then(done, done.fail);
});
````

DouglasMeyer:
@briandipalma Thanks for the pointer.
@slackersoft I was hoping for tests that look more like an ES7 async function, but thanks.

I'll have to look into Mocha again (it's been a long time).

briandipalma:
@DouglasMeyer I wrote the kind of tests you are talking about using my fork of Jasmine (before moving to Mocha), i.e. https://github.com/briandipalma/br-analyzer/blob/master/spec/bladeset-spec.js#L20

PR #736, opened 2015-01-08
Link to the Jasmine NPM module

moonmaster9000:
Also, put the jasmine-<language> links above the confusing "install jasmine on your local box" stuff (why would I want to do that?)


infews:
While you're in here, can you change the targets to http://github.com/jasmine from http://github.com/pivotal?

slackersoft:
Merged with aeae141fd42928d587b433003d962bbfa1831474

PR #743, opened 2015-01-13
 Done handling still isn't quite right

### LINKED ISSUE #709: Done handling still isn't quite right
I'm on version 2.0.2 

Think there were some pretty big changes to this from 2.0 and it's made it loads better but I'm getting the following problem:

if you have a setup like this:

```
beforeEach(function(done){
    console.log('start');
    doSomethingThatWillCall(done);
    doSomeOtherStuff();
    console.log('end of beforeEach');
});

it('is a test', function(){
    console.log('in test');
});
```

If doSomethingThatWillCall calls done immediately (rather than in an asynchronous manner) then you end up dropping into the test before the subsequent call is done.  Eg you'd get the following in the console:

'start'
'in test'
'end of beforeEach'

Im not sure where the fix would go in the jasmine code but I'd suggest you could do something like this when the before each is run

```
function functionThatSortsItOut(beforeEachFunction, existingDoneFunction) {
    var endReached = false;
    var triedToCallDoneEarly = false;

    var patchedDone = function () {
        triedToCallDoneEarly = true;

        if(!endReached)
            return;

        existingDoneFunction();
    };

    beforeEachFunction(patchedDone);
    endReached = true;

    if(triedToCallDoneEarly)
        existingDoneFunction();
}
```

It sort of feels like that could be simpler but it's nearly home time and i'm tired!

That way it calls done only if the end of the function has already been hit - either when it gets run or later when done gets called asynchronously.

---

As a bit of an aside nice one with including beforeAll and afterAll... although I've not tried them out yet (I've patched in my own version of beforeAll already).  I really like this library - in particular nesting suites is a really great feature - lets you save repeating setup code and bits of test names + something that puts it ahead of similar frameworks in other languages.

JonnyRa:
This fixes #709

### DIFF
--- a/src/core/QueueRunner.js
+++ b/src/core/QueueRunner.js
@@ -108,7 +108,7 @@ getJasmineRequireObj().QueueRunner = function(j$) {
       }
 
       try {
-        queueableFn.fn.call(self.userContext, next);
+        ensureSynchronousCodeCompletes(queueableFn.fn.bind(self.userContext), next);
       } catch (e) {
         handleException(e, queueableFn);
         next();
--- a/src/core/QueueRunner.js
+++ b/src/core/QueueRunner.js
@@ -53,6 +53,36 @@ getJasmineRequireObj().QueueRunner = function(j$) {
         handleException(e, queueableFn);
       }
     }
+    
+    function ensureSynchronousCodeCompletes(beforeEachFunction, existingDoneFunction) {
+      var endReached = false;
+      var triedToCallDoneEarly = false;
+
+      var patchedDone = function () {
+          //this tells you if done has been called before the end of the beforeEach has been hit
+          triedToCallDoneEarly = true;
+
+          //if we haven't dropped off the end of the before each method then don't call done yet
+          //code further down will do that for us instead at the appropriate point
+          if(!endReached)
+              return;
+
+          //if the end has already been reached then call the done function to move to the next test
+          existingDoneFunction();
+      };
+
+      //kick off the before each method with our more cautious version of done
+      beforeEachFunction(patchedDone);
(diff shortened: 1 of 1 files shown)


slackersoft:
By calling `done` you are telling jasmine that you are ready to proceed to the next step in your spec. If you are doing it before then, jasmine doesn't want to wait until you're done. If anything I'd rather see it throw and error and fail your suite in this situation.

JonnyRa:
I can see you're not convinced.  I fixed it in my local copy because this situation caught me out yet again and it was half an hour or so of debugging to figure out what was going on.

I suppose throwing an exception would stop the mystery but it would need some pretty wierd wrapping code to then call done correctly to avoid the exception that you'd be making your users write if they are in the 'sometimes this call needs to do something asynchronous' situation.

eg:
    beforeEach(function (done) {
        var wrappedDone = function () {
            window.setTimeout(done, 10);
        };
    });

That seems worse than just making it work seamlessly.  The execution of done when run using the patch I suggested effectively behaves like the call to done is asynchronous, so it's not like its a very surprising way for some javascript to behave.

It would actually be possible to patch it using the setTimeout thing above aswell (which I hadn't realised til now!) but I feel the patch I before suggested is clearer + wont introduce (albeit minor) delays into the calls.

It's up to you - haven't really got any more to add, I think I've tried to pitch this to you enough times now and feel like you understand what the problem is.   It'd be a shame if you didn't fix it though, I'm a big fan of the library, think this is my only issue with it - apart from that it's great, really like the flexibility that comes from everything just being a function call.

infews:
@JonnyRa Jasmine is working as expected. We don't see this as a bug.

The pattern of using ` done()` this way came from Mocha - it's expected and a simple way to provide asynchronous support.

Thanks for using Jasmine.

PR #763, opened 2015-02-02
Fix missing comma on bower.json

gabrielhpugliese:
I've just installed it in my project and suddenly stopped working. Now I think it's fixed.

PR #764, opened 2015-02-02
Fix invalid bower.json

joshuacc:
This just broke Bower installation on a project of mine. :frowning: 

Maybe adding a JSON lint step to the build would be help prevent issues like this in the future?


gabrielpugliese-luizalabs:
:+1: I've opened #763 too :(

slackersoft:
Closing, merged #763 with the same fix.

PR #767, opened 2015-02-03
Allow arrays from different frames or contexts to be equal

juliemr:


src/core/matchers/matchersUtil.js
```diff
@@ -167,11 +167,19 @@ getJasmineRequireObj().matchersUtil = function(j$) {
...
       // Objects with different constructors are not equivalent, but `Object`s
-      // from different frames are.
-      var aCtor = a.constructor, bCtor = b.constructor;
-      if (aCtor !== bCtor && !(isFunction(aCtor) && (aCtor instanceof aCtor) &&
-        isFunction(bCtor) && (bCtor instanceof bCtor))) {
-        return false;
+      // or `Array`s from different frames are.
+      var areArrays = className === '[object Array]';
+      if (!areArrays) {
+        if (typeof a != 'object' || typeof b != 'object') {
```
Gerg: This line appears to be a duplicate of [line 126](https://github.com/jasmine/jasmine/pull/767/files#diff-4b1cbd917e62d27b2645d46dd0d01be3L126)

src/core/matchers/matchersUtil.js
```diff
@@ -167,11 +167,19 @@ getJasmineRequireObj().matchersUtil = function(j$) {
...
+          return false;
+        }
+
+        var aCtor = a.constructor, bCtor = b.constructor;
+        if (aCtor !== bCtor && !(isFunction(aCtor) && aCtor instanceof aCtor &&
+               isFunction(bCtor) && bCtor instanceof bCtor) &&
+               ('constructor' in a && 'constructor' in b)) {
```
Gerg: What case is this intended to catch? This appears unrelated to arrays.

Gerg:
Thanks for the PR.

A couple issues:
1. I commented on a couple lines with specific questions.
2. This PR does not include any tests. As per our [contribution guide](https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md), we require tests, so that there are not future regressions.

For the test, it would be ideal if we could reproduce this issue without relying on Node. If this is not possible, you can add node-specific tests to the [npmPackage directory](https://github.com/jasmine/jasmine/tree/master/spec/npmPackage).

src/core/matchers/matchersUtil.js
juliemr: Not sure, it was present in the underscore `isEqual` function (https://github.com/jashkenas/underscore/blob/95337a876784d5df7e62b2e55080ddf9199b6e9b/underscore.js#L1127). Tests pass without it, so I'll removed it.

src/core/matchers/matchersUtil.js
juliemr: removed.

juliemr:
Thanks for taking a look!

The two ways that I know to reproduce the issue are
1) in node with `vm` (won't work in browser)
2) using multiple frames (won't work in node)

More context: http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/

I think my preference would be to add tests to `matchersUtilSpec.js` which only run if `require` is available, much like there are now functions which only run if `document` is available. Does that sound reasonable? If so, I'll go ahead and do that.

slackersoft:
That sounds good. I like having the specs be closer to where you would expect to find them for equality.

juliemr:
OK - I've added tests for both the iframes case and the vm case. The iframes case requires appending an element to the DOM, which I clean up at the end of the test - but let me know if you don't like this.

spec/core/matchers/matchersUtilSpec.js
```diff
@@ -171,6 +171,63 @@ describe("matchersUtil", function() {
...
+
+    it("passes for equivalent objects in different frames", function() {
+      if (typeof document === 'undefined') {
+        return;
+      }
+
+      var frame = document.createElement("iframe");
+      document.body.appendChild(frame);
+      xObject = window.frames[window.frames.length-1].Object;
+      var obj = new Object();
```
slackersoft: I think this should say `new xObject()` instead of just `Object`

slackersoft:
Given that the fix is the same, let's just go with the `vm`-based specs and leave the frame based ones out.

juliemr:
OK, removed the iframe based tests.

PR #770, opened 2015-02-05
Started simple cleanup of Env

mjgpy3:


slackersoft:
Thanks for taking a shot at this. We're not quite ready to start refactoring `Env` right now, although it does need some work. We need to do some more lower level refactoring in the rest of jasmine before we're ready to start working on `Env`.

We'll keep these in mind when we get there. Thanks for using jasmine.

PR #772, opened 2015-02-07
Error for incorrect spy return value

mjgpy3:
When testing I frequently find myself accidentally writing:
```javascript
... jasmine.createSpyObj('...', [...]).and.returnValue(...);
```

This errors out (as one might expect) but the message is not very clear if the spy is far away from the actual expectation (as it usually is, being setup in `beforeEach` hooks):
> TypeError: Cannot call method 'returnValue' of undefined

Maybe this is just an bad habit I find myself in, but I thought it might be nice to get more clear feedback as to why I typed the wrong thing.

Let me know if this is valuable, I would love to discuss it further.


slackersoft:
I don't think this is something we really want in core. It causes some tight coupling between `createSpyObj` and `SpyStrategy` by needing the `InvalidSpyStrategy` to know all the methods that exist.

You should probably be able to write a wrapper around `createSpyObj` that does the same thing if you find yourself in this situation a lot.

Thanks for using jasmine. Closing.

Issue #774, opened 2015-02-09
Specs inside xdescribe are not reported at all


Derbeth:
See http://jasmine.github.io/2.2/introduction.html: xdescribe "A spec" contains a spec "is just a function, so it can contain any code". This nested spec is completely ignored in Jasmine results:

1. nested "it" is counted in the total number of tests, but appears neither in executed nor in skipped tests, so the numbers don't sum up and cause confusion ("Ran 80 of 81 specs - run all / 80 specs, 0 failures, 3 pending specs" - so, are there 80 or 81 specs?)
2. you cannot find the "it" text nowhere in the report, only the text of "xdescribe" can be seen

slackersoft:
I think it probably makes sense that `it`s in an `xdescribe` should be reported the same as if they were all `xit`s. We'd be happy to review a pull request that implemented this.

PR #776, opened 2015-02-11
Report skipped specs for disabled suites

seanparmelee:
My attempt at addressing #774 as I too would like to be able to report which specs are skipped as the result of an `xdescribe`.

### DIFF
--- a/spec/core/SuiteSpec.js
+++ b/spec/core/SuiteSpec.js
@@ -145,7 +145,7 @@ describe("Suite", function() {
 
   it("can be disabled, but still calls callbacks", function() {
     var env = new j$.Env(),
-      fakeQueueRunner = jasmine.createSpy('fake queue runner'),
+      fakeQueueRunner = function(attrs) { attrs.onComplete(); },
       onStart = jasmine.createSpy('onStart'),
       resultCallback = jasmine.createSpy('resultCallback'),
       onComplete = jasmine.createSpy('onComplete'),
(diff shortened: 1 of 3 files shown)


slackersoft:
I think the goal of #774 is to have `it`s inside an `xdescribe` be reported the same way as an `xit` would. This would also mean that no changes are needed in reporter to surface the skipped `it`s. 

Currently, this pull request requires changes to all reporters that wish to present the `describe`s and `it`s that were not run. Additionally, it looks like only direct children of the `xdescribe` will be reported, so this wouldn't report the `it`:

````javascript
xdescribe('foo', function() {
  describe('bar', function() {
    it('skipped', function() {
      expect(1).toBe(2);
    });
  });
});
````

seanparmelee:
Thanks for the feedback @slackersoft.

I had made the assumption that we'd want this functionality to be "opt-in", but it sounds like that's not the case so I'll go ahead and rework this PR. I'll also address the issue with nested `it`s, thanks for pointing that out.

seanparmelee:
Just pushed another take at this. Now `xdescribe` will call `disable` on all its children being continuing execution. This allows `it`s to be reported when one of their ancestor Suites is disabled.

slackersoft:
This looks pretty good. Unfortunately, we're not going to pull it in because we ended up doing a very similar thing when pulling out the processing/running of the spec tree. Once we pull that code into master your code won't necessarily apply cleanly. We've just finished some cleanup of the TreeProcessor code and will be pulling it into master sometime today. If you have a chance to take a look, let us know if the new code solves your problem.

Thanks for using jasmine!

seanparmelee:
:+1: Works great, thanks!

PR #777, opened 2015-02-12
make jasmine Timer use performance.now() if browser supports it - for higher resolution timing

Malkiz:


infews:
This request does not have tests and it does not modify the correct files. Please follow the Contributor's guide for all Jasmine pull requests. 

Can you point us to `performance.now()`? We don't know enough about this API to consider whether or how this should be integrated into Jasmine core.

Thanks for using Jasmine!

Malkiz:
@infews It is a browser feature available in modern browsers: https://developer.mozilla.org/en-US/docs/Web/API/Performance.now

slackersoft:
We haven't heard any other need for this from the community, and there are a number of issues with the pull request as it stands right now. See the list from @infews above, as well as the fact that this doesn't work in node (see the travis build).

I'm going to close this for now, but please open a new pull request if you still want this feature and can fix these issues.

Thanks for using jasmine!

Issue #529, opened 2014-02-22
Fail with error & stack trace in asynchronous code


stevecj:
When an error is thrown from asynchronous code, Jasmine currently has no way of knowing that it happened, skipping subsequent stages (e.g. "it" following error in "beforeEach") and reporting a trace.

As a workaround, I have written helper code to catch and record the error from a function execution so that it can be re-thrown from an "afterEach" block. This solution at least gives a failure and a stack trace, but still leaves a lot to be desired.

For my partial workaround, see the "asyncStep" and "getCaptureAsyncError" functions in https://github.com/stevecj/headway/blob/master/spec/support/userContextExt.js . To see how those are used, see https://github.com/stevecj/headway/blob/master/spec/indexedDbAdapterSpec.js .

slackersoft:
We have this tracker story (https://www.pivotaltracker.com/story/show/66359332) as well

mik01aj:
I don't know if this is the same or different issue, but I discovered that when I do this somewhere:

```
setTimeout(function () {
    throw 'kaboom';
}, 1);
```

then what I get is:

```
PhantomJS 1.9.8 (Mac OS X) ERROR
  kaboom
  at undefined

Finished in 0.002 secs / 0.014 secs

SUMMARY:
✔ 16 tests completed
```

So the reporter thinks that my tests pass... but it actually ran only 16 out of 52 tests I currently have. :scream:

slackersoft:
The fact that the exception does not actually cause the suite to fail is the problem here. I'm not sure why it didn't run all of your tests, and am actually unable to reproduce that issue.

mik01aj:
One more similar issue: when I do:

        setTimeout(function() {
            expect(true).toBe(false);
        }, 10);

...it fails some random async spec later.

slackersoft:
@mik01aj that actually doesn't surprise me. Once you get into `setTimeout` there isn't really any way to trace the failure back to any particular spec. So, if your spec doesn't have a `done` callback that it invokes at the end of the `setTimeout` function, jasmine will just associate the expectation with whichever spec is actually being executed when the timeout elapses and the function is invoked.

That is probably not something we're going to worry about, since it's failing the suite and you've basically set it up so jasmine can't figure out where the correct place to put the failure is.

adrianolek:
The issue is when a spec does have a `done` callback and the async code throws some exception before executing `done()`.
Something like this:

```js
describe('suite', function () {
  it('spec 1', function (done) {
[...]

PR #784, opened 2015-02-18
Support `jasmine.getEnv().done()` to mark an async test as done outside of the current test lexical scope

mattheworiordan:
There are numerous examples where an external bit of Javascript knows the test has failed and wants the test to fail immediately.  The `jasmine.getEnv().fail()` command is useful, however it does not cause an async test to stop so you have to wait for it to timeout.  This is incredibly inefficient for large test suites.

This fork is now being used locally and it has had my test suite far less verbose because my `fail()` handler called from anywhere can instruct the currently running async methods to be marked as done.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -450,6 +450,7 @@ getJasmineRequireObj().Env = function(j$) {
     var runnableResources = {};
 
     var currentSpec = null;
+    var currentQueueRunner = null;
     var currentlyExecutingSuites = [];
     var currentDeclarationSuite = null;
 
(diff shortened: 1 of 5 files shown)


slackersoft:
It seems strange to me that some part of your code that knows it is running a spec, doesn't have access to the `done` callback for the spec being run. If you have the `done` callback, `done.fail()` is the provided way to cause an async spec to  fail and continue immediately.

Hope this helps. Thanks for using jasmine.

mattheworiordan:
Hi @slackersoft, sorry for the slow reply.  In the end, we've reverted back to Nodeunit in fact because it was just too painful to try and migrate everything to Jasmine because of these issues.

I can give you two cases when an external bit of code needs to call the `fail()` on global in an async test.

* Uncaught exceptions caught with Node's `process.on('uncaughtException')` or browser `window.onerror`.
* Because we were migrating a huge test suite from Nodeunit, it made sense to retain the same test methods so we were using the [Chai test assertions](http://chaijs.com/).  Chai raises exceptions when an assert fails, however that would not work with Jasmine because an exception in a test block may not raise the exception in the test's lexical scope i.e. the assert is in a callback.  So we needed some way to "plug" into Jasmine and fail the current test.

slackersoft:
It sounds like what you really need then is #529 so that exceptions in asynchronous code are properly associated with the spec in which they occurred, and we're looking to get that in for the next release of jasmine (2.3).

I'm going to close this issue as a duplicate of #529.

PR #810, opened 2015-03-25
Contextual Jasmine Functions

lbaker2:
Dear Maintainers,

Thanks for an awesome library that has been easy to use and convince my rspec familiar coworkers to start writing spec tests for javascript. Thinking of my coworkers, I continued to write some contextual functions for Jasmine Env and Jasmine Interface. The functions will decrease redundancy and improve readability of jasmine tests. I believe the community would enjoy implementing these functions as my coworkers have found our specs much easier to write and understand amongst each other. The only drawback the contextual functions has is the naming of functions '_with' and '_without', which I prefixed with underscores due to the Javascript 'with' keyword (I prefixed without to match with). I don't think the underscore prefix is a huge deal, but if someone can come up with another name or prefix I wouldn't be offended.

Included are spec tests and core code.

Thanks,
Lawrence Baker

PS - I've started working on Subject and Let functions to improve writing and maintaining spec tests, though I might have to rename 'let' to something else for ES6. Thanks again.

### DIFF
--- a/lib/jasmine-core/example/node_example/lib/jasmine_examples/Person.js
+++ b/lib/jasmine-core/example/node_example/lib/jasmine_examples/Person.js
@@ -35,8 +35,4 @@ Person.prototype.removeSongs = function(){
 	this.songs = [];
 };
 
-Person.prototype.isHappy = function() {
-  return (this.player && this.songs.length);
-};
-
 module.exports = Person;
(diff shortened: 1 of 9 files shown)


infews:
Thanks for all of this _very_ detailed work. And will be useful to some projects.

But this is a lot of top-level change to Jasmine core. We prefer to keep Jasmine very simple - not everyone is ready for/requires/wants this type of interface. So we don't want to take this into core.

We'd love to support and promote this as an add-on to Jasmine for those users. Have you considered publishing this independently?

Closing.

Issue #769, opened 2015-02-04
v2.2 jasmine.objectContaining doesn't work with Object.defineProperty


logankd:
I'm using TypeScript to create classes. The properties are added with Object.defineProperty. When I update from Jasmine 2.0 to 2.2 my test started failing. This same test passes in v2.0.

Here's my test:
```javascript
var myObject = (function () {
			function myObject(id){
				this.a = ['a','b','c'];
				this.empty = [];
				this._id = id;
				this.update = function(differentT){
					// pretend this is a real function to spyOn
					
				}
			};
			return myObject;
		 })();
		Object.defineProperty(myObject.prototype, "id", {
            get: function () {
                return this._id;
            },
            enumerable: true,
            configurable: true
        });
describe("Object containing and toHaveBeenCalledWith", function() {
		
	it('objectContains matcher with property', function(){
		var t2 = new myObject(444);
		spyOn(test, 'update');
		test.update(t2);
		expect(test.update).toHaveBeenCalledWith(jasmine.objectContaining({ id: 444 }));
	});
});
```
Is there a workaround or can the equality be improved?

```Object.prototype.hasOwnProperty.call(other, property)``` is false in the expect above

Extra Note: I had about 7 broken tests (of 900) after updating.
* expect(myObject.array).toEqual([]) from Knockout observable arrays using the ES5 plugin. had to be changed to myObject.array.length).toBe(0)
*  or comparing arrays values to splitting it to:
expect(acceptedTypes[0]).toBe(accepts[0]);			
expect(acceptedTypes[1]).toBe(accepts[1]);			
expect(acceptedTypes[2]).toBe(accepts[2]);

I'm not sure how I feel about those changes breaking my tests.

Gerg:
It appears that this test is now failing because the property is defined on the object's prototype. Either way, they should probably still match, so we will work on a fix. 

For the array equality expectations, do the arrays have properties defined on them? Jasmine 2.2 compares array properties when checking equality. There was an [issue](https://github.com/jasmine/jasmine/issues/766) with printing array equality failure messages in 2.2, which should be fixed on master. 

It might make sense for there to be a matcher that only compares indexed array values, though I'm not sure exactly what that would look like. Would you find this useful?

logankd:
Here is more information and some tests to duplicate the array issues I'm seeing.
The Chrome debugger shows it as Array[0] with a lot of extra functions. model.saleList instanceof Array is true.
[...]

PR #817, opened 2015-04-05
An order-independent Array matcher

jhamon:
I was reading the discussion between @slackersoft, @logankd and others on #769 and became interested in a matcher that doesn't care about the order of items in an Array.  This type of matcher would be useful any time a person is dealing with an array holding a set-like collection of data.  Since JavaScript doesn't have a native set implementation, I find myself in this situation somewhat often.

Right now we have a few options for testing an Array being used as a set of items.  First, we can assert equality with the entire array.  This makes the spec brittle to implementation changes.  It also means we have to be thinking about initialization state while writing a spec about the side effects of a different method. In this example, "can load passengers" is trying to assert that Luke and Obi-Wan were loaded properly. We don't really care about the other people who were already on board.

``` javascript
describe("MilleniumFalcon", function() {
  beforeEach(function() {
    var ship = new MilleniumFalcon();
    expect(ship.peopleOnBoard()).toEqual(["Han", "Chewie"]);
  });

  it("can load passengers", function() {
    ship.loadPassenger("Luke");
    ship.loadPassenger("Obi-Wan");

    // This test needs to know the order of the people array, and will break
    // if that order changes in the future
    expect(ship.peopleOnBoard()).toEqual(["Han", "Chewie", "Luke", "Obi-Wan"]);
  });
});
```

To express this idea in a less brittle way, right now we need to write a long series of `toContain` expectations.  This works but is a lot of typing.

``` javascript
// How you might write the test today if you don't care about order
describe("MilleniumFalcon", function() {
  beforeEach(function() {
    var ship = new MilleniumFalcon();
    expect(ship.peopleOnBoard()).toContain("Han");
    expect(ship.peopleOnBoard()).toContain("Chewie");
  });

  it("can load passengers", function() {
    ship.loadPassenger("Luke");
    ship.loadPassenger("Obi-Wan");

    expect(ship.peopleOnBoard()).toContain("Luke");
    expect(ship.peopleOnBoard()).toContain("Obi-Wan");
  });
});

```

I would prefer to write something like this, with an order-independent array matcher:

``` javascript
describe("MilleniumFalcon", function() {
  beforeEach(function() {
    var ship = new MilleniumFalcon();
    expect(ship.peopleOnBoard()).toContainAll("Han", "Chewie");
  });

  it("can load passengers", function() {
    ship.loadPassenger("Luke");
    ship.loadPassenger("Obi-Wan");

    expect(ship.peopleOnBoard()).toContainAll("Luke", "Obi-Wan");

    // Some other valid statements might be:
    expect(ship.peopleOnBoard()).toContainAll("Luke", "Chewie", "Obi-Wan", "Han");
    expect(ship.peopleOnBoard()).toContainAll("Luke", "Chewie");
    expect(ship.peopleOnBoard()).toContainAll("Obi-Wan", "Luke");
  });
});
```

One especially concerning thing is that if you had a need like this and guessed that `toContain` accepts multiple arguments you would get **false positives**.  The current implementation ignores additional arguments.  To me, that's a pretty big violation of my expectations.

``` javascript
describe("Misleading toContain", function() {
  it("false positives", function() {
    expect([1,2,3]).toContain(1, 2, 3, 4); // landmine!
  });
});
```
### Specifically looking for feedback on these points

This PR is my work in progress.  I'm sure you'll have feedback, but I'm specifically looking for input on the following:
1. What should this be named? I've been using the working title `toContainAll`, but the more I think about it the more it feels like an extension of the current functionality in `toContain`.  I'd love to be able to write `expect([1,2,3]).toContain(3, 1)` and have it work in the expected way.  
   
   One drawback of changing/extending the behavior of the existing `toContain` matcher is that people working with legacy versions of jasmine may see examples of the new usage, try the new multi-argument syntax in their project, and think it's working for them when really they are getting the false-positive behavior described above.
   
   We might also want to think about how the name we choose might suggest related matchers in the future the way `toHaveBeenCalled` and `toHaveBeenCalledWith` naturally go together.
2. How should duplicates be handled? Keeping track of which items have already been _seen_ makes the implementation more difficult than simply mapping over the actual with `util.contains`, but I think it produces the least surprising results.

``` javascript
describe("duplicate handling", function() {
  it("should not allow double-counting", function() {
    expect([1,1]).toContainAll(1, 1);
    expect([1]).not.toContainAll(1,1);
  });
});
```

### DIFF
--- a/spec/core/matchers/toContainAllSpec.js
+++ b/spec/core/matchers/toContainAllSpec.js
@@ -1,0 +1,107 @@
+describe("toContainAll", function () {
+  it("passes when all items are present", function () {
+    var util = { equals: j$.matchersUtil.equals };
+    var matcher = j$.matchers.toContainAll(util);
+
+    var result = matcher.compare([1, 2, 3], 1, 2, 3);
+    expect(result.pass).toBe(true);
+  });
+
+  it("passes with shuffled arrays", function () {
+    var util = { equals: j$.matchersUtil.equals };
+    var matcher = j$.matchers.toContainAll(util);
+
+    var result = matcher.compare([1, 2, 3], 3, 2, 1);
+    expect(result.pass).toBe(true);
+  });
+
+  it("passes with a valid subset", function () {
+    var util = { equals: j$.matchersUtil.equals };
+    var matcher = j$.matchers.toContainAll(util);
+    var result;
+
+    result = matcher.compare([1, 2, 3], 1, 2, 3);
+    expect(result.pass).toBe(true);
+
+    result = matcher.compare([1, 2, 3], 3, 1);
+    expect(result.pass).toBe(true);
+
+    result = matcher.compare([1, 2, 3], 2);
+    expect(result.pass).toBe(true);
+
+    result = matcher.compare([1, 2, 3]);
+    expect(result.pass).toBe(true);
+  });
+
+  it("handles arrays of mixed types", function () {
+    var util = { equals: j$.matchersUtil.equals };
+    var matcher = j$.matchers.toContainAll(util);
+    var result;
+
+    result = matcher.compare([10, /fuzzy/, "pickles"], /fuzzy/, 10);
+    expect(result.pass).toBe(true);
+
+    result = matcher.compare([1, '2', '3'], '2', '3', 1);
+    expect(result.pass).toBe(true);
+
+    result = matcher.compare([1, {foo: 2}, 3], 3, {foo: 2}, 1);
+    expect(result.pass).toBe(true);
+  });
+
+  it("handles duplicates", function () {
+    var util = { equals: j$.matchersUtil.equals };
+    var matcher = j$.matchers.toContainAll(util);
+    var result;
+
+    result = matcher.compare([1, 2, 3], 3, 3);
+    expect(result.pass).toBe(false);
+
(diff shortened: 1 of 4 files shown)


slackersoft:
If what you want to assert is that all of the items in the expectation are there without regard to any other elements in the array, you should be able to do this now with the `arrayContaining` asymmetric equality tester. This would look like:

`expect([1,2,3]).toEqual(jasmine.arrayContaining([2,1]);`

However, jasmine doesn't currently have a built in way to assert that an array contains exactly a certain set of elements, but in any order.

We could also update `toContain` so it threw an Error when called with more than one argument. The only other matcher that does this currently is `toHaveBeenCalled` to make sure users don't expect it to do argument matching for them, but it's not totally unprecedented.

benchristel:
@jhamon, @slackersoft —Would this feature still be valuable?

If so, I can finish it up. I'm going to be joining @jhamon's Labs project next week, so if this would be useful to her it'll be useful to me!

slackersoft:
I'm thinking now that the most extensible way to do this would be to provide an asymmetric equality tester for it. Something like `jasmine.arrayMatching` would mimic RSpec's `match_array`. This would also allow the matching to be used in spy assertions and other types of tests where Jasmine already support asymmetric equality testers.

PR #828, opened 2015-04-20
Alias toBeFalsy/toBeFalsey to allow for both wording styles.

benlakey:
There are many developers who use the wording 'falsy' and many others who use the wording 'falsey'. 

* http://english.stackexchange.com/questions/109996/is-it-falsy-or-falsey
* https://gist.github.com/jfarmer/2647362

Jasmine provides a 'toBeFalsy' matcher. It's easy to get it wrong and expect 'toBeFalsey' to work as well. This pull is to allow either to be used in jasmine:

    toBeFalsy()
    toBeFalsey()

### DIFF
--- a/spec/core/matchers/toBeFalsySpec.js
+++ b/spec/core/matchers/toBeFalsySpec.js
@@ -1,38 +1,46 @@
 describe("toBeFalsy", function() {
-  it("passes for 'falsy' values", function() {
-    var matcher = j$.matchers.toBeFalsy(),
-      result;
 
-    result = matcher.compare(false);
-    expect(result.pass).toBe(true);
+  function test(matcher) {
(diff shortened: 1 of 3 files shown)


slackersoft:
Our inclination is to keep jasmine itself free of any aliasing. Keeping it so there is only a single way to do things within jasmine keeps jasmine's code and tests clean and easy to read.

I'd be more inclined to look into a feature that enabled a user of jasmine to specify which matchers they would like to be aliased themselves, instead of jasmine providing some aliases out of the box, but we haven't heard much other interest from the community for aliasing, so it doesn't seem like something that would be widely used.

Closing this pull request, as this isn't something we think we want in jasmine itself.

Thanks for using jasmine!

benlakey:
Ok thanks for considering anyway!

PR #836, opened 2015-04-29
Add package.json in order to publish 1.3.x to NPM

jrolfs:


slackersoft:
Jasmine 1.3 isn't really in a position to be published to NPM. Additionally, we are no longer doing new development on 1.3. You are encouraged to [upgrade to jasmine 2.0](https://jasmine.github.io/2.0/upgrading.html) to get new features. Jasmine 2.0 is available on NPM as well.

Closing. Thanks for using jasmine!

jrolfs:
With the few changes I made it seems to work fine when published to NPM. I understand that since 1.3 is essentially deprecated/unsupported that maybe you don't even want it published. My motivation was for a plugin we're working on (using Browserify and NPM) that will aid in migrating large codebases from 1.3 to 2.0.

slackersoft:
I don't think that the bare `jasmine.js` file is exactly the right NPM entry point for jasmine either, it definitely doesn't work the same way that the main entry point for 2.0 does. So I don't think we want the confusion.

PR #851, opened 2015-05-14
Make the HtmlReport CSS classes "unique enough"

### LINKED ISSUE #844: using common CSS class names on Jasmine DOM elements can cause unintended behavior
I'm a relative new-comer in the field of front-end web dev, and there is no other comp sci / programming field in which I'm more experienced. *(But hear me out anyway, ha ha.)*

While working on a student project, I noticed something about Jasmine which you might consider a "bug" which causes unintended DOM interactions between Jasmine's part of the DOM and the app-that's-being-tested's part of the DOM.

And this bug does not have to exist at all; you can painlessly avoid it.

So, what is this problem?

Well, I first noticed a specific case of this bug when I saw some of Jasmine's text appearing in one of the app's divs on a student project I was working on &mdash; the words "Spec List" and "Failures" appeared on top of what should have been in my app's `.menu` div. This occurred whenever a specification failed.

I have reproduced a case of this bug here:
http://codepen.io/prather-mcs/pen/bdVrdN
(this Pen is loading all necessary Jasmine assets from [here](https://cdnjs.com/libraries/jasmine))

When you un-comment the test `expect(false).toBe(true);`, then you will see the bug happen: the app's CSS for `.menu` is affecting the position of Jasmine's DOM elements that also have class `.menu`

I'm sure you want to always keep Jasmine's information down in Jasmine's part of the document.

So that's a specific case where `.menu` is colliding.

But I think the general problem you would want to fix is the creation by Jasmine of DOM elements with such commonly-used class-names as these:

* `.bar`
* `.menu`
* `.results`
* `.summary`
* `.banner`

In summary, the Issue is this:

While none of Jasmine's CSS rules in `jasmine.css` change how the app appears, style rules in the tested app which affect any class name that `jasmine.css` also uses can cause Jasmine's DOM content to be styled in un-intended ways, including causing it to jump right out of Jasmine's part of the page and onto the app's part of the page.

---

I would be happy to get credit for fixing this bug -- I would go to the `jasmine-html.js` code that creates DOM Nodes and ensure that all CSS class names are unique (enough) by giving them slightly new names, such as changing code that would add the class `.menu` to a DOM Node to actually add the class `.jasmine-menu` instead. Then I would make corresponding changes to `jasmine.css` as well.

However, I have not contributed to this project before, and have only been looking into the code-base for the first time this week.

I'm thinking that there are people with seniority and authority who would probably have a better idea of how to name the CSS classes than I have, and they would have a better idea of the best Jasmine app code to write to efficiently implement the fix as well.

With that being said, if you would allow me to write a fix and create a Pull request to fix this bug, I would like to get some credit contributing to open source....

prather-mcs:
As described in Issue Report 844...
https://github.com/jasmine/jasmine/issues/844

...style rules in the app-being-tested may incidentally affect elements
in the Jasmine HTML Report container, as long as there is a chance that
the app-being-tested has CSS style rules for classes (or IDs) that
Jasmine uses.

This fix attempts to bring Jasmine to a state where each and every class
it uses always ends with "_jasmine-css" which should be unique enough to
ensure that CSS in the app-being-tested won't affect the Jasmine report,
because no app-being-tested is ever likely to use classes that end with
"_jasmine-css"

I'll be surpised if this commit is good enough as it is now, on the
first attempt to fix #844, because of reasons I'll explain in either
the Issue or the Pull Request.

### DIFF
--- a/spec/html/HtmlReporterSpec.js
+++ b/spec/html/HtmlReporterSpec.js
@@ -14,20 +14,20 @@ describe("New HtmlReporter", function() {
 
     // Main top-level elements
     expect(container.querySelector("div.jasmine_html-reporter")).toBeTruthy();
-    expect(container.querySelector("div.banner")).toBeTruthy();
-    expect(container.querySelector("div.alert")).toBeTruthy();
-    expect(container.querySelector("div.results")).toBeTruthy();
+    expect(container.querySelector("div.banner_jasmine-css")).toBeTruthy();
+    expect(container.querySelector("div.alert_jasmine-css")).toBeTruthy();
+    expect(container.querySelector("div.results_jasmine-css")).toBeTruthy();
 
-    expect(container.querySelector("ul.symbol-summary")).toBeTruthy();
+    expect(container.querySelector("ul.symbol-summary_jasmine-css")).toBeTruthy();
 
     // title banner
-    var banner = container.querySelector(".banner");
+    var banner = container.querySelector(".banner_jasmine-css");
 
-    var title = banner.querySelector("a.title");
+    var title = banner.querySelector("a.title_jasmine-css");
     expect(title.getAttribute('href')).toEqual('http://jasmine.github.io/');
     expect(title.getAttribute('target')).toEqual('_blank');
 
-    var version = banner.querySelector(".version"),
+    var version = banner.querySelector(".version_jasmine-css"),
       versionText = 'textContent' in version ? version.textContent : version.innerText;
     expect(versionText).toEqual(j$.version);
   });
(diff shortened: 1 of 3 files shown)


slackersoft:
In general, I think this looks pretty good, but I prefer to have `jasmine-` prepended to the previous classes ([as suggested by](https://github.com/jasmine/jasmine/issues/844#issuecomment-100034868) @nertzy). Using a prefix makes it feel like jasmine is the parent in the name instead of the child, and the `css` is redundant as they are all style rules anyways.

If you can fix those up, I think I can merge this pull request. Thanks.

prather-mcs:
*et voila!*

PR #864, opened 2015-05-26
To support filter at suite level.

ammanyer:
Hello, 
My team at Fidelity want to have a filler on the suite level as it already exist for spec level. It is mostly used by protractor and I already did a pull request for the protractor changes.

Thanks,
Ammanuel

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -801,7 +801,7 @@ getJasmineRequireObj().Env = function(j$) {
 
       runnableLookupTable[spec.id] = spec;
 
-      if (!self.specFilter(spec)) {
+      if (!self.specFilter(spec, suite)) {
         spec.disable();
       }
 


infews:
We cannot take this pull request in its current form as it does not adhere to our guidelines.

Notably:

- there are no tests, so we can't be sure that this change does not break existing functionality
- this pull request patches `jasmine.js` directly instead of the component files, so the change will be lost on the next build

Please review the [contributor's guide][1] for how to improve this pull request. Please consider talking to the jasmine-js@googlegroups.com mailing list for additional help.

Thanks for trying to make Jasmine better.

[1]: https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md

slackersoft:
Additionally, it is unclear to me what it would mean to have a filter on the suite level. The current filtering setup for jasmine examines the full description name for the spec (which includes all of the suites it is nested in), so you should be able to filter a suite simply by setting the suite's full description as the filter to get all of the specs define inside it.

ammanyer:
It would help to tag all the specs at the suite level instead of repeating
the tag in each spec.

Thanks,
Ammanuel

Issue #855, opened 2015-05-17
Pending vs disabled


ljwall:
Hi,  I know there's been some discussion of pending vs disabled specs (https://github.com/jasmine/jasmine/issues/840).  I was slightly surprised to discover that disabled specs are not reported (a least in jasmine-npm) so I raised an issue at https://github.com/jasmine/jasmine-npm/issues/45, but i understand this is perhaps the intended behaviour.

Currently I think `xit` marks as pending and `xdescribe` marks as disabled.  It would be good if they did the same thing. If it's clear what the behaviour *should* be I could (try to) work up a pull request.

To my mind, pending is more useful. I'm not sure what the use-case is for disabling specs and not having them show up in reporting?

slackersoft:
As of version 2.3, every spec that is defined in the suite will be reported to all registered reporters. Previously, any specs that were children of an `xdescribe` (and any suites) were marked as disabled and never run at all and did not get reported. With the addition of `fdescribe` and `fit` in jasmine 2.1, it is also possible for a suite or spec to be marked `disabled` if it is not part of the focus run (i.e. not an `fit` or child of an `fdescribe`).

I think you're probably right that `xdescribe` and `xit` should put specs into the same status. I'm not totally sure whether it should be `pending` or `disabled`. I'm leaning towards `pending` because that would mean you get a report at the end of the run of all the specs you've turned off entirely in your suite.

Changing this in either direction would be a breaking change as far as how the behavior of `xit` or `xdescribe` is reported, but I could possibly see it as a bug fix.

I'd be happy to review a pull request that changed it so child specs of an `xdescribe` were marked as `pending` for that run instead of `disabled`.

LJWall:
Ok great, I'll have a go at that then.... (I've not looked properly at the jasmine-core code much, so not yet sure where on the trivial/hard spectrum it is..)

nertzy:
Would it make sense to introduce a new state called `excluded`? If it's hard to decide between `pending` or `disabled` then that is probably a good reason not to use either.

LJWall:
For my part, I thing `pending` is ok.  It accurately describes my use of `xdescribe` and `xit` - i.e. the given suite or spec is pending some correction.

PR #869, opened 2015-05-29
xdescribe marks pending, plus associated tests.

ljwall:
As discussed in issue https://github.com/jasmine/jasmine/issues/855.

Suites have a `.pend()` method, and pending propagates down to children (until hitting a focused item).  This is handled within `env`.  I did wonder if it should be handled by `TreeProcessor`?


ljwall:
Hi, Any comments from the Jasmine team on this?  Any modification I need to make?
Thanks,

PR #879, opened 2015-06-16
Add endsWith() expectation

guyellis:
First time doing a PR on Jasmine. I realize that it's not complete without a test but before I write one: (1) Is this useful and would it be accepted and (2) am I on the right track for the files to be changed?

Thanks.

### DIFF
--- a/src/core/matchers/matchersUtil.js
+++ b/src/core/matchers/matchersUtil.js
@@ -25,6 +25,21 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       return !!haystack && haystack.indexOf(needle) >= 0;
     },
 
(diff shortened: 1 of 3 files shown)


slackersoft:
Because you can get the same functionality provided by `toEndWith` (or `toStartWith`) matchers from the `toMatch` matcher, we are unlikely to implement them in core. (e.g. `toEndWith('foo')` is equivalent to `toMatch(/foo$/)`)

As for the actual implementation of the matcher, we don't want to pull things into `matchersUtil` until they need to be used in multiple matchers, so the actual `endsWith` logic should just be embedded in the matcher directly.

slackersoft:
Closing this since we haven't seen any updates in a while.

PR #908, opened 2015-08-18
Reverse suite afterEach behavior to match semantics?

mcamac:
Hi, thanks for the great library! I've noticed something which I have a question about.

If there are multiple `afterEach`es in a suite, they get run in reverse order of declaration. The current test `spec/core/integration/SpecRunningSpec.js` appends `afterEach2` before `afterEach1`, which is counter-intuitive to me since the test is named `"should run multiple befores and afters in the order they are declared"`.

I've made changes to the `Env` src file and the corresponding spec - what do you think about this behavior change?

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -558,7 +558,7 @@ getJasmineRequireObj().Env = function(j$) {
 
         while(suite) {
           befores = befores.concat(suite.beforeFns);
-          afters = afters.concat(suite.afterFns);
+          afters = afters.concat(suite.afterFns.reverse());
 
           suite = suite.parentSuite;
         }
--- a/spec/core/integration/SpecRunningSpec.js
+++ b/spec/core/integration/SpecRunningSpec.js
@@ -217,10 +217,10 @@ describe("jasmine spec running", function () {
         "beforeEach1",
         "beforeEach2",
         "outer it 1",
-        "afterEach2",
         "afterEach1",
-        "runner afterEach2",
-        "runner afterEach1"
+        "afterEach2",
+        "runner afterEach1",
+        "runner afterEach2"
       ];
       expect(actions).toEqual(expected);
       done();
--- a/src/core/Env.js
+++ b/src/core/Env.js
@@ -108,7 +108,7 @@ getJasmineRequireObj().Env = function(j$) {
 
         while(suite) {
           befores = befores.concat(suite.beforeFns);
-          afters = afters.concat(suite.afterFns);
+          afters = afters.concat(suite.afterFns.reverse());
 
           suite = suite.parentSuite;
         }


bahmutov:
+1 - this makes the runtime order a lot more intuitive and avoids surprises. Also matches the execution in Mocha

m1gu3l:
+1

m1gu3l:
When can next release be expected?

slackersoft:
We're working on it

just-boris:
My tests just have been blown because of this change.

Is there any release note, where I could know it before it happened?

just-boris:
My example is:

1. I am using `angular-mocks` to test my application. It has a `beforeEach` and `afterEach` hooks with angular environment setup.
2. Then have some my app-specific hooks, relied on angular-mocks stuff.

Everything worked fine, but after this change angular's `afterEach` is fired first. Then my hook is fired and it can't finish, because angular environment already has been disposed.

To work around it, now I have to split my helper file to two and place may `afterEach` first, and `beforeEach` as last. 

```
'helpers-finish.js',
'angular-mocks.js',
'helpers-init'
````

This is an insane

rymai:
We just experienced the same issue. In our case, the [`afterEach` of `jasmine-flight`](https://github.com/zeno/jasmine-flight/blob/webpack/lib/jasmine-flight.js#L177-L202) was trigerred before our own `afterEach` in which we were accessing `this.component` which would have been already teardowned by `jasmine-flight`.

I agree that the reason for this change is good but in reality, this will probably break many test suites.

just-boris:
Also found that `mocha` calls `afterEach` as you did it before. Now behaviour of these two tools are different. I think it is not good.

slackersoft:
@just-boris @rymai  sorry for the breaking change, it wasn't intended. The primary reason we even considered this PR was the note from @bahmutov that mocha ran `afterEach` blocks in the new order.

If we are going to make this change (which your use cases seem to suggest it would be a bad idea) we should have done it as part of a major version bump.

I'm going to revert and drop a new release and we can have a bigger discussion about which ordering makes more sense.

just-boris:
@slackersoft Thank you!
I am looking forward for future discussion

rymai:
Thanks @slackersoft! /cc @dharFr

m1gu3l:
@just-boris 
> Also found that mocha calls afterEach as you did it before. Now behaviour of these two tools are different. I think it is not good.

mocha runs afterEach in order of declaration: http://codepen.io/anon/pen/PZoGvW
jasmine (every version except 2.4.0) runs afterEach in reversed order of declaration http://codepen.io/anon/pen/xZxRxv

just-boris:
@m1gu3l yeah, you are right.
I messed with it yesterday, sorry.

amilligan:
In this case the (original) Jasmine behavior is correct.  As several people have discovered, setting up and tearing down the state of tests behaves somewhat like a stack: the last thing you set up should be the first thing you tear down.  Ordering issues become very difficult to track otherwise, as you deal with various libraries, stubs, the clock, etc.

slackersoft:
I'm inclined to agree with @amilligan on ordering of `afterEach` invocations. Running them in the reverse declaration order allows libraries to have global setup and teardown that can be guaranteed to run in the same context.

Jasmine won't be changing the ordering of `afterEach` for the foreseeable future.

alexander-alzate:
I think you should document this behaviour. [Introduction.js](https://jasmine.github.io/edge/introduction.html) doesn't say anything about it.

slackersoft:
@alexander-alzate The source for the [Jasmine documentation](https://jasmine.github.io) has its [own github repository](https://github.com/jasmine/jasmine.github.io). We do accept [pull requests](https://github.com/jasmine/jasmine.github.io/pulls) for our documentation, if you'd like to help out, or [issues](https://github.com/jasmine/jasmine.github.io/issues) if you have things you'd like see addressed but aren't sure how.

PR #919, opened 2015-09-01
Check for child before accessing property

sabrehagen:
I'm encountering a crash in jasmine when there is no 0th child element. It should be checked before accessing the element.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2145,7 +2145,7 @@ getJasmineRequireObj().Suite = function(j$) {
   };
 
   function isAfterAll(children) {
-    return children && children[0].result.status;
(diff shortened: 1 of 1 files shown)


slackersoft:
Thanks for bringing this up. Unfortunately, there are a few problems with your pull request as it stands right now:

1. There are no tests.
1. You have modified `jasmine.js` directly which is a generated file
1. You haven't provided a change to the source file so your changes to `jasmine.js` will be overwritten the next time the file is generated.

Please take a look at [CONTRIBUTING.md](https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md) for more info.

slackersoft:
We haven't heard anything else from the community on this issue and the PR hasn't been updated to solve the issues, so it doesn't seem like the issue is still occurring.

PR #925, opened 2015-09-17
Allow a fake default empty function for SpyStrategy.callFake()

verheyenkoen:
This change allows you to use the `callFake` spy strategy parameterless. A default empty function (which is also the default spy strategy) is used instead.


slackersoft:
For this functionality, what you want to use is `stub`

verheyenkoen:
Damn, totally missed that.

PR #927, opened 2015-09-25
Random tests

marcioj:
This is a ~~WIP~~ PR which enables random tests. Everything is working fine in the browser, I just need to add more tests. And also when running in the node console, display the seed used.

In the `HTMLReporter` a new option was added called `run tests in random order` and it's checked by default. This behavior is also handled by query params when `?random=true`.

In order to reproduce a random order in a consistent way, a seed is generated each time the tests run and passing `seed=myseed` will keep the order, like rspec does.

Let me know if you agree with these changes, and feel free to suggest changes.

### DIFF
--- a/src/core/Order.js
+++ b/src/core/Order.js
@@ -1,0 +1,46 @@
+/*jshint bitwise: false*/
+
+getJasmineRequireObj().Order = function() {
+  function Order(options) {
+    this.random = 'random' in options ? options.random : true;
+    var seed = this.seed = options.seed || generateSeed();
+    this.sort = this.random ? randomOrder : naturalOrder;
+
+    function naturalOrder(items) {
+      return items;
+    }
+
+    function randomOrder(items) {
+      var copy = items.slice();
+      copy.sort(function(a, b) {
+        return jenkinsHash(seed + a.id) - jenkinsHash(seed + b.id);
+      });
+      return copy;
+    }
+
+    function generateSeed() {
+      return String(Math.random()).slice(-5);
+    }
+
+    // Bob Jenkins One-at-a-Time Hash algorithm is a non-cryptographic hash function
+    // used to get a different output when the key changes slighly.
+    // We use your return to sort the children randomly in a consistent way when
+    // used in conjunction with a seed
+
+    function jenkinsHash(key) {
+      var hash, i;
+      for(hash = i = 0; i < key.length; ++i) {
+        hash += key.charCodeAt(i);
+        hash += (hash << 10);
+        hash ^= (hash >> 6);
+      }
+      hash += (hash << 3);
+      hash ^= (hash >> 11);
+      hash += (hash << 15);
+      return hash;
+    }
+
+  }
+
+  return Order;
+};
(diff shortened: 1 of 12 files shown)


lib/jasmine-core/jasmine.js
```diff
@@ -452,6 +453,46 @@ if (typeof window == void 0 && typeof exports == 'object') {
...
+      items.sort(function(item) {
+        return jenkins(seed + item.id);
+      });
+    }
+
+    function generateSeed() {
+      return String(Math.random()).slice(-5);
+    }
+
+    function jenkins(key) {
```
slackersoft: I had to go look up why this would just be called `jenkins` to find the Bob Jenkins algorithm. It would be nice to have something a bit more descriptive

src/core/TreeProcessor.js
```diff
@@ -68,6 +69,8 @@ getJasmineRequireObj().TreeProcessor = function() {
       } else {
         var hasExecutableChild = false;
 
+        willProcessNode(node);
```
slackersoft: `willProcessNode` doesn't really mean anything to me. I think it should be something like `orderChildren`, possibly just taking the children in. I would prefer to be more specific to start with, especially with an internal interface, until we actually find something else that would need to plug in here.

src/core/Order.js
```diff
@@ -0,0 +1,39 @@
...
+    this.random = 'random' in options ? options.random : true;
+    var seed = this.seed = options.seed || generateSeed();
+    this.sort = this.random ? randomOrder : naturalOrder;
+
+    function naturalOrder(items) {
+      return items;
+    }
+
+    function randomOrder(items) {
+      items.sort(function(item) {
```
slackersoft: Ideally, this should just return the new sorted array and not change the order of the existing array.

src/core/Order.js
```diff
@@ -0,0 +1,39 @@
...
+        return jenkins(seed + item.id);
+      });
+    }
+
+    function generateSeed() {
+      return String(Math.random()).slice(-5);
+    }
+
+    function jenkins(key) {
```
slackersoft: I had to go look up why this would just be called `jenkins` to find the Bob Jenkins algorithm. It would be nice to have something a bit more descriptive

slackersoft:
This looks reasonably good. See my comments inline and it looks like the build is failing as well.

A few higher level notes:
* Please don't include the generated `jasmine.js` and `jasmine-html.js` in the pull request. We will do the generation when the PR is merged.
* As part of 2.x, we don't want to change the default ordering of specs. Random should be off by default so that suites that have an order dependency don't break when upgrading a minor version. (We'll probably turn it on by default when 3.0 is released)

marcioj:
Thanks for the feedback @slackersoft ! I updated the code based on your comments.
Please let me know if I need to change something.

myitcv:
Thanks @marcioj we're after exactly this to help strengthen our test suite.

myitcv:
:+1: thanks @slackersoft

myitcv:
For anyone using `jasmine` via `karma`, [this PR](https://github.com/karma-runner/karma/commit/00d63d0b965a998b04d1917d4c4421abc24cec18) was a useful reference. It allows custom options to be passed through, one of which can be a specific seed (useful when reproducing a failed test scenario)

myitcv:
@marcioj - thanks again, this has already helped to catch a significant bug in our code.

Issue #928, opened 2015-09-25
jasmineToString


myitcv:
I notice [this discussion](https://groups.google.com/forum/#!topic/jasmine-js-dev/Fohd0x0bZEg) in the Google Group forum. 

In particular I agree with André's comment:

> shouldn't their custom toString be preferred over the recursive pretty print?

I have expectations like the following:

```typescript
expect(res).not.toBeDefined();
```

where `res` is expected to be undefined but if this fails then it will be an instance of [`immutable-js` `List`](https://facebook.github.io/immutable-js/docs/#/List). 

Is the expectation here that `immutable-js` should define `jasmineToString()` functions for all their types?

myitcv:
**Method 2** in [this SO answer](http://stackoverflow.com/a/26324116/2097883) provides a cleaner means of overriding this behaviour. Still I would argue falling back to `toString()` is sensible behaviour.

slackersoft:
The `jasmineToString` method was introduced specifically to make overriding existing methods in the pretty printer unnecessary.

I would be happy to review a pull request that allows a custom `toString` to take precedence over just the fallback recursion. I would like it to be pretty sure that it is actually custom and will print something more useful than just `[object Object]` or similar.

myitcv:
@slackersoft just created #929. Thoughts/feedback welcomed there

PR #929, opened 2015-09-25
Use toString for objects if it has been overriden

myitcv:
That. Per conversation in #928

### DIFF
--- a/src/core/PrettyPrinter.js
+++ b/src/core/PrettyPrinter.js
@@ -30,6 +30,8 @@ getJasmineRequireObj().pp = function(j$) {
         this.emitScalar('HTMLNode');
       } else if (value instanceof Date) {
         this.emitScalar('Date(' + value + ')');
+      } else if (value.toString && typeof value === 'object' && !(value instanceof Array) && value.toString !== Object.prototype.toString) {
+        this.emitScalar(value.toString());
       } else if (j$.util.arrayContains(this.seen, value)) {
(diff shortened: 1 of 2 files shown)


src/core/PrettyPrinter.js
```diff
@@ -30,6 +30,8 @@ getJasmineRequireObj().pp = function(j$) {
         this.emitScalar('HTMLNode');
       } else if (value instanceof Date) {
         this.emitScalar('Date(' + value + ')');
+      } else if (value && value.toString && typeof value === 'object' && !(value instanceof Array) && value.toString !== Object.prototype.toString) {
```
slackersoft: How many of the conditions in this `if` are required to make this functionality work properly? If we have gotten this far down the `else if` chain, we should know that the `value` is not `undefined` or `null`, so at least the first truthy check on `value` should not be necessary.

myitcv: All but the first (as you rightly point out)

* The `value.toString` check ensures we have a definition for `toString` (if we have a null prototype we won't, it may have explicitly been set to null etc)
* We want to limit this to objects that aren't arrays
* And then the final check ensures we have an actual override of the `toString` method

myitcv:
@slackersoft - force-pushed a revised commit that removes the unnecessary first condition.

myitcv:
@slackersoft - any thoughts on this?

myitcv:
:+1: thanks @slackersoft

Issue #912, opened 2015-08-28
Allow me to pass a reason to `xit`


donaldpipowitch:
It seems I have no way to pass a reason, if I use `xit`. It would be great to do this to get rid of the `No reason given` warning. Can we use a reason as a third param? `xit('should foo', fooFn, 'ignore this')`

slackersoft:
`xit` is meant to be a very short term method to stop a spec from running, so I don't think it makes sense to pass an explicit message. I could however see changing the message for an `xit` to read something like "Temporarily disabled with xit" instead of just "no message given". This would also be a bit more consistent with other BDD frameworks like rspec for ruby.

I'd would be happy to review a pull request to change the default message for `xit`.

JoseRoman:
@slackersoft Is there any particular place in the UI where this message should be placed?

slackersoft:
There shouldn't need to be any changes to the UI for this, it should simply be a change to how `xit` calls `pend` on the spec, so that it passes the message argument.

PR #930, opened 2015-09-25
Allow `xit` to take an optional message to display to the user

lalunamel:
See #912.

This commit allows `xit` to take an optional message to display to the user. It also changes the default message for a pending spec from 'No reason given' to 'Temporarily disabled with xit.'

For example:
```
xit('pending spec', 'function() {
  expect(foo).toEqual(true)
}, 'Waiting on an implementation of `foo`');
```
would print
```
1) pending spec
  Waiting on an implementation of `foo`
```

### DIFF
--- a/spec/core/EnvSpec.js
+++ b/spec/core/EnvSpec.js
@@ -19,6 +19,35 @@ describe("Env", function() {
     });
   });
 
+  describe('#xit', function() {
+    var spyPend, spyIt, pendingReason;
+    beforeEach(function() {
+      spyPend = jasmine.createSpy('pend');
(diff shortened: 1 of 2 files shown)


lalunamel:
Happy to submit a PR for the docs, too.

slackersoft:
Sorry, I think I may have been a bit unclear in the other spec. Jasmine doesn't actually want to allow the message for `xit` to be configurable, it should just be given a hard-coded message by default. If you want to set a pending message, you should be using the `pending` method.

Feel free to either update this PR or create a new one.

PR #946, opened 2015-10-20
Add ES6 Symbol Support to objectContaining

jonnyreeves:
objectContaining will not detect ES6 Symbols in the expected object and use them for comparison


jonnyreeves:
Tests that are failing under CI are due to `Symbol` not being present on the window (the test-cases I've added rely on it; however the library changes do not).

Please could someone suggest the preffered way to rectify this?

jonnyreeves:
Actually, this is probably the wrong approach - feel free to get in touch if you feel otherwise :)

PR #951, opened 2015-10-23
Added "returnModified" spyStrategy

Tireo:
While recently creating a lot of unit tests i lacked:
- ability to easily use original return value of function but with slight changes (and usually different over multiple tests)
- ... without additional (repeatable over tests) lines of code / pre-created mock data.

So I created it ;) 
And hope You will also like it.

### DIFF
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -46,6 +46,29 @@ describe("SpyStrategy", function() {
     expect(returnValue).toEqual(17);
   });
 
+  it("can return a modified result of original function when executed", function() {
+    var originalFn = jasmine.createSpy("original").and.returnValue({
+          a: 1,
+          b: 2
+        }),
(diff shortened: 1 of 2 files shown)


Tireo:
Hi,

It seems that this change suggestion is neither "deny" nor "accept" material for You ;) At least judging from it being the only one not processed in the meantime ;)

I do now believe that my suggestion should be expanded by additionally passing array of original fn arguments as second param to the callback... but unless you think the original suggestion is sound... 
no sense uploading additions to it :)

slackersoft:
This isn't a spy strategy that we want to support in core. Jasmine's preferred solution is to eventually allow users the ability to specify custom strategies spies have access to for their suite.

We have a story to support custom SpyStrategies in our [tracker project](https://www.pivotaltracker.com/n/projects/10606), but it currently isn't planned until after a 2.5 release is out. We would be happy to review a pull request for custom strategies if someone has time though.

I'm going to close this as it's not something we're looking for in jasmine core. Thanks for using jasmine!

PR #953, opened 2015-10-24
feat(spy): add option to respy

### LINKED ISSUE #931: ability to configure "already been spied upon" behavior
currently I get errors if I try to re-spy upon a function. 

this makes my code stateful - which is a serious issue for me. 
I need to remember if this is a spy or not. 
for example: if I decide to have a default spy after written a lot of tests,  I now need to modify the code in all tests from `spyOn(object,'myMethod')` to `object.myMethod...`. seems redundant to me.  

It would be nicer if it worked similar to jquery wrapper.. 

`$('div')` and `$($('div'))` - give the same result. 

`spyOn(obj,'myMethod')` should simply return the spy if spy already exists.

guy-mograbi-at-gigaspaces:
will not throw error if spyOn called on a spy

fixes #931

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -677,6 +677,35 @@ describe("Env integration", function() {
     env.execute();
   });
 
+  it('allows to respy functions', function () {
+
+    var env = new j$.Env(),
+        foo = {
+          'bar': function () {
+            return 1;
+          }
+        };
+
+    env.allowRespy(true);
+
+    env.describe('test suite', function(){
+        env.it('spec 0', function(){
+          env.spyOn(foo,'bar');
+
+          var error = null;
+
+          try {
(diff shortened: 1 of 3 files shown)


sgravrock:
It looks like the allowRespy setting is global. That could be problematic especially in larger test suites. Personally I'd be more comfortable with a design that allowed respying on a per-spy basis, sort of like Cedar's stub_method(...).again(). That way the system could still catch an accidental double-spy in one place even if there's an intentional double-spy somewhere else.

guy-mograbi-at-gigaspaces:
@sgravrock why would anyone want to have an error thrown on re-spy? what is the use case? 

for, it should be the other way around - the default should be `don't throw error ever` and if you want to catch respy, configure it specifically. 
However that is a huge change which I guess is never going to happen.

Issue #896, opened 2015-07-23
describe should throw when it receives a function that expects a `done` callback.


mik01aj:
I spent half an hour debugging this.

    describe('stuff', function (done) {
        it('works good', function () {
            myHttp.get(url).then(function (response) {
                expect(response.data.length).toBeGreaterThan(10000);
            }).then(done, fail);
        });
    });

I got an error in some other unrelated test, with no stacktrace (because async), and had to fire up a debugger to understand what's happening. And it was just a simple mistake.

(In case you still didn't notice: I put `done` in `describe` instead of `it`.)

I believe this wouldn't be hard to do (as you already look at the arguments of the function given in `it`) and would save other programmers from mistakes like this.

slackersoft:
Unfortunately, the `function`s for `it` and `describe` are evaluated in different ways and by different parts of the code. This means that we don't get to automatically take advantage of the existing code for checking argument count.

That said, I wouldn't be averse to having a check for argument count for a `describe`, if you want to work on a pull request.

IgorAufricht:
Sorry for commenting on a closed issue, but I'm not sure it's worth opening a new issue.

The error message can be a bit deceiving if you put some other parameter to the function. For example you could do something like this by mistake:
```
describe('a test', function ($timeout) {
  // ...
});
```
In this case the error is deceiving, since we don't use the `done` parameter anywhere. In this case a message along the lines of 'describe() function expects zero parameters' and/or the location of the specific `describe()` function would help.

A real-life example can be seen here: http://stackoverflow.com/questions/34899998/jasmine-mocking-angular-http-error-describe-does-not-expect-a-done-parameter

mik01aj:
+1 for the idea, I'd just change the text to: "function passed to describe() shouldn't have arguments"

mik01aj:
FYI everyone: this error can also be detected be eslint: https://github.com/tlvince/eslint-plugin-jasmine/blob/master/docs/rules/no-suite-callback-args.md

Maviza101:
Thanks for sharing! Helped me fix the problem.

PR #955, opened 2015-10-25
Added a throw error block in describe in-case a function with argument…

himajasuman:
…s is passed in describe

fixes: https://github.com/jasmine/jasmine/issues/896


himajasuman:
Ping @jridgewell.

himajasuman:
Hi, I tried figuring it out. I am a new to node and jasmine development, but used Jasmine for angular apps though. Some suggestions on where to catch the error(I mean in which file) would really help .

PR #956, opened 2015-10-26
Remove padding and margin from .jasmine_html-reporter

yuyokk:
Hello,

As I see there is a horizontal scroll on the [site](http://jasmine.github.io/2.3/introduction.html)
Removing padding and margin on .jasmine_html-reporter fixes the issue.
Please let me know if this makes sense.

Thank you


slackersoft:
I'm not sure that jasmine-core is the right place to fix this issue. The html reporter is designed for use as the only thing on the page, and if the docs site layout is screwing with things, it should probably be fixed [over there](https://github.com/jasmine/jasmine.github.io).

Closing this in favor of trying to get a fix into the docs themselves. Thanks for using jasmine!

yuyokk:
@slackersoft oh I see, thanks for head up, I will do it [here](https://github.com/jasmine/jasmine.github.io)

PR #965, opened 2015-11-02
Promises

dejakob:
I thought it was a cool idea to implement this:

```
ExternalService.init();
spyPromise(ExternalService, 'init');

$rootScope.$digest();

expect(ExternalService.init).toHaveBeenResolved();
```

(This example is of course only Angular, but it could be useful for all types of promises)

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -99,6 +99,8 @@ getJasmineRequireObj().requireMatchers = function(jRequire, j$) {
       'toHaveBeenCalled',
       'toHaveBeenCalledWith',
       'toHaveBeenCalledTimes',
+      'toHaveBeenResolved',
+      'toHaveBeenRejected',
       'toMatch',
       'toThrow',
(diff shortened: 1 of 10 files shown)


urish:
+1

I actually ended up implementing a similar solution here:

https://github.com/urish/angular-promises-toolkit/blob/master/jasmine-promise-matchers.js

wormyy:
+1

lpww:
Nice idea!

agirorn:
https://github.com/agirorn/jasmine-node-promise-matchers does something similar for Promises when running jasmine in node.js

```javascript
it('should resolve', function(done) {
  expect(ExternalService.init()).toResolve(done);
});
```

slackersoft:
I think mocked promises is something we want to keep out of Jasmine itself for the time being. It should be reasonably easy to provide a library that has a mock Promise and some expectations for them without needing anything added to Jasmine.

Closing this. Thanks for using Jasmine!

Issue #512, opened 2014-01-31
Make bootstrap AMD/CommonJS compliant


i-like-robots:
Currently Jasmine 2 cannot be used (or at least requires a lot of brute force) to test modular JavaScript, such as files authored as AMD modules to be used with a script loader such as RequireJS. Fundamentally I think the only limitation is this line: https://github.com/pivotal/jasmine/blob/master/lib/jasmine-core/boot.js#L187

I think it would be extremely useful if the boostrap could export a simple interface for use with a script loader. For example:

``` js
function kickoff() {
  if (currentWindowOnload) {
    currentWindowOnload();
  }
  htmlReporter.initialize();
  env.execute();
};

if ( typeof define === 'function' && define.amd ){
  define(function() {
    return kickoff;
  });
}
else if ( typeof module !== 'undefined' && module.exports ) {
    module.exports = kickoff;
}
else {
  window.onload = kickoff;
}
```

I'm quite probably way oversimplifying the issue (I don't have much knowledge of the Jasmine internals) but this would be really, really useful!

PR #970, opened 2015-11-11
UMD wrapping to support AMD natively

amenadiel:
This pull request adds AMD support by the means of
1. Detecting if the `define` function is present, and proceeding accordingly
2. Wrapping `jasmine-html.js` in an IIFE  to decide if the `jasmineRequire` object should be loaded with AMD, CommonJS or directly from the window object.

I had added the new  `amd_boot.js` script to comply with @infews statement in https://github.com/jasmine/jasmine/issues/512 that boot scripts should be per-environment... but Travis won't pass with that extra file sitting in there, and I'm not sure it should be inlined either. So, the AMD boot stript should be part of the require main script:

``` js
requirejs.config({
    paths: {
        "jasmine":"vendor/jasmine-core/jasmine",
        "jasmine-html":"vendor/jasmine-core/jasmine-html",
        "css": "vendor/require-css/css",
    }
});

define([
    'vendor/jasmine-core/jasmine',
    'vendor/jasmine-core/jasmine-html',
    'css!vendor/jasmine-core/jasmine'
], function (jasmineRequire) {

    var jasmine = jasmineRequire.core(jasmineRequire);
    jasmineRequire.html(jasmine);
    var env = jasmine.getEnv();
    var jasmineInterface = jasmineRequire.interface(jasmine, env);
    var queryString = new jasmine.QueryString({
        getWindowLocation: function () {
            return window.location;
    }
    });

    var catchingExceptions = queryString.getParam('catch');
    env.catchExceptions(typeof catchingExceptions === 'undefined' ? true : catchingExceptions);

    var throwingExpectationFailures = queryString.getParam('throwFailures');
    env.throwOnExpectationFailure(throwingExpectationFailures);

    var random = queryString.getParam('random');
    env.randomizeTests(random);

    var seed = queryString.getParam('seed');
    if (seed) {
        env.seed(seed);
    }
    var htmlReporter = new jasmine.HtmlReporter({
        env: env,
        onRaiseExceptionsClick: function () {
            queryString.navigateWithNewParam('catch', !env.catchingExceptions());
    },
        onThrowExpectationsClick: function () {
            queryString.navigateWithNewParam('throwFailures', !env.throwingExpectationFailures());
        },
        onRandomClick: function () {
            queryString.navigateWithNewParam('random', !env.randomTests());
    },
        addToExistingQueryString: function (key, value) {
            return queryString.fullStringWithNewParam(key, value);
        },
        getContainer: function () {
        return document.body;
        },
        createElement: function () {
            return document.createElement.apply(document, arguments);
    },
        createTextNode: function () {
            return document.createTextNode.apply(document, arguments);
        },
    timer: new jasmine.Timer()
    });

    env.addReporter(jasmineInterface.jsApiReporter);
    env.addReporter(htmlReporter);

    var specFilter = new jasmine.HtmlSpecFilter({
        filterString: function () {
            return queryString.getParam('spec');
        }
    });

    env.specFilter = function (spec) {
        return specFilter.matches(spec.getFullName());
    };

    /**
     * Setting up timing functions to be able to be overridden. Certain browsers (Safari, IE 8, phantomjs) require this hack.
     */
    window.setTimeout = window.setTimeout;
    window.setInterval = window.setInterval;
    window.clearTimeout = window.clearTimeout;
    window.clearInterval = window.clearInterval;

    var currentWindowOnload = window.onload;

    /**
     * since you're using AMD, there's a chance that the window.onload event has already fired. We check that with the readyState property of the DOM
     */
    if (window.document.readyState === 'complete') {
        htmlReporter.initialize();
    } else {
        window.onload = function () {
            if (currentWindowOnload) {
        currentWindowOnload();
            }
        htmlReporter.initialize();
        };
    }

    var describe = env.describe,
        expect = env.expect,
        it = env.it,
        afterEach = env.afterEach;

    describe('Standard Test Cases: ', function () {
       ...
    });

});


```


rafaellyra:
:+1:

slackersoft:
This looks like a fairly low impact change to get better support for AMD projects, but there are a few problems with this PR as it stands:

* You have not made changes to the src files used to generate `jasmine.js` and `jasmine-html.js`
* You have included the generated files `jasmine.js` and `jasmine-html.js` in your pull request

Additionally, since this work is largely useless without the amd-boot file, it should be included in the project. We will need to either exclude it from the npm package or update the test for where it should be included.

amenadiel:
@slackersoft you're right, I'll look for the src files to avoid modifying the generated ones directly. 

I'll update the test also to allow travis to pass.

amenadiel:
@slackersoft I'm hitting a bit of a wall. The build task concatenates jasmine-html asumming there is a `jasmineRequire` object in the global namespace. 

However, in AMD environment, at some point there has to be an async method to allow jasmine-html.js to wait for jasmine.js to be completely loaded. Leaving aside the implied coupling of a global jasminRequire object, I can't control the loading order unless I add an explicit shim config to say that jasmine-html depends on jasmine. I'd leave that as a last resort.

I can't use the concat task to wrap jasmine-html in a proper factory function, because the only workaround grunt-contrib-concat offers is to tamper with the banner and footer templates, but those are already taken.

slackersoft:
This fell off my radar for a bit, but we would like to have a bit of a better way for Jasmine to run inside this type of environment. We might need to go so far as to generate a second set of files `jasmine.js` and `jasmine-html.js` that would work better in an environment like this, or provide a boot file that would setup whatever shims are necessary to make it work. I would prefer to keep all of that kind of logic within the boot file if possible, even if that means some extra shims and things.

I haven't worked much in and AMD environment to know what is really necessary, but I would also be happy to have an additional travis setup that attempts to use Jasmine through this path.

Hope this helps!

amenadiel:
In case anybody cares, I ran the example (jasmine standalone for browser environment) using [jspm](https://github.com/jspm/jspm-cli) and [system.js](https://github.com/systemjs/systemjs)

```bash
jspm install npm:jasmine-core
```

I then wrote a shim for the package stating:

```json
{
  "main": "lib/jasmine-core/jasmine-html.js",
  "format": "global",
  "meta": {
    "lib/jasmine-core/jasmine.js": {
      "exports": "getJasmineRequireObj",
      "format": "global"
    },
    "lib/jasmine-core/jasmine-html.js": {
      "deps": [
        "./jasmine.js"
      ],
      "format": "global",
      "exports": "jasmineRequire"
    },
    "*.json": {
      "format": "json"
    }
  }
}
```
This means:
 - when you load `jasmine-core` it will point you to `lib/jasmine-core/jasmine-html.js`
 - `jasmine-html.js` is a global module, exports `jasmineRequire` to the global scope and depends on `jasmine.js`
 - `jasmine.js` is a global module and exports `getJasmineRequireObj` to the global scope

then you make your own `amdboot.js` where you would load `jasmine-core` from an AMD module using 

```js
define(['jasmine-core'], function() {

  // everything that goes on boot.js
  // EXCEPT for the window.onload listener
  // this module returns:
  var startReporter = function () {
    htmlReporter.initialize();
    env.execute();
  };

  return startReporter;
});
```

Then you would run the tests doing:

```js
define([
	'./jasmineboot.js',
	'./src/Player.js',
	'./src/Song.js',
	'./spec/SpecHelper.js',
	'./spec/PlayerSpec.js'
], function (startReporter) {

	startReporter();
});

```


I left a minimal working example at: https://embed.plnkr.co/soSHAA2W6X9juyWHp7MY/

Issue #971, opened 2015-11-11
Is there a possibility in jasmine not to override messages from passed expects?


lynchblue:
I am using in my project protractor for my e2e tests with jasmine2 as framework. I am using as well the libraries [jasmine2-custom-message](https://github.com/avrelian/jasmine2-custom-message) and [protractor-jasmine2-html-reporter](https://github.com/Kenzitron/protractor-jasmine2-html-reporter). 

I have realised that the message from passed expects would be overridden with the word: Passed.

After a little research I realised that the file that is actually overridden this message is the following:
node_modules/protractor/node_modules/jasmine/node_modules/jasmine-core/lib/jasmine-core/jasmine.js

Specifically the function:
```javascript
getJasmineRequireObj().buildExpectationResult = function () {
    function buildExpectationResult(options) {
        var messageFormatter = options.messageFormatter || function () {
                },
            stackFormatter = options.stackFormatter || function () {
                };

        var result = {
            matcherName: options.matcherName,
            message: message(),
            stack: stack(),
            passed: options.passed
        };

        if (!result.passed) {
            result.expected = options.expected;
            result.actual = options.actual;
        }

        return result;

        function message() {
            if (options.passed) {
                return 'Passed.';
            } else if (options.message) {
                return options.message;
            } else if (options.error) {
                return messageFormatter(options.error);
            }
            return '';
        }

        function stack() {
            if (options.passed) {
                return '';
            }

            var error = options.error;
            if (!error) {
                try {
                    throw new Error(message());
                } catch (e) {
                    error = e;
                }
            }
            return stackFormatter(error);
        }
    }

    return buildExpectationResult;
};
```

I wonder if there is a way to override this function (buildExpectationResult), for example in protractor config file. I have already tried that by adding in protractor config file the following.
```javascript
onPrepare: function () {
    jasmine.buildExpectationResult = function () {
        function buildExpectationResult(options) {
            var messageFormatter = options.messageFormatter || function () {
                    },
                stackFormatter = options.stackFormatter || function () {
                    };

            return {
                matcherName: options.matcherName,
                expected: options.expected,
                actual: options.actual,
                message: message(),
                stack: stack(),
                passed: options.passed
            };

            function message() {
                if (options.passed && !options.message) {
                    return 'Passed.';
                } else if (options.message) {
                    return options.message;
                } else if (options.error) {
                    return messageFormatter(options.error);
                }
                return "";
            }

            function stack() {
                if (options.passed) {
                    return "";
[...]

PR #972, opened 2015-11-12
Avoid overriding 'expect' messages when 'expect' passes

lynchblue:
When using jasmine reporters it would be nice if we could also see the messages from those 'expects' that passed. 

In case that the 'expect' doesn't provide a message, we could set it to 'Passed.' (current behaviour for all passed 'expects') but if message is provided we should populate it instead of overriding it with 'Passed'.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -1556,7 +1556,7 @@ getJasmineRequireObj().buildExpectationResult = function() {
     return result;
 
     function message() {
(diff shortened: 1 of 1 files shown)


slackersoft:
See my note in #971.

Additionally, please read the [contributors guide](https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md), as you have modified the generated file, you haven't written any tests, and you haven't modified the source files.

Closing.

PR #986, opened 2015-12-04
Adding callBefore and callAfter to SpyStrategy

magicalhobo:
This makes writing async tests easier, because you can allow the original function to be called, then immediately make assertions or continue to the next step.

### DIFF
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -92,6 +92,46 @@ describe("SpyStrategy", function() {
     expect(returnValue).toEqual(67);
   });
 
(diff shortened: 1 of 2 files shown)


slackersoft:
This isn't a spy strategy that we want to support in core. Jasmine's preferred solution is to eventually allow users the ability to specify custom strategies spies have access to for their suite.

We have a story to support custom SpyStrategies in our [tracker project](https://www.pivotaltracker.com/n/projects/10606), but it currently isn't planned until after a 2.5 release is out. We would be happy to review a pull request for custom strategies if someone has time though.

I'm going to close this as it's not something we're looking for in jasmine core. Thanks for using jasmine!

Issue #872, opened 2015-06-01
.toHaveBeenCalledWith and modified objects


shamansir:
The actual problem happens thanks to object references—I realize it's not a bug, but it really looks like that from the first sight. Probably you have a solution or some advice for that.

```javascript
describe('spy issue', function() {
    var foo = {};
    function injectAndCall(prop, val, cb) {
        foo[prop] = val;
        console.log(foo);
        cb(foo);
    };

    it('spy issue', function() {
        var mySpy = jasmine.createSpy('test');

        injectAndCall('a', 0, mySpy);
        injectAndCall('b', 1, mySpy);

        expect(mySpy).toHaveBeenCalledWith({ a: 0 });
        expect(mySpy).toHaveBeenCalledWith({ a: 0, b: 1 });
    });
});
```

The response is: 

`Expected spy test to have been called with [ Object({ a: 0 }) ] but actual calls were [ Object({ a: 0, b: 1 }) ], [ Object({ a: 0, b: 1 }) ].`

If you check the console, however, you'll get:

```
{ a: 0 }
{ a: 0, b: 1 }
```

May be this could be solved with deep-cloning arguments for spies?

This example is quite easy, but this also could happen when `foo` has a reference to another object inside, and the latter in its turn is a subject to modifications.

shamansir:
...or could there be a special matcher like `.toHaveBeenCalledWith(jasmine.identicalTo({ a: 0 }))`?

shamansir:
I've looked up for a possible reason in source and I discovered that indeed there's a deep-check there, but of course for this case object in this line: https://github.com/jasmine/jasmine/blob/master/src/core/matchers/matchersUtil.js#L159 is already modified. So the case in the example is easily solved with: 

```javascript
to_test('a', 0, mySpy);
expect(mySpy).toHaveBeenCalledWith({ a: 0 });
to_test('b', 1, mySpy);
expect(mySpy).toHaveBeenCalledWith({ a: 0, b: 1 });
```

Could not be as easy in complex cases when one have no ability to put an expectation between the calls, though.

shamansir:
In my code I solved it with:

```javascript
mySpy.and.callFake(function() {
    if (mySpy.calls.count() === 1) {
        expect(mySpy).toHaveBeenCalledWith({ 'a': 0 }, jasmine.anything());
    } else if (processSpy.calls.count() === 2) {
[...]

PR #1000, opened 2015-12-22
Cloned args on calltracker

a-r-d:
Created an optional object argument clone switch on the CallTracker. See this issue that was marked ready to work:
https://github.com/jasmine/jasmine/issues/872#issuecomment-164467203

Let me know I need to change anything! I'll be happy to fix any issues.


ideadapt:
Does anyone know why `saveArgumentsByValue()` is not the default behaviour? Because the way it works now by default (storing call args by reference), is surprising, its not what one expects. 
In my case, this false positive behaviour even led to a bug in production and cost me some time to spot the root cause, because all test were green :)

slackersoft:
This new functionality is not the default partially by virtue of being newer to Jasmine, but also because copying the arguments by value is more expensive. Both in time and complexity to do the copy, as well as the potential for memory bloat in having multiple copies of the same data around.

How often you encounter this also depends a bit on how you write your expectations. I tend to use object literals directly in my `toHaveBeenCalledWith()` expectations which bypasses this issue and helps me ensure I'm checking for things I care about and not everything that is possible.

I'd be happy to look into making this the default (and probably removing the non-copy version), but I'm not sure if this might break existing suites, so it might have to wait until a major version.

PR #1006, opened 2015-12-24
Remove an extra `topSuite` `queueRunner` construction parameter

jurko-gospodnetic:
The `Suite` constructor function does not expect or use its `attrs` parameter's `queueRunner` attribute, so setting it serves no purpose and only clutters the code and makes it more difficult to understand.

This is a pure refactoring commit with no functional changes.


jurko-gospodnetic:
Yay... it's in... already forgot about this pull request with it hanging here as long as it did :-D

PR #1009, opened 2015-12-30
Adding support to a fallback report in case of no added reports

mauricioborges:
Hi,

I'm struggling with the following. Using jasmine on node, when I try to add a custom report through a helper using:

```node
jasmine.getEnv().addReport(myReportObject);
```
When jasmine executes, it outputs my custom report and also the default one.

I'd like to complement the solution proposed in https://github.com/jasmine/jasmine-npm/issues/21 with a support for a single fallback instead of a default report in jasmine-core. It would solve this problem, which I've seen in some blogs that is a common problem and people prefer to workaround it.

Having this pull request approved, I'd like to submit another one th jasmine-npm, regarding the usage of fallbackReport while loading the default one.

Sorry if I'm missing the point here. Thanks and regards

### DIFF
--- a/spec/core/ReportDispatcherSpec.js
+++ b/spec/core/ReportDispatcherSpec.js
@@ -37,4 +37,27 @@ describe("ReportDispatcher", function() {
       dispatcher.foo(123, 456);
     }).not.toThrow();
   });
+
+  it("allows providing a fallback reporter in case there's no other report", function() {
+    var dispatcher = new jasmineUnderTest.ReportDispatcher(['foo', 'bar']),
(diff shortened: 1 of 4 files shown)


invernizzie:
Is it possible that this broke reporting on nodejs? Since 2.5, I'm getting no output when running my tests.

jbunton-atlassian:
@invernizzie, yeah I think you're right. See https://github.com/jasmine/jasmine-npm/issues/90

slackersoft:
Fix for the broken reporting should be available in 2.5.1

Issue #1025, opened 2016-01-20
Improve error messages


dhoko:
Hi,
Some errors are confusing ex: with spies, toThrow etc.

Ex:
```
spyOn(obj.list)
// No method name supplied
```

It's hard to figure out where is the error.
So I'm rewriting errors:

Ex
![screenshot from 2016-01-20 14 16 41](https://cloud.githubusercontent.com/assets/713283/12450062/b573d6dc-bf81-11e5-8200-d9cf35c6aebb.png)
to

![screenshot from 2016-01-20 14 16 19](https://cloud.githubusercontent.com/assets/713283/12450067/c5037ee0-bf81-11e5-903d-c647fea12c68.png)

Do you have any suggestions?

PR #1026, opened 2016-01-20
Improve errors with the domain and how to use the API 

dhoko:
cf issue #1025 

Ex error:

screenshot from 2016-01-20 16 04 53


dhoko:
Ping ?

PR #1029, opened 2016-01-27
[promise] Added promise support

frankandrobot:
Instead of 
```
it('handles a promise', done => {
   promise
      .then(result => expect(result).toEqual(expected))
      .catch(error => expect(error).toBeUndefined())
      .finally(done)
})
```
You can now do this:
```
it('handles a promise', () => {
   return promise.then(result => expect(result).toEqual(expected))
})
```


slackersoft:
Jasmine does not currently want to support promises in this way. We feel that the `done` and `done.fail` functions makes the interface simple enough without any additional code in Jasmine.

Additionally, there are some issues with this pull request:

* You have not included any tests
* The PR build on travis is failing on PhantomJS
* The PR build on travis is failing due to jshint errors.

Thanks for using Jasmine! Closing.

Issue #675, opened 2014-09-18
toEqual should list differences instead of pretty printing objects


mwisnicki:
Consider this simple example ([jsfiddle](http://jsfiddle.net/6y65oh1g/)):

``` js
var actual = { x: 1, y: { z: 2 } };
var expected = { x: 1, y: { z: 3 }};
expect(actual).toEqual(expected);
```

This outputs:
`Expected { x : 1, y : { z : 2 } } to equal { x : 1, y : { z : 3 } }.`

In other words it pretty prints both objects.

Now imagine we are in real world where objects are quite complex and toEqual quickly becomes useless.

I would expect output like:

```
Expected no differences but found 1:
1. Expected y.z = 2 to equal 3
```

i.e. list all differences with paths.

mzgol:
This is what [chai]() prints:

```
      AssertionError: expected { Object (id, issued_at, ...) } to deeply equal { Object (id, issued_at, ...) }
      + expected - actual

               "featuredCollectionsCount": 0
             }
             "avatar": {
               "origin": "gravatar"
      +        "url": "about:blankk"
      -        "url": "about:blank"
             }
             "email": "foo.bar@example.com"
             "fullname": "Foo Bar"
             "id": 42
      
```
So nice, I can immediately see what's wrong. In large objects like that Jasmine output is basically useless, it's completely unreadable. I have to copy the output to a separate file, hunt for "equal to" and split the string by that, copy both parts to two files and compare using my own eyes.

slackersoft:
We'd be happy to review a pull request for this that can work both in the `HtmlReporter` and `ConsoleReporter` (probably the one in [jasmine-npm](https://github.com/jasmine/jasmine-npm) since core's is now deprecated).

mzgol:
@slackersoft Mocha uses the [diff](https://www.npmjs.com/package/diff) package to display these readable diffs. Would you accept a PR that would use this package to display object differences in Jasmine?

slackersoft:
We'd like to keep the core of jasmine with no dependencies. Aside from just not wanting to install a bunch of dependencies on your system just so you can test your javascript, this also means we can release a standalone version that doesn't need much setup, and the platform specific versions (ruby, python, npm) don't need to figure out how to get the same dependencies in each environment.

mzgol:
@slackersoft 
> Aside from just not wanting to install a bunch of dependencies on your system just so you can test your javascript

[...]

PR #1042, opened 2016-02-05
toEqual reports detailed comparison of objects and arrays

benchristel:
This is a fix for issue #675. It causes the `toEqual` matcher to pinpoint mismatches in its failure message when the items being compared are `Object`s or `Array`s.

The output from comparing these objects...

```javascript
var actual = {
      foo: [
        {bar: 1, things: ['a', 'b']},
        {bar: 2, things: ['a', 'b']}
      ],
      baz: [
        {a: {b: 1}}
      ]
    }

var expected = {
      foo: [
        {bar: 2, things: ['a', 'b', 'c']},
        {bar: 2, things: ['a', 'd']}
      ],
      baz: [
        {a: {b: 1, c: 1}}
      ]
    }
```

...looks like this:

```
Expected { ... } to equal { ... }.
    Expected $.foo[0].bar = 1 to equal 2
    Expected $.foo[0].things = Array[2] to have length 3
    Expected $.foo[1].things[1] = "b" to equal "d"
    Expected $.baz[0].a to have properties
        c: 1
```

(`{...}` above is a placeholder for the entire pretty-printed object—i.e. what's currently displayed in the output.)

### DIFF
--- a/src/core/matchers/toEqual.js
+++ b/src/core/matchers/toEqual.js
@@ -1,4 +1,5 @@
 getJasmineRequireObj().toEqual = function() {
+  'use strict';
 
   function toEqual(util, customEqualityTesters) {
     customEqualityTesters = customEqualityTesters || [];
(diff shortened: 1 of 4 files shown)


src/core/matchers/toEqual.js
```diff
@@ -11,9 +12,126 @@ getJasmineRequireObj().toEqual = function() {
...
         result.pass = util.equals(actual, expected, customEqualityTesters);
 
+        if (!result.pass) {
+          result.message = buildMessage(actual, expected);
+        }
+
         return result;
       }
     };
+
```
benchristel: One concern that I have is that maybe `buildMessage` and its helper functions should be moved into a separate file. I didn't extract functions as aggressively as I otherwise would have because I wasn't comfortable with the amount of nesting that it produced when everything was in this file. However, I wasn't sure where else to put this functionality — maybe just in another file in `matchers`? I'd like to get the project maintainers' opinion on this.

Gerg:
Thanks for the PR. We are definitely interested in something like this. I'll try to find time to check in out sometime soon. @slackersoft should take a look also.

ymhuang0808:
Is the feature suitable for `toHaveBeenCalledWith()` and so on?

slackersoft:
After some time to review this, my major concern is that mechanism to generate the failure message involves going back through the object graph and effectively re-doing the equality comparison at each level. This means that the primary use case for breaking out the differences (large objects) are hit even harder when they fail.

To fix this, the `eq` method would have to keep track of differences as it found them (this is actually what happened in 1.3, but I don't think we actually did anything with it). The first iteration probably doesn't need to worry about custom equality testers and asymmetric matchers, but it's something we'll have to figure out at some point.

@ymhuang0808 the current version of this PR would only change the error message for `toEqual`, moving the difference generation down into `eq` should allow anything else that uses it, including `toHaveBeenCalledWith`, access to the same error messages.

ymhuang0808:
Ok, @slackersoft .
I don't know why the large objects may be hit harder.

benchristel:
@slackersoft I originally tried what you suggest — putting the diff generation in `eq` — but moved away from that strategy because

- `eq` seemed like a good single-purpose tool for checking equivalence of values, and I didn't want to overload it with extra responsibilities
- `eq` is pretty complex and I didn't want to break it. The test coverage seems pretty good but the current version of `eq` has been battle-tested.
- If the diff generation is done only when the matcher fails, performance costs will be borne only by failing tests. The performance of green suites will not be impacted.

@ymhuang0808 re: performance in the current PR: I believe the number of times any individual node is touched by `eq` when generating the diff is O(n) in its depth, so worst case, the diff generation is O(n * m), where n is the max depth and m is the number of nodes.

I can do some informal performance testing and find out how bad this really is for large objects.

Gerg:
| If the diff generation is done only when the matcher fails, performance costs will be borne only by failing tests. The performance of green suites will not be impacted.

This should still be the case, since you would only have to do diff work when there is a comparison failure inside the object, which will result in a failed expectation.

benchristel:
Good point. I'll work on moving the diff code to `eq` this weekend. Should I open a new pull request or make new commits on this one?

slackersoft:
I think it's fine to keep the code in this PR, but we'll want to squash it down, so if it's easier for you to just make a new PR that's fine as well.

lele85:
Any news on that?

benchristel:
I've restarted this on a new branch: #1163

PR #1044, opened 2016-02-10
Internet Explorer is dead, long live Internet explorer

WilliamDunne:


Gerg:
This is based off of what browsers Jasmine tests against on Travis: https://github.com/jasmine/jasmine/blob/master/.travis.yml. Jasmine will continue to support old IEs as long as possible since some of our users still work on projects that need to support them.

PR #1046, opened 2016-02-15
When the HtmlReporter has a 'spec' query param, the spec list only shows matching specs/suites

### LINKED ISSUE #510: Filtering out specs in HTML report
In Jasmine 1.3, clicking on a suite or a specific spec (e.g. `TestRunner.html?spec=A spy`) would only show that spec (or tree) in the HTML report, which was really handy. Jasmine 2.0 seems to write out the entire tree, with disabled specs having a different style, which can be quite verbose (especially with thousands of specs in the report). 

Is there a way to enable 1.3-like reports in 2.0? If not, would you guys be open to a pull request to restore 1.3-like behavior (or at least have the option to revert to that mode with a query parameter)?

d-reinhold:
[Fixes #510]
- This implementation does not introduce a new option/checkbox, which I find to be unnecessary/undiscoverable. If a user has filtered their specs, the Spec List should only show matches. A concern with this approach is that specs might be missing due to syntax errors. If this is the case, the user will notice the discrepancy when running the full suite with no filters. I doubt many users of jasmine commit their changes after only running a filtered set of suites (I could be wrong though).
- This commit does not close github issue #510; it just implements a possible fix (suggested by github user 'callmevlad'). I'd like to discuss this implementation with the Jasmine maintainers before submitting a full patch with additional tests.

### DIFF
--- a/lib/jasmine-core/boot/boot.js
+++ b/lib/jasmine-core/boot/boot.js
@@ -49,6 +49,8 @@
     getWindowLocation: function() { return window.location; }
   });
 
+  var filterSpecs = !!queryString.getParam("spec");
+
   var catchingExceptions = queryString.getParam("catch");
   env.catchExceptions(typeof catchingExceptions === "undefined" ? true : catchingExceptions);
 
@@ -76,7 +78,8 @@
     getContainer: function() { return document.body; },
     createElement: function() { return document.createElement.apply(document, arguments); },
     createTextNode: function() { return document.createTextNode.apply(document, arguments); },
-    timer: new jasmine.Timer()
+    timer: new jasmine.Timer(),
+    filterSpecs: filterSpecs
   });
 
   /**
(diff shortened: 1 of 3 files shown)


slackersoft:
I think we probably have enough other feedback on the page that specs are being skipped, that it should be ok to hide the specs that weren't run from the summary view as long as the symbol summary and alert sections stay the same.

d-reinhold:
@slackersoft Alert section and symbol summary sections are unaffected by this change. I've rebased this commit against jasmine/master and added a test to the HtmlReporterSpec. Let me know what you think.

voltrevo:
@d-reinhold This kinda breaks the way I use the ui. My tests take a long time to run, so to get the spec list I frequently use ?spec=asdf so I can use it to run tests by clicking on them.

@slackersoft Would you still be receptive to a PR which would add a checkbox for this feature? Or perhaps is there another way to see the spec list I'm missing?

(While using fit/fdescribe is sometimes useful, it's not my preferred method. I don't want to wait for a new build, and I also like to host static test environments that include the test runner.)

slackersoft:
@voltrevo I would definitely still be happy to take a look at a pull request that adds a checkbox to show the full spec tree even with filtered specs.

PR #1055, opened 2016-02-25
Now spies preserve original function arity

### LINKED ISSUE #991: Jasmine spies do not preserve function arity
Whether using `spyOn` to replace an object property whose value is a function, or using `.and.callFake` to delegate to a supplied function, the resulting spy always has arity (`.length`) of zero — even if the original function has arity of 1, 4, whatever.

This is irrelevant much of the time but becomes critical when testing libraries that check function arity. Two well-known examples:

* libraries which check to see if a `done` param is defined, in order to determine whether to execute asynchronously. Jasmine itself does this, ironically.
* libraries like Express which register "error-handling middleware" as being functions with arity 4 (as opposed to the usual 3 or less), and which execute those functions if a truthy value has been passed into a previously-supplied `next` function (or if an error was thrown during the execution of "normal" middleware).

The best spies which substitute for existing functions do so transparently, so that the library is not able to determine the difference between a spy and an original. Jasmine spies would be improved by matching the arity of their originals, reducing the possibility of the framework being tested exhibiting different behavior.

kapke:
Resolves #991 
Sadly I needed to use `eval` for make spies working as expected.

### DIFF
--- a/src/core/Spy.js
+++ b/src/core/Spy.js
@@ -1,0 +1,58 @@
+getJasmineRequireObj().Spy = function (j$) {
+
+  function Spy(name, originalFn) {
+    var args = buildArgs(),
+      /*`eval` is the only option to preserve both this and context:
+        - former is needed to work as expected with methods,
+        - latter is needed to access real spy function and allows to reduce eval'ed code to absolute minimum
+        More explanation here (look at comments): http://www.bennadel.com/blog/1909-javascript-function-constructor-does-not-create-a-closure.htm
+       */
+      /* jshint evil: true */
+      wrapper = eval('(function (' + args + ') { return spy.apply(this, Array.prototype.slice.call(arguments)); })'),
+      spyStrategy = new j$.SpyStrategy({
+        name: name,
+        fn: originalFn,
+        getSpy: function () {
+          return wrapper;
+        }
+      }),
+      callTracker = new j$.CallTracker(),
+      spy = function () {
+        var callData = {
+          object: this,
+          args: Array.prototype.slice.apply(arguments)
+        };
+
+        callTracker.track(callData);
+        var returnValue = spyStrategy.exec.apply(this, arguments);
+        callData.returnValue = returnValue;
+
+        return returnValue;
+      };
+
+    function buildArgs() {
+      var args = [];
+
+      while (originalFn instanceof Function && args.length < originalFn.length) {
+        args.push('arg' + args.length);
+      }
+
+      return args.join(', ');
+    }
+
+    for (var prop in originalFn) {
+      if (prop === 'and' || prop === 'calls') {
+        throw new Error('Jasmine spies would overwrite the \'and\' and \'calls\' properties on the object being spied upon');
+      }
+
+      wrapper[prop] = originalFn[prop];
+    }
+
+    wrapper.and = spyStrategy;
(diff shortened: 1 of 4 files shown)


glebec:
I haven't tested it carefully yet but it looks good to me.

> Sadly I needed to use `eval` for make spies working as expected.

Yes, but as you point out, there isn't much choice for dynamically making spies of a given arity. It would appear for example that [Sinon uses hard-coded proxy functions limited to 12 arguments at most](https://github.com/sinonjs/sinon/blob/2c89183415ad34f1627dc87094dd88bc1e57b2f2/lib/sinon/spy.js#L78) and they warn against using `eval` due to products that use it blocking that keyword. Do we have similar concerns? Which is the lesser of the two evils — limitless arity or inclusion of `eval`? That is perhaps up to the project maintainers to choose.

src/core/Spy.js
```diff
@@ -0,0 +1,58 @@
...
+
+  function Spy(name, originalFn) {
+    var args = buildArgs(),
+      /*`eval` is the only option to preserve both this and context:
+        - former is needed to work as expected with methods,
+        - latter is needed to access real spy function and allows to reduce eval'ed code to absolute minimum
+        More explanation here (look at comments): http://www.bennadel.com/blog/1909-javascript-function-constructor-does-not-create-a-closure.htm
+       */
+      /* jshint evil: true */
+      wrapper = eval('(function (' + args + ') { return spy.apply(this, Array.prototype.slice.call(arguments)); })'),
```
glebec: What if we did:

```js
var spy = /* full spy def., which can close over wrapper */,
    wrapper = (function(spy){ return Function(args, 'return spy.apply(this, arguments);') }(spy))
```

Then you get a "closure" really just formal param from the IIFE) of the spy, and you also get a working `this` value.

kapke: I was trying with something like that before I used an `eval`. It looks that when function is created with `new Function` constructor, then it losts all of context in which it was created. So then I tried passing a `spy` as first argument, but then I went into troubles with preserving `this` which resulted the line above.

glebec: Dang. After playing around with it for a while, I now better understand exactly what you mean — `Function` does not create closures. You can sort of force the `spy` inside the function body if you `bind(spy)` and use `this.apply` but then you can't change the binding of the returned spy. Basically it's just not possible to do without `eval`, as you already figured out. Or you can hard-code in the proxies like Sinon does but then you should add documentation to the effect of "Jasmine spies preserve arity up to length XX but not longer" or similar.

How frustrating.

glebec:
Following our discussion I would highlight as a **TL;DR** for the project maintainers. Make your choice:

* Allow `eval` into Jasmine. This may be a breaking change. Currently in this PR.
* Change this to be a bunch of hard-coded proxies à la Sinon. That limits how high our simulated arity can go. Can easily refactor this PR to use that technique. If you do this, be sure to document it also so nobody with a 45-argument function is surprised.

I have no opinion either way, the first is better from a dynamic perspective but the pros and cons are worth considering.

glebec:
@slackersoft any idea of when/if this PR might be considered for merge? I have a couple projects on the back burner until Jasmine spies preserve arity. For now I'm working around it with sinon spies but it would be nice to drop that extra dep.

PR #1062, opened 2016-02-29
Bluebird3.0 warnings

### LINKED ISSUE #992: new warnings in bluebird 3.0 and jasmine
First, this isn't really jasmine's fault, so I understand if you don't think it should be fixed here.
This does seem to be the simplest place to fix it, and bluebird promises are popular...
I can make a pull request for it, but don't want to go through the effort if it's likely to be rejected.

Bluebird has, in 3.0, added a warning flagging cases where a promise is created in a promise handler, but none is returned.
There's a more detailed explanation [here](http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-none-were-returned-from-it) .

This warning is triggered whenever promises are handled like this:
```javascript
it("should check the promises result", function (done) {
    returnsAPromise()
    .then(function (res) {
        expect(res).toBe(expectedResult);
    })
    .then(done, done.fail);
});
```

The warning is easy to turn off, but I like the intent behind it so would like to keep it enabled.
But that means that I need to wrap done and done.fail in a function that returns null everywhere instead of ```.then(done, done.fail)```

```javascript
.then(function () {
    done();
    return null;
}, function (err) {
    done.fail(err);
    return null;
})
```

Alternatively, change jasmine so that the done and done.fail functions created in the QueueRunner always return null.

```javascript
  function once(fn) {
    var called = false;
    return function() {
      if (!called) {
        called = true;
        fn();
      }
      return null;
    };
  }
```

```javascript
      next.fail = function() {
        self.fail.apply(null, arguments);
        return next();
      };
```

marneborn:
Resolves #992.

### DIFF
--- a/spec/core/QueueRunnerSpec.js
+++ b/spec/core/QueueRunnerSpec.js
@@ -220,6 +220,22 @@ describe("QueueRunner", function() {
       jasmine.clock().tick(1);
       expect(nextQueueableFn.fn.calls.count()).toEqual(1);
     });
+
+    it("should return a null when you call done", function () {
+      // Bluebird promises want handlers to return anything but undefined to help catch "forgotten returns".
+      // http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-none-were-returned-from-it
+      // When doing ".finally(done, done.fail)" done and done.fail are hanlders so need to return null.
+      var doneReturn,
+        queueableFn = { fn: function(done) {
+          doneReturn = done();
+        } },
+        queueRunner = new jasmineUnderTest.QueueRunner({
+          queueableFns: [queueableFn]
+        });
+
+      queueRunner.execute();
+      expect(doneReturn).toBe(null);
+    });
   });
 
   it("calls exception handlers when an exception is thrown in a fn", function() {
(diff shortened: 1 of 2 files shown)


slackersoft:
It looks like you've introduced a whole new development dependency and integration test for this feature. I would prefer a unit test for this feature instead with some kind of basic description as to why you might need it instead of including a whole library simply for an error it generates.

marneborn:
@slackersoft, yep I should have spent more time figuring out how to test this without including bluebird.
I've redone the failing test to simply check that done returns ```null``` instead of ```undefined```

Here's the new spec in QueueRunnerSpec.js (since the change is in QueueRunner.js)
```javascript
    it("should return a null when you call done", function () {
      // Bluebird promises want handlers to return anything but undefined to help catch "forgotten returns".
      // http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-none-were-returned-from-it
      // When doing ".finally(done, done.fail)" done and done.fail are hanlders so need to return null.
      var doneReturn,
        queueableFn = { fn: function(done) {
          doneReturn = done();
        } },
        queueRunner = new jasmineUnderTest.QueueRunner({
          queueableFns: [queueableFn]
        });

      queueRunner.execute();
      expect(doneReturn).toBe(null);
    });
  });
```

Anything else you'd like me to change?
I rebased and pushed with --force to clean up the commits. Hope that's OK.

PR #1067, opened 2016-03-04
Added support for ES6 sets to toContain and toEqual.

alur:


vbraun:
+1, I just noticed that `expect(new Set([1])).toEqual(new Set([2]))` currently passes

Issue #1016, opened 2016-01-13
There could be a better error message if non-funtion is passed to spy.and.callFake()


jrencz:
I often change from
```
spy.and.returnValue(someExistingVariable);
```

to 

```
spy.and.callFake(() => someVariableThatCanBeAssignedLater);
```

and I sometimes forget to pass a function:

```
// Triggers error
spy.and.callFake(someVariableThatCanBeAssignedLater);
```

The error that is returned does not tell me immediately what have I done wrong:

```
TypeError: 'undefined' is not a function (evaluating 'plan.apply(this, arguments)')
```

There could be a better error message that would hint the user that what has been passed to `callFake` is not a function.

slackersoft:
We would be happy to review a pull request that throws an error if `callFake` is called with a non-function argument.

Thanks for using Jasmine!

PR #1077, opened 2016-03-20
issue #1016 adding a type check on callFake

cwmoo740:
callFake now throws a helpful error if the supplied argument is not a function

Should an additional test be added for this?


slackersoft:
Closing as a duplicate of #1059

PR #1091, opened 2016-04-11
Correctly pretty print objects from other contexts (e.g. iframes) and which do not override toString

### LINKED ISSUE #1087: Pretty-printing objects from a different frame results in "[object Object]"
Attempting to pretty-print an object that came from another frame (or any other situation where `value.toString !== Object.prototype.toString` despite it not actually having a custom toString) results in the normal output of Object.prototype.toString being printed, `"[object Object]"`. This seems to be due to [the well-intentioned check for whether an object has a custom toString](https://github.com/jasmine/jasmine/blob/d0ad9b98cde20845c780a49e8517fdb429fe477d/src/core/PrettyPrinter.js#L33) in order to prefer that over the normal pretty printing behaviour. Specifically, it is the `value.toString !== Object.prototype.toString` portion which causes a false positive.

This could be fixed by speculatively calling value.toString and comparing its output against the normal Object.prototype.toString output. If the output is different, then the value being pretty printed definitely has a custom toString and the output can be used as is.

thatfulvioguy:
Fixes #1087

The issue in question was caused by globals (including prototypes and functions on them) having different identities in different contexts (e.g. iframes and web workers), therefore making `value.toString !== Object.prototype.toString` where value is from another context. The approach taken here is to compare the result of the value's toString with Object.prototype.toString to see if they _really_ are different. If they differ, then the value definitely has a custom toString.

### DIFF
--- a/spec/core/PrettyPrintSpec.js
+++ b/spec/core/PrettyPrintSpec.js
@@ -187,6 +187,15 @@ describe("jasmineUnderTest.pp", function () {
     };
 
     expect(jasmineUnderTest.pp(obj)).toEqual("my toString");
+
+    // Simulate object from another global context (e.g. an iframe or Web Worker) that does not actually have a custom
+    // toString despite obj.toString !== Object.prototype.toString
+    var objFromOtherContext = {
+      foo: 'bar',
+      toString: function () { return Object.prototype.toString.call(this); }
+    };
+
+    expect(jasmineUnderTest.pp(objFromOtherContext)).toEqual("Object({ foo: 'bar', toString: Function })");
   });
 
   it("should handle objects with null prototype", function() {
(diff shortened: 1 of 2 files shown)


thatfulvioguy:
Hang on, it's now misprinting foreign-context arrays because of the `value instanceof Array` check that should not be there. I'll quickly fix that and cover that case in the tests.

thatfulvioguy:
Fixing the printing of foreign-context arrays was trivial. However, writing a test for that without creating another context and ensuring that it also works in IE8 seems to be impossible, as far as I'm aware.

Is this PR sufficient as-is or do I need to do something more creative to test that part?

thatfulvioguy:
Is there anything preventing this PR from being merged?

thatfulvioguy:
Any news on this? We've had to create an internal fork with this PR and we'd rather not have it exist for too long.

PR #1106, opened 2016-05-02
improves incorrect usage warnings for describe

mbildner:
provides legible error message when user fails to provide a callback to a `describe` call.


slackersoft:
I like the thought of getting a better error message when no `function` is passed to a `describe`, but there are a number of issues with this pull request as it stands right now.

1. You have commited your `jasmine.js` file, which is generated and causes merge conflicts
1. There is a bunch of code here that is not related to the declared purpose of the pull request

Please pull out the `RunnableResourceRepository` code into its own pull request and rebase and squash the remaining changes for the error message here and we can take another look at merging this.

Thanks.

slackersoft:
Closing this due to inactivity.

PR #1107, opened 2016-05-03
remove unused code from Env

mbildner:
`runnableLookupTable` is only ever written to, the information in it is never used.

Looks like its behavior ended up being subsumed into the `runnableResources` object.


Gerg:
Resolved by #1129, thanks for the PR!

PR #1114, opened 2016-05-15
Add vulnerability checking.

adam-moss:
Add `snyk` to perform vulnerability checking on node-js dependencies.


Gerg:
I don't think dependency vulnerabilities are an important concern for a testing framework. Please reopen if you have an argument for why we should include this in the Jasmine codebase.

PR #1125, opened 2016-06-01
add comment for external dependencies

klaus-donnert:
Show new users where to place libraries, such jQuery in SpecRunner.html.


Gerg:
As far as Jasmine is concerned, external dependencies are just source files. Loading source files before `boot.js` may interfere with Jasmine's setup.

PR #1137, opened 2016-06-23
throwOnExpectationFailure works with an async function

jbunton-atlassian:
How to reproduce:
* `mkdir demo`
* `npm install jasmine`
* `$(npm bin)/jasmine init`
* Edit `spec/support/jasmine.json` and set `stopSpecOnExpectationFailure=true`
* Run an async test which calls `done.fail()` from within a promise callback

Expected behaviour:
* The test stops immediately
* The failure message is printed to the console

Actual behaviour:
* The test hangs until the timeout expires
* The failure message is printed to the console along with the "Async callback was not invoked within timeout" error

Here's an example spec you can use to see the problem. The change in this PR makes it behave as expected. It also passes the test suite locally in Node and in Firefox.

```
describe('demo', () => {

    beforeEach((done) => {
        Promise.resolve().then(() => {
            console.log('calling done.fail');
            done.fail('this should stop the test immediately, but it does not');
        });
    });

    it('demo', () => {
        console.log('does not get here!');
    });

});
```

Thanks :)

### DIFF
--- a/spec/core/QueueRunnerSpec.js
+++ b/spec/core/QueueRunnerSpec.js
@@ -115,6 +115,30 @@ describe("QueueRunner", function() {
       expect(queueableFn2.fn).toHaveBeenCalled();
     });
 
+    it("throwOnExpectationFailure works with an async function", function() {
+      var asyncFn = { fn: function(done) {
+          setTimeout(function() { done.fail('foo'); }, 100);
+        } },
+        afterFn = jasmine.createSpy('afterFn'),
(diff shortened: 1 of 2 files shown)


jbunton-atlassian:
Hi @Gerg, I see you've been making commits to the repo recently. Sorry to ping you directly but I wasn't sure who else to contact :)

Would you mind having a look at this PR for me please? Is there anything else I should do?

Thanks!

slackersoft:
I think this solution is too specific to `done.fail` where the ultimate problem is something more like #529. We'd like to see a more general solution that would work for any `Error`s thrown in asynchronous code.

Closing this. Thanks for using Jasmine!

PR #1142, opened 2016-06-30
calling fail() with an Error should not create a new Error

arypurnomoz:
When an error has been thrown in an async function there is no way to let jasmine print it stack, the current implementation will throw a new Error which which has a different Error.stack


Gerg:
Hey, a couple issues with this PR:

slackersoft:
Closing. We'll be happy to review a pull request for this functionality that meets our contributing guidelines. Thanks for using Jasmine!

PR #1151, opened 2016-07-15
Upgrade devDependencies glob and shelljs

amavisca:


Gerg:
@amavisca Looks like this broke the node specs. I reverted for now.

amavisca:
Whoops - think I meant to point the devDependency to the npm release of the node runner rather than doing the github reference since it's weird/complicated to point at an unreleased version. I'm slowly trying to untangle the core/npm complexity into a more typical "depend on released versions" workflow.

I'll send up some new PRs/clean this one up today.

PR #1155, opened 2016-07-18
Make `toEqual` pass for arrays with equivalent properties

benchristel:
I think this was a bug—arrays with equivalent properties and contents were treated as different by `eq`.

The `keys` function is inside `eq`'s scope when it doesn't have to be, which created an opportunity for the bug to exist. I think `keys`, `has`, and `isFunction` could be moved out of `eq` with no ill effect. If the project maintainers agree, I'll open another PR that moves those functions.


Gerg:
Thanks for the PR. I'd be cool with moving the functions out.

PR #1157, opened 2016-07-19
Test comparison of objects from different frames

benchristel:
There was some untested code in `eq` that I had a hard time understanding. I think the tests I've added express the desired behavior.


benchristel:
Actually, I'm going to incorporate this into the follow-up task for #1155 (moving the functions in `eq` to a higher scope)

PR #1158, opened 2016-07-19
Move functions in `eq` to a higher scope

benchristel:
This is a follow-up task to #1155. It also backfills some missing tests around comparison of objects with different constructors, and object Objects from different frames.

### DIFF
--- a/spec/core/matchers/matchersUtilSpec.js
+++ b/spec/core/matchers/matchersUtilSpec.js
@@ -100,17 +100,6 @@ describe("matchersUtil", function() {
       expect(jasmineUnderTest.matchersUtil.equals(new One(), new Two())).toBe(false);
     });
 
-    if (typeof document === 'object') {
-      it("passes for equivalent objects from different frames", function() {
-        var iframe = document.createElement('iframe');
-        document.body.appendChild(iframe);
-        console.log(iframe);
-        iframe.contentWindow.eval('window.testObject = {}');
-        expect(jasmineUnderTest.matchersUtil.equals({}, iframe.contentWindow.testObject)).toBe(true);
-        document.body.removeChild(iframe);
-      });
-    }
-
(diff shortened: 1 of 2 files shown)


spec/core/matchers/matchersUtilSpec.js
```diff
@@ -93,6 +93,24 @@ describe("matchersUtil", function() {
...
 
+    it("fails for objects with different constructors", function() {
+      function One() {}
+      function Two() {}
+
+      expect(jasmineUnderTest.matchersUtil.equals(new One(), new Two())).toBe(false);
+    });
+
+    if (typeof document === 'object') {
+      it("passes for equivalent objects from different frames", function() {
```
Gerg: This spec is testing the same behavior as these: https://github.com/benchristel/jasmine/blob/5f27c1f199a9ffe57869e7e9fc99b9abddfca508/spec/core/matchers/matchersUtilSpec.js#L209-L234

I'm cool with also testing this in browsers, but can you move this down with the others (maybe in a shared describe). 

Also we should unify on a testing style for specs that only execute in certain environments.

Gerg:
Thanks for the PR!

In addition to my comment, can you also remove the changes to the built jasmine? I'll build it when I merge: https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md#submitting-a-pull-request

spec/core/matchers/matchersUtilSpec.js
benchristel: Whoops—I missed those other specs because I wasn't running the tests in Node on a regular basis (see: #1160).

benchristel: Also, I leaned towards not calling `it` if there's no `document` rather than just terminating the test early, because if you do the latter you get console warnings about specs that have no assertions.

PR #1163, opened 2016-07-21
toEqual reports detailed comparison of objects and arrays - take 2

benchristel:
This implements #675. It addresses @slackersoft's comments on #1042, which was my first attempt at implementing this feature.

**EDIT: This is now "done" and ready for review.**

~~I'm opening a PR for it because I'd love to get some intermediate feedback on the approach. I keep discovering edge cases that seem to require refactoring within `eq` to enable adding the diff functionality, and want to make sure I'm not going down the wrong path or inadvertently changing behavior. Also, any feedback about the preferred format for these error messages would be immensely helpful at this stage.~~

~~Thanks in advance for any feedback you have time to provide. I'll keep working on this in the meantime since I'm between projects at Pivotal Labs for the next week.~~

### DIFF
--- a/spec/core/matchers/toEqualSpec.js
+++ b/spec/core/matchers/toEqualSpec.js
@@ -1,20 +1,39 @@
 describe("toEqual", function() {
+  "use strict";
+
+  function compareEquals(actual, expected) {
+    var util = jasmineUnderTest.matchersUtil,
+      matcher = jasmineUnderTest.matchers.toEqual(util);
+
+    var result = matcher.compare(actual, expected);
+
+    return result;
+  }
+
   it("delegates to equals function", function() {
     var util = {
-        equals: jasmine.createSpy('delegated-equals').and.returnValue(true)
+        equals: jasmine.createSpy('delegated-equals').and.returnValue(true),
+        buildFailureMessage: function() {
+          return 'does not matter'
+        },
+        DiffBuilder: jasmineUnderTest.matchersUtil.DiffBuilder
       },
       matcher = jasmineUnderTest.matchers.toEqual(util),
       result;
 
     result = matcher.compare(1, 1);
 
-    expect(util.equals).toHaveBeenCalledWith(1, 1, []);
+    expect(util.equals).toHaveBeenCalledWith(1, 1, [], jasmine.anything());
     expect(result.pass).toBe(true);
   });
 
   it("delegates custom equality testers, if present", function() {
     var util = {
-        equals: jasmine.createSpy('delegated-equals').and.returnValue(true)
+        equals: jasmine.createSpy('delegated-equals').and.returnValue(true),
+        buildFailureMessage: function() {
+          return 'does not matter'
+        },
+        DiffBuilder: jasmineUnderTest.matchersUtil.DiffBuilder
       },
       customEqualityTesters = ['a', 'b'],
       matcher = jasmineUnderTest.matchers.toEqual(util, customEqualityTesters),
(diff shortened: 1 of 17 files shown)


benchristel:
@slackersoft, @gerg —it looks like the discussion on #675 never reached a consensus about whether to pretty-print the entire object if there's a deeply-nested mismatch. The code in this PR currently does not (in general) print the whole object.

I considered the idea of adding the pretty-print but greatly limiting the length and depth to which the objects would be printed, by passing `MAX_PRETTY_PRINT_DEPTH` and `MAX_PRETTY_PRINT_ARRAY_LENGTH` as parameters to the printer. Would you be OK with that?

nicbou:
Our team is eagerly waiting for this PR to make it to the release branch. Thanks for your work Ben

jeromecovington:
I know "plus one-ing" is not the most helpful comment to a conversation, but getting detailed comparisons to expected data would really be helpful. This is a real pain point when working with large objects, sometimes deeply nested. So +1. ;-)

centigrade-thomas-becker:
@benchristel even if there is no concensus I would argue that anything is better than the current state. Manually comparing large objects is a real pain currently. I am eagerly looking forward to this feature.

jgrund:
Any update on this? Sorely needed when comparing large objects.

mehdi-d:
Any update on this?

surdu:
How can we beta test this before it's released ?

Tried to npm install this as the jasmine-core in my local jasmine installation, but doesn't appear to work as I still see the old style diff.

Also, will this also work for strings ?

benchristel:
@surdu sorry, I hadn't run `grunt buildDistribution`, which builds the jasmine.js file from the src files. I've done it now, so you should be able to npm install from this branch and see the new diff behavior.

Currently, it does not do anything special for strings. I don't think it would be too hard to add string-diffing, but maybe it should wait for a future PR.

spec/core/matchers/toEqualSpec.js
```diff
@@ -1,28 +1,432 @@
...
+  it("reports mismatches between Sets of different lengths", function() {
+    jasmine.getEnv().requireFunctioningSets();
+
+    var actual = new Set([1, 2]),
+      expected = new Set([2]),
+      message = 'Expected Set( 1, 2 ) to equal Set( 2 ).';
+
+    expect(compareEquals(actual, expected).message).toEqual(message);
+  });
+
```
slackersoft: There are a bunch of `xit`d specs below here. We should either implement them or pull them out as not relevant

src/core/matchers/matchersUtil.js
```diff
@@ -168,25 +205,30 @@ getJasmineRequireObj().matchersUtil = function(j$) {
...
+        diffBuilder.record(a, b);
         return false;
       }
       var iterA = a.values(), iterB = b.values();
       var valA, valB;
       do {
         valA = iterA.next();
         valB = iterB.next();
-        if (!eq(valA.value, valB.value, aStack, bStack, customTesters)) {
+        if (!eq(valA.value, valB.value, aStack, bStack, customTesters, j$.NullDiffBuilder())) {
```
slackersoft: Shouldn't this be the `diffBuilder` local and not grabbing a new `NullDiffBuilder`?

benchristel: Sending in the NullDiffBuilder will cause both Sets to be pretty-printed in the failure message of their containing object. Basically, I was uncertain about how to display set diffs, so I just punted on this one.

Ideally, I'd like something similar to the diff that's displayed for objects when `a` and `b` have different sets of keys. However, the way set equality is implemented right now, the insertion order of items in the set matters. Given that that's the case, we could treat sets like arrays. But that, too, presents problems, because then you'd get output like

```
Expected $[0] to equal 'bar', but got 'baz'
```

when the objects in question are Sets, not Arrays. This output seems a bit misleading since set items can't actually be accessed by index.

spec/core/matchers/toEqualSpec.js
benchristel: :+1:

benchristel: The `xit`d specs have been implemented. The one for custom equality checkers turned into an integration test.

surdu:
Merged! 🎉 

So, did it finally also included big strings in the diff or only big objects ?

slackersoft:
@benchristel It looks like this actually has some failing tests in the various IE browsers. If you, or someone else, have some time to look into the failures (https://travis-ci.org/jasmine/jasmine/builds/176577479) we need to get these fixed.

PR #1169, opened 2016-07-25
Fix the build in ie8

benchristel:
My last PR broke the build. Sorry about that!

Later this week, I might write up some docs about how to get the tests running in ie8 when your dev machine is a mac (for my own reference, here are the pages I want to link to or draw from)

- http://www.digitaltrends.com/computing/how-to-install-and-run-internet-explorer-on-mac-os-x-with-oracle-vm-virtualbox/
- https://www.virtualbox.org/wiki/Downloads
- http://stackoverflow.com/questions/1261975/addressing-localhost-from-a-virtualbox-virtual-machine

### DIFF
--- a/spec/core/matchers/matchersUtilSpec.js
+++ b/spec/core/matchers/matchersUtilSpec.js
@@ -157,13 +157,14 @@ describe("matchersUtil", function() {
     });
 
     describe("when running in a browser", function() {
-      beforeEach(function() {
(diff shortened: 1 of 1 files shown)


Gerg:
Here is a good place to get official windows vms: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

Issue #1174, opened 2016-07-28
noTryCatch option to make debugging easier


craigcomstocks:
When a spec throws an exception and a developer has dev tools open it would be better if the exception is thrown all the way out and causes test execution to stop and the browser to open the source of the exception and begin debugging.

Our current implementation simply avoids the try/catch in QueueRunner attemptSync and attemptAsync functions.

craigcomstocks:
I am investigating whether Env.catchExceptions(false) is suitable for this purpose or can be modified for this purpose.

craigcomstocks:
With catchExceptions, the exception is thrown from a different location than it's source and so with Chrome or IE11 the debugger position lands on QueueRunner.handleException() instead of at the source as it would if there was no try/catch.

slackersoft:
My inclination here would be to solve the problems of correct stack traces and making sure break on errors works correctly, by using a global error handler like `window.onerror` to handle all errors to associate them with the currently running spec. This would also solve #529 for async specs as well. I don't think we want yet another option to configure exception handling in Jasmine.

I'm going to close this in favor of a more general change to handle all errors the same way at a global level that also solves the async case. Thanks for using Jasmine!

craigcomstocks:
Sounds good. Thanks for looking at the bug and PR.

PR #1177, opened 2016-08-05
noTryCatch option to disable try/catch around spec functions

craigcomstocks:
Addressing https://github.com/jasmine/jasmine/issues/1174

Add Env noTryCatch option which allows us to NOT use the try/catch block around executing spec functions in QueueRunner.execute. This allows exceptions to throw immediately and enhances the in-browser debugging experience by placing the debugger at the point of exception instead of at the point of re-throwing as in the case of catchException/QueueRunner.handleException which will lose the context of the exception.

### DIFF
--- a/spec/core/QueueRunnerSpec.js
+++ b/spec/core/QueueRunnerSpec.js
@@ -234,6 +234,26 @@ describe("QueueRunner", function() {
       queueRunner.execute();
       expect(doneReturn).toBe(null);
     });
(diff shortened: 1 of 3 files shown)


slackersoft:
See [my comment](https://github.com/jasmine/jasmine/issues/1174#issuecomment-261619138) on #1174. I don't think we want just another configuration option for this.

Issue #1188, opened 2016-08-31
2.5.0 Regression issue with expect().toHaveBeenCalledWith()


DerFlatulator:
After installing 2.5.0, many of our tests are failing with:
```
PhantomJS 2.1.1 (Windows 8 0.0.0) Some test... FAILED
        TypeError: undefined is not a constructor (evaluating 'allKeys[i].match(/^[0-9]+$/)') in C:/<path>/node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 2988)
        c:/<path>/specFile.js:79:85
```

All the tests I've looked at so far call `toHaveBeenCalledWith`, e.g.:
```js
expect(aSpy).toHaveBeenCalledWith({
  someProperties: "string",
  someOtherProperties: true
})
```

The full function that is throwing an error is:

```js
    function keys(obj, isArray) {
      var allKeys = Object.keys ? Object.keys(obj) :
        (function(o) {
            var keys = [];
            for (var key in o) {
                if (has(o, key)) {
                    keys.push(key);
                }
            }
            return keys;
        })(obj);

      if (!isArray) {
        return allKeys;
      }

      var extraKeys = [];
      for (var i in allKeys) {
        if (!allKeys[i].match(/^[0-9]+$/)) {
          extraKeys.push(allKeys[i]);
        }
      }

      return extraKeys;
    }
```

(edit)

## Temporary Workaround

Lock `jasmine-core` to version `2.4.1` until this is addressed.

VolleMelk:
Same here. The folllowing test fails:

```javascript
it('should equal', function() {
  expect(true).toEqual(true); // ok
  expect([true]).toEqual([true]); // fail
}
```
Fails with
```
PhantomJS 1.9.8 (Windows 8 0.0.0) This should equal FAILED
        TypeError: 'undefined' is not a function (evaluating 'allKeys[i].match(/^[0-9]+$/)')
            at D:/toEqualSpec.js:7
```

SuneRadich:
For now I have downgraded to 2.4.1 and all my tests run again

AlexChesters:
@SuneRadich we've done the same and that solves our failures as well.

gebeuhler:
Am experiencing this issue as well. Will downgrade to 2.4.1.

dmcquillan314:
+1

xenobytezero:
I am also seeing this issue, downgrade has solved it.

[...]

PR #1192, opened 2016-09-01
fix for issue #1188

logankd:
checked for allKeys.length and used for instead of for...in

### DIFF
--- a/src/core/matchers/matchersUtil.js
+++ b/src/core/matchers/matchersUtil.js
@@ -223,10 +223,14 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       }
 
       var extraKeys = [];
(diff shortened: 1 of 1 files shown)


logankd:
I apologize, but I can't take time to figure out how to write a test to make sure this doesn't happen again.

src/core/matchers/matchersUtil.js
```diff
@@ -223,10 +223,14 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       }
 
       var extraKeys = [];
```
bkoteshwarreddy: Do we need to initialize this when allKeys is empty? We can move lines 226-228 before line 224.

PR #1197, opened 2016-09-03
Add env execution complete callback with execution result

bcaudan:
To add a callback after test execution, the common way seems to add a reporter and use the `jasmineDone` callback.

Having a dedicated method to do that would be a little less hacky and could be helpful to fix https://github.com/jasmine/jasmine-npm/issues/90 with https://github.com/jasmine/jasmine-npm/pull/92.

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -1746,4 +1746,95 @@ describe("Env integration", function() {
 
     env.execute();
   });
+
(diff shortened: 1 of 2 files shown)


slackersoft:
We're going to go with the approach of letting a user clear any registered reporters to solve this problem. This way, Jasmine-NPM will register its ConsoleReporter immediately upon creation which can be removed before execution begins.

PR #1208, opened 2016-09-14
Pontential fix for https://github.com/jasmine/jasmine-npm/issues/88

m-a-r-c-e-l-i-n-o:
At first glance it seems like this pr would address this issue: https://github.com/jasmine/jasmine-npm/issues/88 without major concerns, however, I'm not sure if there will be any repercussions to the reporter fallback logic (if there are tests targeting that, we are in the clear). But here's what I do know, [this](https://github.com/jasmine/jasmine-npm/blob/31897dd9450f7e719ce962f23761948803950573/lib/jasmine.js#L177) executes before [this](https://github.com/jasmine/jasmine/blob/master/lib/jasmine-core/jasmine.js#L2025), so the "reporters" length can never be 0. This one flip of the switch was enough to resolve my issue, with my current setup. I suspect that if this is not the right answer, it will empower someone with more knowledge about the system to squash this bug at the root.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2013,7 +2013,7 @@ getJasmineRequireObj().ReportDispatcher = function() {
     this.addReporter = function(reporter) {
       reporters.push(reporter);
     };
-    
+
     this.provideFallbackReporter = function(reporter) {
       fallbackReporter = reporter;
     };
@@ -2022,7 +2022,7 @@ getJasmineRequireObj().ReportDispatcher = function() {
     return this;
 
     function dispatch(method, args) {
-      if (reporters.length === 0 && fallbackReporter !== null) {
+      if (reporters.length === 1 && fallbackReporter !== null) {
           reporters.push(fallbackReporter);
       }
       for (var i = 0; i < reporters.length; i++) {


slackersoft:
This is not a good solution to the issue in jasmine/jasmine-npm#88 as it breaks the functionality for a default reporter in any other use case.

Additionally, the pull request doesn't meet out [contributing guidelines](https://github.com/jasmine/jasmine/blob/master/CONTRIBUTING.md). In particular:

* You have not provided any tests for your functionality
* You have modified the generated `jasmine.js` file
* You haven't actually modified the source files that would cause the functionality to be present in the future.

Closing this.

m-a-r-c-e-l-i-n-o:
@slackersoft, the solution might not be good, but does the statement above ' ... [this](https://github.com/jasmine/jasmine-npm/blob/31897dd9450f7e719ce962f23761948803950573/lib/jasmine.js#L177) executes before [this](https://github.com/jasmine/jasmine/blob/master/lib/jasmine-core/jasmine.js#L2024), so the "reporters" length can never be 0 ... ' hold true, though? In other words, is there ever a case where `dispatch` gets called before` this.addReporter(this.exitCodeReporter)`?

Also my solution " ... breaks the functionality for a default reporter in any other use case ... ", are there no tests currently that would catch this? After all, this existing functionality is suppose to be in theory working and tested, no?

slackersoft:
There are tests for this functionality, but because you didn't modify the reportDispatcher in the `src` directory and only modified `jasmine.js`, the tests didn't run against the code you changed.

There isn't a case where adding the `ExitCodeReporter` in the jasmine-npm package will come after the `dispatch` call, but this is only true for people using that package to run jasmine. At this point none of the other numerous ways to run Jasmine tests will ever add another reporter, especially not the `ExitCodeReporter` in jasmine-npm.

Issue #1175, opened 2016-07-31
beforeAll and afterAll execute asyncronious functions inside xdescribe, unlike their -Each counterparts


koroliov:
The behaviour inside disabled suites (i.e. in the xdescribe function calls) of beforeAll and beforeEach differs considerably, and I suppose this is not an expected behavior.

In the example below, if the `beforeAll` call is changed to `beforeEach` then the suite passes, otherwise it is not.

One can use such constructs to setup a requirejs module before all specs, so this is not an artificial example, it is just simplified.

```
describe('TestBug suite\n', function() {
  'use strict';
  xdescribe('xdecsr\n', function() {
    beforeAll(function(done) {
      setTimeout(function() {
        throw 1;
        done();
      }, 0);
    });
    it('descr', function() {
      expect(1).toEqual(1);
    });
  });
  describe('decsr\n', function() {
    it('descr', function() {
      expect(1).toEqual(1);
    });
  });
});
```

The same issue is with afterAll.

slackersoft:
This definitely looks like a bug in Jasmine. We would be happy to review a pull request that fixes this issue.

slackersoft:
It looks like this broke when `xdescribe` was switched to mark the Suite as pending instead of disabled for better consistency with `xit`. Since nothing will actually set a Suite to `disabled`, all the disabling stuff should be removed and Suites should report that that aren't executable when they are pending.

alex-svyatenko:
It seems that I found a workaround here, while this issue will be fixed.
If you set **"xdescribe"** and **"xbeforeAll"** (x can be replaced by another letter), test will be not executed and have pending status.
An example, result in console :
Pending:
1) <describe name> encountered a declaration exception

1 spec, 0 failures, 1 pending spec 

Thanks,
Alex.

PR #1209, opened 2016-09-14
Fix `TypeError: Cannot read property 'result' of undefined`

colllin:
...via safer property access.  

The previous implementation is throwing errors when `this.children` is empty.  I don't know enough about jasmine to know when that situation occurs, but I believe I should be seeing some kind of test or code error rather than a jasmine error related to `TypeError: Cannot read property 'result' of undefined`.

Console output before this change:
```
return children && children[0].result.status;
                              ^

TypeError: Cannot read property 'result' of undefined
at isAfterAll (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:2337:35)
at Suite.onException (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:2300:8)
at Suite.onException (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:2312:27)
at QueueRunner.onException (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:2419:28)
at onException (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1980:12)
at Timeout._onTimeout (/path/to/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1966:11)
at tryOnTimeout (timers.js:232:11)
at Timer.listOnTimeout (timers.js:202:5)
```

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2334,7 +2334,7 @@ getJasmineRequireObj().Suite = function(j$) {
   };
 
   function isAfterAll(children) {
-    return children && children[0].result.status;
+    return children && children[0] && children[0].result && children[0].result.status;
   }
 
   function isFailure(args) {


Gerg:
Is this PR associated to a Github issue? How can I reproduce this error? I would prefer to solve the underlying issue than add more guards.

In addition:

Gerg:
Closing due to inactivity.

JR-Utily:
I have no clue why this is happening to me, but I have the same issue.
The children variable is an empty array, so it is defined but not children[0].
I'm using jasmine 2.5.2.
I have no knowledge in jasmine code base, so it would be very difficult to me to understand what is happening and why, but I can add breakpoints where you think is efficient to watch the state of the execution.
Here is the stacktrace at the moment of the error.
image

tmschlot:
I'm seeing the same issue in v2.5.3, but my stack trace is a bit different higher up than what @JR-Utily posted (full path omitted)...

```
\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:2348
    return children && children[0].result.status;
                                  ^

TypeError: Cannot read property 'result' of undefined
    at isAfterAll (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:2348:35)
    at Suite.onException (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:2311:8)
    at Suite.onException (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:2323:27)
    at QueueRunner.queueRunnerFactory.onException (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:2572:34)
    at onException (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:1988:12)
    at Timeout._onTimeout (\node_modules\jasmine-core\lib\jasmine-core\jasmine.js:1974:11)
```

I see the issue _nearly_ every time I run my tests but not quite 100% of the time.  I can't tell what's different the few times it has actually succeeded.  

I'm using promises in my tests and the done() callback, and I've not had issues using the pattern in the past.

slackersoft:
This sounds like it it probably cause by Jasmine calling a `beforeAll` or `afterAll` when it shouldn't (like #1175). Calling them incorrectly should be fixed in #1225 which just hasn't been released yet.

If you can run your suite against HEAD it should work, and I'll work on getting a 2.6 release out.

PR #1222, opened 2016-10-18
Require arguments to beforeEach, it, etc, to be actual functions

### LINKED ISSUE #1004: Test results not reported when teardown is undefined
When the teardown function is undefined, the test result is not reported.

Failed tests are missed.

``` javascript
describe('not reported suite', function(){
  afterEach(undefined);
  it('failing test case', function(){
    expect(false).toBe(true);
  });
});
```

In this case it there might be an obvious error, but jasmine should still treat it as an error.

The following example is not that obvious:

``` javascript
describe('not reported suite', function(){
  var tearDown;
  beforeEach(function(){
    tearDown = function(){}
  });

  afterEach(tearDown);

  it('failing test case', function(){
    expect(false).toBe(true);
  })
});
```

The tests itself run and report their failure, however the teardown execution throws an exception that is caught and not handled, while discarding the test results.

See [plnkr](http://plnkr.co/edit/VL84Nu?p=preview).

voithos:
This fixes #1004. Let me know if there are any concerns or ideas for improvement! Note that this _does_ change the public API in a few ways, namely:
- Calling `describe`, `beforeEach`, et al, without providing a function argument will now throw an error instead of silently skipping tests (this is the goal).
- For the edge cases of `xdescribe` and `xit` (since the suite/spec isn't actually ran), I chose to disallow non-function arguments _if they are given_ (that is, pending specs that omit the fn argument completely, like `it('foo')`, still work as intended).

JSHint is green, tests pass in Node and browser environments (haven't tested IE yet).

### DIFF
--- a/src/core/Env.js
+++ b/src/core/Env.js
@@ -292,6 +292,12 @@ getJasmineRequireObj().Env = function(j$) {
       return spyRegistry.spyOnProperty.apply(spyRegistry, arguments);
     };
 
+    var ensureIsFunction = function(fn, caller) {
+      if (!j$.isFunction_(fn)) {
+        throw new Error(caller + ' expects a function argument; received ' + j$.getType_(fn));
+      }
+    };
+
     var suiteFactory = function(description) {
       var suite = new j$.Suite({
         env: self,
@@ -307,6 +313,7 @@ getJasmineRequireObj().Env = function(j$) {
     };
 
     this.describe = function(description, specDefinitions) {
+      ensureIsFunction(specDefinitions, 'describe');
       var suite = suiteFactory(description);
       if (specDefinitions.length > 0) {
         throw new Error('describe does not expect any arguments');
@@ -319,6 +326,7 @@ getJasmineRequireObj().Env = function(j$) {
     };
 
     this.xdescribe = function(description, specDefinitions) {
+      ensureIsFunction(specDefinitions, 'xdescribe');
       var suite = suiteFactory(description);
       suite.pend();
       addSpecsToSuite(suite, specDefinitions);
@@ -328,6 +336,7 @@ getJasmineRequireObj().Env = function(j$) {
     var focusedRunnables = [];
 
     this.fdescribe = function(description, specDefinitions) {
+      ensureIsFunction(specDefinitions, 'fdescribe');
       var suite = suiteFactory(description);
       suite.isFocused = true;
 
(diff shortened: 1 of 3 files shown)


slackersoft:
Thanks for taking a shot at this. Unfortunately, you have introduced a breaking change that we don't want. An `it` without a function argument ha special meaning in Jasmine, it is a quick way to have a pending test that will be implemented later. The rest of the methods don't make sense without a function argument though. 

Please update this PR so that an `it` with no function is still pending. 

Thanks for using Jasmine!

src/core/Env.js
```diff
@@ -288,6 +288,13 @@ getJasmineRequireObj().Env = function(j$) {
       return spyRegistry.spyOn.apply(spyRegistry, arguments);
     };
 
+    var ensureIsFunction = function(fn, caller) {
+      var fnType = Object.prototype.toString.call(fn);
```
slackersoft: You should be able to just use `j$.isFunction_` here instead of writing it again

voithos: Done, thanks. I think it'd be valuable to include the type of the (invalid) argument, so I extract the Object.prototype.toString call out to a helper, j$.getType_.

voithos:
Thanks for the review! I modified it to allow an `it` or `xit` without a function argument (I felt that `xit` made sense - and in fact, there was already a spec that used `xit` without an argument. If a function argument is given, however, the current code will still check that it is of the correct type). Let me know if there are other changes you'd like to see.

voithos:
@slackersoft Do you have any more feedback on this PR? Anything I should change?

voithos:
@slackersoft Friendly ping?

slackersoft:
This looks pretty good now, but I just ran the specs in IE (they don't run for pull requests on travis due to secure environment variables) and I'm seeing a bunch of errors for IE8 that look like:

```
Failures:
          Env #beforeEach throws an error when it receives a non-fn argument

          Expected function to throw an exception with a message matching /beforeEach expects a function argument; received \[object (Undefined|DOMWindow)\]/, but it threw an exception with message 'beforeEach expects a function argument; received [object Object]'.
          No stack trace present.
```

So it looks like IE (at least in version 8) isn't getting the type information we expect.

voithos:
I took a look, and indeed, it seems like IE8 returns `[object Object]` whenever `Object.prototype.toString.apply` is called with either `undefined` or `null`. Thankfully, though, it seems like it returns the correct thing for functions (`[object Function]`), so I believe that the actual type logic will still work out (it's just that the error message will be slightly unclear - but that's still better than no error message!).

I modified the specs to also accept `[object Object]` in the thrown exception message, which should allow the specs to be run on IE8 (although I haven't had a chance to actually do so, since I don't have easy access to IE8). Let me know if anything else isn't working.

PR #1224, opened 2016-10-23
Add support of pend() reason to Suites.

### LINKED ISSUE #1132: Add a way to specify custom reason for disabled suites
Currently we can provide custom reason for pending specs using `pend()`. How do we do the same for suites where we disable it using `xdescribe`? See http://stackoverflow.com/questions/37791690/any-way-to-provide-reason-to-a-disabled-suite-in-jasmine for more details. Thanks!

voithos:
This fixes #1132, mostly by recording the `.pend()` message and making it available for reporters (this PR also updates the `HtmlReporter` to take the message into account). I suppose a next-step would be to update the console reporter in `jasmine-npm` (https://github.com/jasmine/jasmine-npm) to do the same.

JSHint is green, passing in node and browser environments.

### DIFF
--- a/spec/core/SuiteSpec.js
+++ b/spec/core/SuiteSpec.js
@@ -95,6 +95,15 @@ describe("Suite", function() {
     suite.pend();
 
     expect(suite.getResult().status).toBe('pending');
+    expect(suite.getResult().pendingReason).toBe('');
+  });
+
+  it("should set the pendingReason", function() {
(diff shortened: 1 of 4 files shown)


slackersoft:
See [my comment](https://github.com/jasmine/jasmine/issues/1132#issuecomment-264522245) on #1132. The `pend` function on a Suite isn't actually publicly exposed, so I don't think this solves a user problem.

Thanks for taking a look at this.

PR #1225, opened 2016-10-24
Fix bug where before/afterAll were being executed in disabled suites.

### LINKED ISSUE #1175: beforeAll and afterAll execute asyncronious functions inside xdescribe, unlike their -Each counterparts

voithos:
This fixes #1175. This change also removes all the `disabled`-related code from `Suite` objects, including the `.disable()` method (I couldn't find any examples of e.g. emitting deprecation warnings, so I just removed it; let me know if there is something else that should be done here).

Note that this now seems to _not_ run the specs inside a disabled suite, whereas in the past they ran but were marked `'pending'` (was this a symptom of the bug?).

### DIFF
--- a/spec/core/SuiteSpec.js
+++ b/spec/core/SuiteSpec.js
@@ -83,13 +83,6 @@ describe("Suite", function() {
     expect(suite.getResult().status).toBe('finished');
   });
 
-  it("retrieves a result with disabled status", function() {
-    var suite = new jasmineUnderTest.Suite({});
-    suite.disable();
-
-    expect(suite.getResult().status).toBe('disabled');
-  });
-
   it("retrieves a result with pending status", function() {
     var suite = new jasmineUnderTest.Suite({});
     suite.pend();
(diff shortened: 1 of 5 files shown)


slackersoft:
The running of the specs and marking them pending is probably the source of this `beforeAll`/`afterAll` bug. We didn't want to change the reporting of specs within an `xdescribe` from previous behavior, but didn't notice the `beforeAll`s being executed. It looks like the `TreeProcessor` should be handling this case, but it would be good to get an integration spec that proves it.

Thanks.

voithos:
Thanks for the response! It looks like there's already an `it("does not run beforeAlls or afterAlls for a disabled node")` spec in `TreeProcessorSpec.js`, and I had added an `it("shouldn't run before/after functions in disabled suites")` integration spec in `integration/SpecRunningSpec.js` to test it at a higher level. Do these not suffice?

patthiel:
@slackersoft are we any closer to seeing this get merged? It adds a lot of unnecessary execution time to our tests.

patthiel:
Thanks @voithos , @slackersoft !

qualityshepherd:
This is still an issue in v2.5.3, which is included in Protractor 5.1.1. Thanks for fixing this... hope it releases soon!

shavo007:
HI ,

I have this issue aswell. 

jasmine-core is ^2.3.4

What version was this resolved in?

Issue #1050, opened 2016-02-22
Add getMock() functionality to mock classes (to work with ECMA6/ TypeScript classes)


vladimir-ivanov:
Would be useful when working with angular2/typescript to be able to stub all methods in one go.

slackersoft:
Can you describe more what the interface for this would look like and how it would behave? Based on what you've said, I don't understand what is needed.

vladimir-ivanov:
class LifeCycle {

    constructor(anotherRealClassInstance) {
        this.dependency = anotherRealClassInstance;
    }

    init() {
        //body here
    }

    change() {
        //body here
    }

    destroy() {
        //body here
    }
}


class LifeCycleConsumer {

    constructor(lifeCycleInstance) {
        lifeCycleInstance.init();
    }
}

const lifeCycleMock = jasmine.mock(LifeCycle) //or jasmine.getMock(LifeCycle);

const lifeCycleConsumer = new LifeCycleConsumer(lifeCycleMock);


lifeCycleMock.expects("init").atLeast(1);
lifeCycleMock.onInit.verify();
`

vladimir-ivanov:
an easy shortcut to spyOn / stub all dependency methods

kapke:
I wrote some time ago simple piece of code resolving that problem. The idea was as follows: 
1. Fiind all method names in prototype chain starting from given constructor(s) - I had some kind of mixins in project so I needed to provide several constructors. 
2. Deduplicate them
3. return `jasmine.createSpyObj(name, methods)`

I that idea fits you I can start new branch with that code under `j$.getMock`.

vladimir-ivanov:
This works for me and gives me what I need, would be nice to be part of jasmine?

      
let lifeCycleMock = getMock(LifeCycle);

function getMock(mockedClass) {
            Object.getOwnPropertyNames(mockedClass.prototype).forEach(m => spyOn(mockedClass.prototype, m));

            return new mockedClass();
        };

lifeCycleMock.init.and.callFake(() => "I am fake");

new LifeCycleConsumer(lifeCycleMock);

expect(lifeCycleMock.init.calls.count()).toEqual(1)

kapke:
The code I've mentioned above is here:
https://github.com/skygate/CodeSamples/blob/master/js/pretender/pretender.js

As it creates jasmine mock object we don't call any of original methods (in some not so rare cases it would cause very strange behaviour).

vladimir-ivanov:
[...]

PR #1230, opened 2016-10-26
Small enhancement to let a user pass an ES6 class into Jasmine and ha…

drtimwright:
Hi Jasmine folks,

Here's a little enhancement for you. Something I miss coming from the Java world is the ability to give Jasmine a class (I use typescript and ES6) and have Jasmine automatically create an object with all the method in the class spied on. This means that I can more easily create spy objects and know that the methods have the right names (refactoring support often misses changing method names when you use jasmine.createSpyObj mechanism).

So here's a pull request that adds that functionality.

Thanks,

Tim

### DIFF
--- a/spec/core/SpySpec.js
+++ b/spec/core/SpySpec.js
@@ -124,4 +124,34 @@ describe('Spies', function () {
       }).toThrow("createSpyObj requires a non-empty array or object of method names to create spies for");
     });
   });
+
+  describe("createSpyForClass", function() {
+
+    // transpiled Typescript class using tsc targeting ES5
+    var TscClassTranspiledToES5 = (function () {
+      function TestClass() {
+      }
+      TestClass.prototype.testMethod1 = function () {
+        return "";
+      };
+      TestClass.prototype.testMethod2 = function () {
+        return "";
+      };
+      return TestClass;
+    }());
+
+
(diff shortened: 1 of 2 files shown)


slackersoft:
I don't think that class based spying is something Jasmine needs to support at this time. See discussion  [here](https://github.com/jasmine/jasmine/issues/1050#issuecomment-259582143) as well.

drtimwright:
I guess that is life. So I've created an npm module with roughly the same code.

https://www.npmjs.com/package/jasmine-es6-spies

ShinDarth:
@slackersoft any second thoughts about this? It would be very useful...

drtimwright:
@ShinDarth - I ended up publishing this as an npm module. https://www.npmjs.com/package/jasmine-es6-spies or on github at https://github.com/drtimwright/jasmine-es6-spies

FrancescoBorzi:
@slackersoft I still think such a utility method would be very useful. Any second thoughts about it?

slackersoft:
Given the presence of `spyOnAllFunctions` from #1581 and the lack of runtime knowledge of what is or isn't a `class`, I don't think this is something that is likely to make it into Jasmine. The biggest problem is that Jasmine is entirely ES5 and that most runtimes really are too, so the detection of "class-ness" is hard. The `spyOnAllFunctions` functionality is also new since this issue was created and seems like it should generally cover most of the functionality desired here.

PR #1242, opened 2016-12-12
[1st contrib] Add toHaveBeenCalledBefore matcher

DamienCassou:
This matcher makes sure the 'actual' spy is called before the 'expected' spy.

This code has been written in cooperation with @BenjaminVanRyseghem.

This is my first contribution to jasmine. Please double check everything :-).

### DIFF
--- a/spec/core/matchers/toHaveBeenCalledBeforeSpec.js
+++ b/spec/core/matchers/toHaveBeenCalledBeforeSpec.js
@@ -1,0 +1,69 @@
+describe("toHaveBeenCalledBefore", function() {
+  it("throws an exception when the actual is not a spy", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        fn = function() {},
+        secondSpy = jasmineUnderTest.createSpy('second spy');
+
+    expect(function() { matcher.compare(fn, secondSpy) }).toThrowError(Error, /Expected a spy, but got Function./);
+  });
+
+  it("throws an exception when the expected is not a spy", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        firstSpy = jasmineUnderTest.createSpy('first spy'),
+        fn = function() {};
+
+    expect(function() { matcher.compare(firstSpy, fn) }).toThrowError(Error, /Expected a spy, but got Function./);
+  });
+
+  it("fails when the actual was not called", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        firstSpy = jasmineUnderTest.createSpy('first spy'),
+        secondSpy = jasmineUnderTest.createSpy('second spy');
+
+    secondSpy();
+
+    result = matcher.compare(firstSpy, secondSpy);
+    expect(result.pass).toBe(false);
+    expect(result.message).toMatch(/Expected spy first spy to have been called./);
+  });
+
+  it("fails when the expected was not called", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        firstSpy = jasmineUnderTest.createSpy('first spy'),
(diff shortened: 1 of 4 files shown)


DamienCassou:
Would it be possible to get a review on this one please?

spec/core/matchers/toHaveBeenCalledBeforeSpec.js
```diff
@@ -0,0 +1,77 @@
...
+
+  it("throws an exception when the expected is not a spy", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        firstSpy = jasmineUnderTest.createSpy('first spy'),
+        fn = function() {};
+
+    expect(function() { matcher.compare(firstSpy, fn) }).toThrowError(Error, /Expected a spy, but got Function./);
+  });
+
+  it("throws an exception when the actual was not called", function() {
```
slackersoft: This should just be a failure case not an exception

spec/core/matchers/toHaveBeenCalledBeforeSpec.js
```diff
@@ -0,0 +1,77 @@
...
+    var matcher = jasmineUnderTest.matchers.toHaveBeenCalledBefore(),
+        firstSpy = jasmineUnderTest.createSpy('first spy'),
+        secondSpy = jasmineUnderTest.createSpy('second spy');
+
+    secondSpy();
+
+    expect(function() { matcher.compare(firstSpy, secondSpy) }).toThrowError(Error, /Expected spy first spy to have been called./);
+  });
+
+  it("throws an exception when the actual was not called", function() {
```
slackersoft: This should also be a failure case, and should note that it is for the `expected` and not the `actual`

spec/core/matchers/toHaveBeenCalledBeforeSpec.js
slackersoft: Also include in the error message which spy was not called

DamienCassou:
I took care of your feedback. Do you usually prefer separated specs for `result.pass` and `result.message` (such as in `has a custom message on failure that tells second spy got called first` and `fails when the actual is called after the expected`) or a single spec (such as `fails when the expected was not called`)?

slackersoft:
A single spec that checks both whether the check passed and what the message is for that case. Remember to also confirm that the message is right even when `pass` is `true` since that is the message that will be used for `.not`

src/core/matchers/toHaveBeenCalledBefore.js
```diff
@@ -0,0 +1,38 @@
...
+        if (!latterSpy.calls.count()) {
+		  result.message = 'Expected spy ' +  latterSpy.and.identity() + ' to have been called.';
+		  return result;
+        }
+
+		result.pass = firstSpy.calls.mostRecent().invocationOrder < latterSpy.calls.mostRecent().invocationOrder;
+
+        if (!result.pass) {
+          result.message = 'Expected ' + firstSpy.and.identity() + ' to have been called before ' + latterSpy.and.identity();
+        }
```
slackersoft: We should also set a message when `result.pass` is `true` because Jasmine will need to use that message if someone uses `.not.toHaveBeenCalledBefore`

DamienCassou:
I took care of all your comments.

DamienCassou:
Any news?

slackersoft:
That all looks good, but I've just realized a potential quirk around multiple calls. Given two spies `spy1` and `spy2`, when they are called like this:

```javascript
spy1();
spy2();
spy1();
```

or

```javascript
spy2();
spy1();
spy2();
```

I think we want to have the matcher fail (not `throw` though) with a message about interleaved calls.

DamienCassou:
@slackersoft: Here is a new version taking care of your feedback. I rebased the 1st commit to latest master but didn't change its content. Instead, I took care of your feedback in a fixup commit so you can have a look at what changed. Please `git autosquash` before merging or ask me to.

DamienCassou:
any news?

PR #1268, opened 2017-01-31
Issue 1255: Have jasmine.any(Object) not pass for null

### LINKED ISSUE #1255: expect(null).toEqual(jasmine.any(Object)) passes
**version**
jasmine v2.5.3
jasmine-core v2.5.2

**environment**
node v6.9.4
npm v3.10.10

**running Jasmine**
from cli using the `jasmine` command to run my tests

I'm expecting this to fail, but it doesn't.
How does `null` equal an Object?

`expect(null).toEqual(jasmine.any(Object));`

JamieMason:
This pull request aims to:

+ Fix #1255 by ensuring `jasmine.any(Object)` returns `false` for `null`.
+ Update npm dependencies.
+ Use npm scripts and remove the need for contributors to install `grunt` globally, also giving control to the project over specifically which version of `grunt` is used by contributors.

Hope this helps, thanks.

/cc @lmj0011

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -78,13 +78,13 @@ To install the Ruby dependencies, you will need Ruby, Rubygems, and Bundler avai
 
 ...should get you to the point that `bundle` can install everything.
 
-To install the Node dependencies, you will need Node.js, Npm, and [Grunt](http://gruntjs.com/), the [grunt-cli](https://github.com/gruntjs/grunt-cli) and ensure that `grunt` is on your path.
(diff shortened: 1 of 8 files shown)


slackersoft:
Please split the npm packaging and script changes out from the null checking stuff so they can be merged separately.

I'm not sure if we can change this behavior for `any` right now, since that has a potential to break existing suites. That said, I think for 3.0 (when we can break things :), we'll want to look at the provided asymmetric matchers to ensure we are providing the right options out of the box.

I'm going to close this PR, but leave the original issue open. I would be happy to review a new PR with just the NPM stuff. Thanks for using Jasmine!

Issue #923, opened 2015-09-11
How to use async/await with Jasmine?


donaldpipowitch:
Hi, how can I use async/await with Jasmine? Is this supported somehow? I'd like to do 

````
it('should run test', async function() {
    browser.get('https://duckduckgo.com/');
    $('#search_form_input_homepage').sendKeys('jasmine');
    $('#search_button_homepage').click();

    let title = await browser.getTitle();
    expect(title).toBe('foo');
  });
````

Instead of 

````
  it('should run test', function(done) {
    browser.get('https://duckduckgo.com/');
    $('#search_form_input_homepage').sendKeys('jasmine');
    $('#search_button_homepage').click();

    browser.getTitle().then(function(title) {
      expect(title).toBe('foo');
      done();
    });
  });
````

Do I need to do that manually like [here](https://github.com/webdriverio/webdriverio/blob/master/lib/frameworks/jasmine.js#L116)?

donaldpipowitch:
This is a modified version from [`runInGenerator`](https://github.com/webdriverio/webdriverio/blob/master/lib/frameworks/jasmine.js#L204) which runs my first example using async/await with `babel-node`:

````
// asyncify
[ 'it', 'beforeEach', 'beforeAll', 'afterEach', 'afterAll' ].forEach(fnName => {
  var origFn = global[fnName],
    specTitle, specFn, specTimeout;

  var runSpec = function(specTitle, specFn) {
    return origFn.call(null, specTitle, function(done) {
      specFn().then(done.bind(null, null), done.fail.bind(null));
    });
  };

  var runHook = function(specFn, specTimeout) {
    return origFn.call(null, function(done) {
      specFn().then(done.bind(null, null), done.fail.bind(null));
    }, specTimeout);
  };

  global[fnName] = function() {
    var args = Array.prototype.slice.call(arguments);

    if(fnName === 'it') {
      specTitle = args[0];
      specFn = args[1];

      // check if 'regeneratorRuntime.async' is used
      if(specFn.toString().includes('regeneratorRuntime.async')) {
        return runSpec(specTitle, specFn);
      } else {
        return origFn.call(null, specTitle, specFn);
      }
    }

    specFn = args[0];
    specTimeout = args[1];

    // check if 'regeneratorRuntime.async' is used
    if(specFn.toString().includes('regeneratorRuntime.async')) {
      return runHook(specFn, specTimeout);
    } else {
      return origFn.call(null, specFn, specTimeout);
    }
  };
});
[...]

Issue #681, opened 2014-09-27
Allow test to return Promise as an alternative to the done callback


analog-nico:
As proposed in issue #567 Jasmine should allow a test to return a Promise as an alternative to using the `done` callback for async testing.

A test would not specify the `done` callback as its parameter and would return a Promise:

``` js
it('should pass', function () {
  return new Promise(function (resolve, reject) {
    resolve('value');
  });
});

it('should fail', function () {
  return new Promise(function (resolve, reject) {
    reject(new Error('reason'));
  });
});
```

mjackson:
It would be really, really awesome if this worked. :(

OliverJAsh:
IMO this is enough reason to use Mocha instead of Jasmine. :+1: to this change.

piuccio:
@OliverJAsh agree

wmadden:
It'd really be great to see support for Promise return values in Jasmine. Using Promises would mean that Jasmine can internally take advantage of the advances entering the language in ES6, like the `async` function.

slackersoft:
Given that jasmine still supports IE8, we won't be able to actually use any of the other new features that aren't supported by the older browsers we still want to work with.

wmadden:
@slackersoft: you don't need any explicit dependency on a Promise implementation to add this behavior.

@bilalq provided sample code that would work. In browsers with no implementation of Promises it'd simply not return a "thenable" and have no effect.

wmadden:
Sorry, I think I misunderstood your last comment. You're right, you won't be able to take advantage of the new JS features in the output code, but you'd still be able to use them with a compiler like Babel.

At any rate the biggest advantage would be to the users of Jasmine, rather than the authors.

joezimjs:
http://www.sitepoint.com/promises-in-javascript-unit-tests-the-definitive-guide/
This seems a lot nicer than using `done` and `done.fail`. The reasoning that I saw for not supporting this sounds like you are just too lazy to add it, though I guess some sort of plugin or library that can handle this and is compatible with Jasmine would work just fine as well.

joezimjs:
Anyone who wants some better Promise compatibility:
https://www.npmjs.com/package/jasmine-promise-matchers

This is something like the [chai-as-promised](https://github.com/domenic/chai-as-promised/) as mentioned in the article that I linked in my previous comment. This still isn't quite as nice as being able to return a promise, but it's a step in the right direction.

matthewjh:
[...]

PR #1270, opened 2017-02-01
[QueueRunner] Support "thenables" returned by tests

aleclarson:
This commit allows `done` or `done.fail` to be called by a returned promise/thenable.

This reduces the following boilerplate:
- `promise.then(done)`
- `promise.catch(done.fail)`

Includes **tests** for resolved promises and rejected promises.
Should I include one for pending promises, too?

### Example

```js
it(() => {
  return fetch('http://www.google.com')
    .then(res => expect(res).toBe(...));
});
```

- If the returned thenable is resolved, the `done` function is called.
- If the returned thenable is rejected, the `done.fail` function is called.

### Failure tests

To test failure cases, you can `catch` the error and call `expect(error.message)`.

```js
it(() => {
  return fetch('bad url')
    .catch(error => expect(error.message).toBe(...));
});
```

### DIFF
--- a/spec/core/QueueRunnerSpec.js
+++ b/spec/core/QueueRunnerSpec.js
@@ -234,6 +234,52 @@ describe("QueueRunner", function() {
       queueRunner.execute();
       expect(doneReturn).toBe(null);
     });
+
+    it("supports promises (and thenables) returned by queued functions", function(done) {
+      var queueableFn1 = { fn: function() {
+          return Promise.resolve();
+        } },
+        queueableFn2 = { fn: jasmine.createSpy('fn2') },
+        completeCallback = jasmine.createSpy('completeCallback'),
+        queueRunner = new jasmineUnderTest.QueueRunner({
+          queueableFns: [queueableFn1, queueableFn2],
+          onComplete: completeCallback
+        });
+
+      queueRunner.execute();
+
+      expect(completeCallback).not.toHaveBeenCalled();
+      expect(queueableFn2.fn).not.toHaveBeenCalled();
+
+      setImmediate(function() {
+        expect(completeCallback).toHaveBeenCalled();
+        expect(queueableFn2.fn).toHaveBeenCalled();
+        done();
+      });
+    });
+
+    it("calls 'done.fail' when a promise is rejected", function(done) {
+      var error, queueableFn1 = { fn: function() {
+          return Promise.reject(error = new Error('foo'));
+        } },
+        queueableFn2 = { fn: jasmine.createSpy('fn2') },
+        failFn = jasmine.createSpy('fail'),
+        queueRunner = new jasmineUnderTest.QueueRunner({
+          queueableFns: [queueableFn1, queueableFn2],
+          fail: failFn
+        });
+
+      queueRunner.execute();
+
+      expect(failFn).not.toHaveBeenCalled();
+      expect(queueableFn2.fn).not.toHaveBeenCalled();
+
+      setImmediate(function() {
+        expect(failFn).toHaveBeenCalledWith(error);
+        expect(queueableFn2.fn).toHaveBeenCalled();
+        done();
+      });
+    });
   });
 
   it("calls exception handlers when an exception is thrown in a fn", function() {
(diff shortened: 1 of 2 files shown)


aleclarson:
Am I not allowed to use `setImmediate` in the tests?

aleclarson:
Fixed an edge case where the `promise` variable was being carried over to the next iteration, resulting in an error being thrown `TypeError: promise.then is not a function`.

Let me know if there are other issues. 👍

BenoitZugmeyer:
Disclaimer: I'm not related to Jasmine in any way, just an enthusiastic user.

This feature is quite popular and has been discussed and PR'd many times (PR #202, #681, PR #696, #923, PR #1029), and has always been rejected with the following workarounds:

* You can just pass `done` and `done.fail` as callbacks to `.then`
* If you are into coroutines, you can use [jasmine-co](https://github.com/gradecam/jasmine-co)
* You can use [jasmine-promises](https://github.com/matthewjh/jasmine-promises)

In my opinion, it would be a great addition to Jasmine:
* Most other popular test runners (mocha, jest, ava, ...) are supporting asynchronous tests via promises out of the box. It seems like a new consensus, like the `done()` callback a few years ago. It would help migrations from one test runner to another, and build tools/helpers compatible with multiple test runners.
* Coroutines were alright a few years ago, but now that `async/await` functions are standardized and supported by Node (last stable version) and some browsers out of the box, there is not much reason to use them anymore. With support for tests returning promises, jasmine would support async functions without any extra code.
* The jasmine-promises doesn't support nodejs (at the moment) and seems a bit abandonned to be honnest (nodejs support is on the master branch, but no release has been done for six months).
* Adding promises support is straightforward and wouldn't add much code, and no extra dependency.

As a side note, @aleclarson : your PR breaks the CI build because of lints reported by jshint (https://travis-ci.org/jasmine/jasmine/jobs/202125138#L380)

aleclarson:
@BenoitZugmeyer Thanks for the heads up! All fixed.

I'll squash the commits once a maintainer speaks up about any nits.

spec/core/QueueRunnerSpec.js
```diff
@@ -234,6 +234,52 @@ describe("QueueRunner", function() {
       queueRunner.execute();
       expect(doneReturn).toBe(null);
     });
+
+    it("supports promises (and thenables) returned by queued functions", function(done) {
+      var queueableFn1 = { fn: function() {
+          return Promise.resolve();
```
slackersoft: Jasmine's tests need to run in IE 8+ and PhantomJS (as well as Edge, Chrome, Firefox, and Node.js) so we can't use bare `Promise` directly in specs for browsers that don't support them.

spec/core/QueueRunnerSpec.js
```diff
@@ -234,6 +234,52 @@ describe("QueueRunner", function() {
...
+          queueableFns: [queueableFn1, queueableFn2],
+          onComplete: completeCallback
+        });
+
+      queueRunner.execute();
+
+      expect(completeCallback).not.toHaveBeenCalled();
+      expect(queueableFn2.fn).not.toHaveBeenCalled();
+
+      setImmediate(function() {
```
slackersoft: Similarly, `setImmediate` isn't available in all browsers that Jasmine supports and as such can't be used in Jasmine's tests.

spec/core/QueueRunnerSpec.js
```diff
@@ -234,6 +234,52 @@ describe("QueueRunner", function() {
...
+        expect(completeCallback).toHaveBeenCalled();
+        expect(queueableFn2.fn).toHaveBeenCalled();
+        done();
+      });
+    });
+
+    it("calls 'done.fail' when a promise is rejected", function(done) {
+      var error, queueableFn1 = { fn: function() {
+          return Promise.reject(error = new Error('foo'));
```
slackersoft: Native `Promise` as above

spec/core/QueueRunnerSpec.js
```diff
@@ -234,6 +234,52 @@ describe("QueueRunner", function() {
...
+          queueableFns: [queueableFn1, queueableFn2],
+          fail: failFn
+        });
+
+      queueRunner.execute();
+
+      expect(failFn).not.toHaveBeenCalled();
+      expect(queueableFn2.fn).not.toHaveBeenCalled();
+
+      setImmediate(function() {
```
slackersoft: `setImmediate` as above

Issue #1252, opened 2017-01-12
toThrowError with <iframe> reports incorrectly


anseki:
- jasmine-standalone-2.5.2 in browser

Hi, thank you for the great framework.

When an app that runs in `<iframe>` is tested, `toThrowError` reports incorrectly. It says "Expected function to throw an Error, but it threw...".
Jasmine uses `Error` constructor in current window, but it was made by `Error` constructor in child window.

In this code:
https://github.com/jasmine/jasmine/blob/8624a52ee0b6f13b3b608ea6417ccc02257c5412/src/core/matchers/toThrowError.js#L31

```js
if (!(thrown instanceof Error)) {
```

It checks with the `Error` in current window always. But the `thrown` was made in different `global` (namespace).

For example:

- `app.html`

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">

  <script>

    function methodA() {
      return 'foo';
    }

    function methodB() {
      throw new Error('bar');
    }

  </script>

</head>

<body>
  app.html
</body>
</html>
```

- Spec

```js
describe('App running in <iframe>', function() {

  function methodC() {
    throw new Error('baz');
  }

  var childWindow;

  beforeAll(function(done) {
    // Prepare <iframe>
    var iframe = document.body.insertBefore(document.createElement('iframe'), document.body.firstChild)
    iframe.addEventListener('load', function() {
      childWindow = iframe.contentWindow;
      // self.Error = childWindow.Error;
      done();
    });
    iframe.src = 'app.html';
  });

  it('should return foo', function() {
    expect(childWindow.methodA()).toBe('foo');
  });

  it('should throw an error', function() {
    expect(function() {
      childWindow.methodB();
    }).toThrowError('bar');
  });

  it('(test in current window)', function() {
    expect(function() {
      methodC();
    }).toThrowError('baz');
  });
});
```

The second spec fails.

I think, this line:

```js
if (!(thrown instanceof Error)) {
```

should be:

```js
if (!(thrown instanceof (thrown.constructor.constructor('return this')()).Error)) {
```

This gets `Error` from the same `global` as `thrown`.

slackersoft:
We would be happy to review a pull request that allows `toThrowError` to correctly detect `Error`s that come from different frames correctly.

anseki:
@slackersoft,
Ok, I will send a PR later.

PR #1275, opened 2017-02-10
toThrowError with another global reports incorrectly #1252

anseki:
Check the thrown error and its constructor that should be `Error`.
The constructor might not be `Error` in current `global`, it might be child window or another window e.g. Electron.

### DIFF
--- a/spec/core/matchers/toThrowErrorSpec.js
+++ b/spec/core/matchers/toThrowErrorSpec.js
@@ -69,8 +69,31 @@ describe("toThrowError", function() {
     return typeof document === 'undefined'
   }
 
+  function isRunningInPhantomJS(minVersion, maxVersion) {
+    if (!window.callPhantom) {
+      return false;
+    }
+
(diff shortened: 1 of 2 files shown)


slackersoft:
It looks like your specs are failing in phantomjs on Travis. Please take a look and get that to green.

Thanks for using Jasmine!

anseki:
I guessed that you are using PhantomJS v1.9, right?
PhantomJS v1.9 has multiple bugs about `iframe` (https://github.com/ariya/phantomjs/issues?q=is%3Aissue+iframe+label%3ABug), and it can not get `global` by `return this` in `iframe`.
That was fixed in v2.0.

I think that v2.0+ should be used,
Anyway, I updated the spec, it skips the spec in PhantomJS v1.9-. Since this is a problem about only PhantomJS v1.9- not Jasmine, I think that it is OK to skip it.
I confirmed that it passed in both PhantomJS v1.9 and v2+.

PR #1289, opened 2017-03-06
create env.unspy

### LINKED ISSUE #236: Spy Restore()
This idea comes from ```sinon.js``` where you can setup a spy then later on restore it back to the original method in the case you no longer need or want the spy.

This works well when having a spy setup that is used in most cases but in the rare case it isn't, it's more useful to set one up then for the special case remove it.

**Note:** this does not clean up after itself for ```removeAllSpies``` but doesn't have any bad side effects.

```javascript
jasmine.Spy.prototype.restore = function() {
  this.baseObj[this.methodName] = this.originalValue;
};
```

```javascript
describe('Spies', function() {
  it('should restore spied function', function() {
    var originalFunctionWasCalled = false;
    var TestCase = {
      someFunction: function() {
        originalFunctionWasCalled = true;
        return 'return value from original function';
      };
    };
    this.spyOn(TestClass, 'someFunction');

    TestClass.someFunction.restore();
    expect(TestClass.someFunction()).toEqual('return value from original function');
    expect(originalFunctionWasCalled).toEqual(true);
  });
});
```

### LINKED ISSUE #683: Restore behavior for spy
I think it would be a good to restore a spy to its original value (not talking about reset). 
Lets say I spy on an function but later on I want to remove my spy , just for clean up purpose.
This would help different parts of the test suites to create their own spies on certain global functions.

It has been done in `sinonJs` library. http://sinonjs.org/docs/#spies

### LINKED ISSUE #975: restore method for a spy
I had a look at two following issues, both of them are closed, however I have an example which I believe would benefit from having a restore method (which would bring it back to its original fn).

https://github.com/jasmine/jasmine/issues/683
https://github.com/jasmine/jasmine/issues/236

Please have a look at the following code and comments (assume it's requirejs + marionette):

```javascript
describe("MyController", function () {
    "use strict";

    var MyController;

    beforeAll(function (done) {
        require(["path/to/my.controller"], function (Controller) {
            MyController = Controller;
            done();
        });
    });

    it("should be defined", function () {
        expect(MyController).toBeDefined();
    });

    describe("instance", function () {

        var controller;

        beforeEach(function () {
            controller = new MyController();
            // assume that the controller uses
            // this.listenTo to register an event handler
        });

        afterEach(function () {
            controller.destroy();
            // this destroy unbinds registered event handlers
        });

        it("should be defined", function () {
            expect(controller).toBeDefined();
        });

        describe("method", function () {

            describe("onProcessError", function () {

                it("should call destroy", function () {
                    spyOn(controller, "destroy").and.callFake(function () {
                        console.log("spy called"); 
                        // triggers twice
                        // so the registered handlers are not unbinded
                    });
                    controller.onProcessError();
                    expect(controller.destroy).toHaveBeenCalled();
                    // controller.destroy.restore(); 
                    // it would cause the console.log to trigger only once
                    // and the handlers would be removed as usual
                    // but it's quite fragile, you might forget about it
                });

            });

        });

    });
    
});

```

I can see two workarounds of the problem described above:
- callThrough (which would cause destroy to be called twice in the following example, I believe it's unnecessary and might cause problems if internal implementation of given library does throw an error in such case),
- rearrange describe/it statements and create an instance in one it statement and use callThrough to destroy it as usual

Both of these unfortunately do not fit the way I'm arranging the specs. I wasn't able to find any other solution despite going through the docs. I could use sinon, but I'd rather not add any additional  dependencies, jasmine is great enough and only this edge case bothers me. Do you have any suggestions?

### LINKED ISSUE #1281: Allow to manually delete/destroy a spy
Following the discussion on Google Groups: https://groups.google.com/forum/#!topic/jasmine-js/0RXvFo707EQ, I think my usecase is valid.

Jasmine does not allow to manually delete/destroy a spy and this is limiting.

```JavaScript
function signIn(done) {
  ...
  // Fails the 2x time with "<spyOn> : bar has already been spied upon"
  const mySpy = spyOn(foo, 'bar').and.callThrough();
  ...
  expect(foo.bar).toHaveBeenCalledTimes(1);

  // Jasmine missing feature:
  //mySpy.destroy();
}

function signOut(done) {
  ...
}

it('should be able to sign in/sign out multiple times', done => {
  signIn(() => {
    signOut(() => {
      signIn(() => { // Fails with "<spyOn> : bar has already been spied upon"
        done();
      });
    });
  });
});
```

signIn, signOut... are bricks (i.e simple functions - not Jasmine tests) that are reused to write complex Jasmine test scenarios like:
- signIn, doSomething1, signOut
- signIn, doSomething1, doSomething2, doSomething3, signOut
- signIn, signOut, signIn, signOut, signIn, signOut

Remarks:
- Currently the second signIn call fails with `<spyOn> : bar has already been spied upon`
- I cannot make spyOn global otherwise error `Spies must be created in a before function or a spec`
- Adding `beforeEach(()=> { spyOn(foo, 'bar').and.callThrough(); })` is not the solution because `expect(foo.bar).toHaveBeenCalledTimes(1)` will fail with the 2x signIn call (because `.toHaveBeenCalledTimes(2)`)

UziTech:
This will save the `object` and `methodName` or `propertyName` along with `restoreObjectToOriginalState` in the `currentSpies` array in the `SpyRegistry`.

A `env.unspy` method will allow the user to search the `currentSpies` array and execute the `restoreObjectToOriginalState` method for the `object` and `methodName` or `propertyName` provided

fixes #1281
fixes #236
fixes #342
fixes #683
fixes #975


slackersoft:
I'm not sure this is a feature that Jasmine needs right now. I would be happy to hear more discussion in #1281 about functionality that requires a user to remove a spy partway through a spec.

UziTech:
i was debating on `spyOff` instead of `unspy`

maybe something to think about.

mgol:
In that case, `spyOff` sounds better as it refers directly to `spyOn`.

UziTech:
I'm going to close this as it seems to not be wanted by the maintainers.

If anyone wants this functionality I created an npm package [`jasmine-unspy`](https://www.npmjs.com/package/jasmine-unspy) that adds a method `jasmine.unspy`

PR #1300, opened 2017-03-26
Added infinity methods, with tests

### LINKED ISSUE #1294: What about including Infinity methods?
I am thinking about two methods like ```toBeInfinity()```, ```toBeNegativeInfinity()```.
Of course one could argue that we could work around this issue with ```toEqual(Infinity)```, but then I would reply "what is then the purpose of ```toBeNaN()``` method?".

For a visual example, you can refer to https://github.com/mhevery/jasmine-node/pull/415/files.

toubou91:
Fixes #1294


toubou91:
Fixes #1294

PR #1304, opened 2017-03-31
Implement "TODO" in toEqual.js

melon-programmer:
The TODO is "only set error message if test fails". When the test fails,
result.pass will be false. By adding a check on result.pass, we ensure
message is only added when the test fails.


slackersoft:
There are a couple of issues with this pull request as it stands right now:

* There are no tests provided. Jasmine is a testing framework and we don't want to lose functionality because of missing tests.
* You have a number of commits for not many changes. Please squash the commits down and update this PR.

Thanks for using Jasmine!

slackersoft:
Closing due to inactivity

Issue #1311, opened 2017-04-08
After hooks fire in an inconsistent order.


gdborton:
### Are you creating an issue in the correct repository?

I think so 🤔 .

### When submitting an issue, please answer the following:

 - What version are you using? 2.5.3
 - What environment are you running Jasmine in (node, browser, etc)? node v4.6.2
 - How are you running Jasmine (standalone, npm, karma, etc)? `jasmine ./path/to/spec.js`
 - If possible, include an example spec that demonstrates your issue. See below.

Given the following test file

```javascript
describe('stuff', function() {
  beforeAll(console.log.bind(null, 'beforeAll1'));
  beforeAll(console.log.bind(null, 'beforeAll1'));

  beforeEach(console.log.bind(null, 'beforeEach1'));
  beforeEach(console.log.bind(null, 'beforeEach2'));

  afterEach(console.log.bind(null, 'afterEach1'));
  afterEach(console.log.bind(null, 'afterEach2'));

  afterAll(console.log.bind(null, 'afterAll1'));
  afterAll(console.log.bind(null, 'afterAll2'));

  it('should pass', console.log.bind(null, 'test'));
});
```

I get the following output:

```
❯ jasmine ./test/what_spec.js --silent
Started
beforeAll1
beforeAll1
beforeEach1
beforeEach2
test
afterEach2
afterEach1
.afterAll1
afterAll2
```

Note that `afterEach` gets its order reversed while `afterAll` does not.

Personally I'd like to see neither of them reversed, but that doesn't matter as much as consistency.  I found this while working on a tool that helps automate some setup/teardown, and noticed that the teardown was happening in the wrong order.

slackersoft:
See the discussion in #908 about what is the correct order for `afterEach` in a BDD framework. Long story short the current behavior is the right behavior and we're probably not going to change it. 

Hope this helps. Closing this. Thanks for using Jasmine!

gdborton:
@slackersoft The issue here isn't that `afterEach` is run in reverse order, but that `afterAll` isn't. Shouldn't these behave consistently with each other?

slackersoft:
Ahh! `afterEach` and `afterAll` should both be run in the opposite order they are declared. If this isn't happening for `afterAll`, we would be happy to review a pull request to correct this. 

Sorry for the confusion.

PR #1312, opened 2017-04-10
[lifecycle hooks] Make afterAll hooks operate in the fashion as afterEach.

gdborton:
It was discovered that afterAll hooks run in the same order that you add them,
while afterEach hooks were running in reverse order.  This commit makes their
order consistent, and adds regression tests.

Relevant issue - https://github.com/jasmine/jasmine/issues/1311


mmikirtumov:
Previously our afterAll logic the following way
-> afterAll1
->afterAll2

Now we see many issues in our test cases.

I think it will be better to add some part of breaking changes part in Release Notes
https://github.com/jasmine/jasmine/blob/master/release_notes/2.6.0.md

Because this kind of change can bring many problems for developers.

Thanks

PR #1330, opened 2017-04-25
remove eval to create wrapper

### LINKED ISSUE #1325: 2.6.0 Breaks CSP in Atom
`EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'"`

UziTech/atom-jasmine2-test-runner#14

UziTech:
fixes #1325

### DIFF
--- a/src/core/Spy.js
+++ b/src/core/Spy.js
@@ -14,9 +14,10 @@ getJasmineRequireObj().Spy = function (j$) {
    * @name Spy
    */
   function Spy(name, originalFn) {
-    var wrapper = function () {
+    var numArgs = (typeof originalFn === 'function' ? originalFn.length : 0),
+      wrapper = makeFunc(numArgs, function () {
         return spy.apply(this, Array.prototype.slice.call(arguments));
-      },
+      }),
       spyStrategy = new j$.SpyStrategy({
         name: name,
         fn: originalFn,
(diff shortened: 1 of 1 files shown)


slackersoft:
It looks to me like this will effectively revert #1055 as spies will no longer have the same arity as the function they are replacing. I'd be happy to talk through ways to solve this without `eval`.

UziTech:
I am hoping to figure out a way to do it with the functions to keep their arity

glebec:
[Like I said](https://github.com/jasmine/jasmine/pull/1055#issuecomment-196168851) in the original issue, [Sinon uses hard-coded functions](https://github.com/sinonjs/sinon/blob/2c89183415ad34f1627dc87094dd88bc1e57b2f2/lib/sinon/spy.js#L78) as a workaround to avoid `eval`.

We looked at this pretty carefully and it seemed like the only possible solutions were either hard-coded functions or `eval`. Maybe with modern (e.g. ES2015) use of metaprogramming tools there's some new trick out there, but I don't know offhand.

UziTech:
we could use `defineProperty` if all of the supported platforms support it

```javascript
Object.defineProperty(wrapper, "length", { value: numArgs });
```

Otherwise hard coded functions will have to do.

slackersoft:
IE 8 only supports `defineProperty` onDOM nodes, so we'll have to stick with hardcoded functions for now

PR #1332, opened 2017-04-25
+ fix GlobalErrors - check for undefined global.process.listeners

gustavomick:
jasmine makes karma fails when global.process.listeners is undefined , karma just stops execution showing jasmine error. this simple extra check fix it.

working with windows 10 and karma-typescript.


slackersoft:
This seems reasonable, but you haven't provided any tests for this, and it looks like at least one other test may have failed.

I've fixed the `nextTick` thing in node.js, so you might want to rebase onto master as well.

gustavomick:
not sure what test could add value there, but let me know. i meantime i rebased and sent other pull request.

PR #1333, opened 2017-04-25
+ fix GlobalErrors - check for undefined global.process.listeners

gustavomick:
jasmine makes karma fails when global.process.listeners is undefined , karma just stops execution showing jasmine error. this simple extra check fix it.

working with windows 10 and karma-typescript.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2174,7 +2174,7 @@ getJasmineRequireObj().GlobalErrors = function(j$) {
     this.uninstall = function noop() {};
 
     this.install = function install() {
-      if (global.process && j$.isFunction_(global.process.on)) {
+      if (global.process && global.process.listeners && j$.isFunction_(global.process.on)) {
         var originalHandlers = global.process.listeners('uncaughtException');
         global.process.removeAllListeners('uncaughtException');
         global.process.on('uncaughtException', onerror);


dadleyy:
Just a note on this - I was running into this same problem for a typescript + karma project I've been working on. [`karma-typescript`](https://github.com/monounity/karma-typescript) uses [`node-process`](https://github.com/defunctzombie/node-process) for some reason but allows users to prevent it from being bundled into the browser via a [`bundlerOptions.addNodeGlobals`](https://github.com/dadleyy/hoctable/blob/master/karma.conf.js#L110-L112) flag. This prevents it from trying to make `global.process` a thing.

I mentioned the issue over at the `process` package [here](https://github.com/defunctzombie/node-process/pull/77#issuecomment-297122530) and it has since been added as a no-op function which I don't think people would want given that it would basically mean turning off the `GlobalErrors` feature added in `jasmine@2.6`.  I think a nice feature for the jasmine core library would be to provide developers a conditional around the [installation of globalerrors](https://github.com/jasmine/jasmine/blob/master/src/core/Env.js#L262) that they can use some sort of configuration option for.

slackersoft:
@dadleyy do you mean giving users the option to not have Jasmine try to associate errors in async specs correctly?

2.6.1 should have the `process.listeners` check in it as well to help alleviate this issue for environments that add extra node like things that aren't really there.

dadleyy:
>  do you mean giving users the option to not have Jasmine try to associate errors in async specs correctly?

Moreso just looking to have the ability to prevent jasmine from [taking over all `uncaughtException` errors](https://github.com/jasmine/jasmine/blob/master/src/core/GlobalErrors.js#L14-L32). Seems like something that invasive should be optional.

slackersoft:
Jasmine will do its best to put back whatever handlers were already there before the test run started, so hopefully this should only be an issue if your specs rely on an `uncaughtException` actually going through to some pre-installed handler. When adding this, we made sure to wait until the suite was being executed to install and uninstall immediately upon the suite finishing.

Unfortunately, the only way to properly associate errors that occur in async code it to listen for `uncaughtException` and hand them back to the spec. As more of our users are using more specs with async code, handling errors and continuing on with the suite is more and more important.

PR #1339, opened 2017-04-29
Update SpyStragegy.js

DRayX:
Allow SpyStrategy to handle an original or fake callable that doesn't have Function in its prototype chain.


slackersoft:
Can you describe the use-case for this change a bit more? A spy is expected to be directly invokable like a `function`, which should mean that it generally needs to basically be a function.

slackersoft:
Closing due to inactivity

PR #1340, opened 2017-04-30
Add ES6 map support to Jasmine

rmehlinger:
This commit adds support for ES6 `Map` objects to Jasmine, and switches the equality comparator for ES6 `Sets` to use `entries` rather than `values`--see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/entries.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2482,12 +2482,12 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       if (!result) {
         return false;
       }
-    } else if (className == '[object Set]') {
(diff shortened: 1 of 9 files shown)


roddolf:
Is there any time to release this feature?

theefer:
I've also hit this issue. Would be great to have a release to fix it as it's an easy source of false positives in tests.

rmehlinger:
I'd love to have this committed as well. Is there any process that I need to go through to get this branch ready for commit?

slackersoft:
Jasmine 2.7 was released on Friday July 28 and includes this feature. @rmehlinger, just update.

rmehlinger:
Sweet! Ty :)

PR #1343, opened 2017-05-05
Don't crash if an error occurs when no error handlers are installed

sgravrock:
It's possible for an exception to be thrown when there aren't any listeners installed with Jasmine's GlobalErrors object. In that case,  it will try to call the nonexistent listener. The resulting exception masks the original problem (in our case, a possible state corruption bug that we're still trying to pin down).


sgravrock:
This doesn't work reliably on node. I'm not sure exactly what's up but I'm seeing situations where the handler is called even after I've verified (via logging) that it's been uninstalled. And something about repeatedly installing and uninstalling the onerror handler seems to greatly increase the chance that Jasmine's own test suite will generate an error when there aren't any listeners hooked up.

PR #1344, opened 2017-05-05
Handle exceptions thrown "during" a suite with no children.

sgravrock:
This is tricky to reproduce, but it can happen in certain async scenarios in the presence of focused specs if an async spec tells Jasmine that it's complete and then causes the global error event to fire later on. In that situation we avoid crashing and record the failure as a top level afterAll failure, so that it at least get surfaced to the user somewhere.

### DIFF
--- a/spec/core/SpecSpec.js
+++ b/spec/core/SpecSpec.js
@@ -357,6 +357,23 @@ describe("Spec", function() {
     }]);
   });
 
+  it("should return true when handling an exception", function() {
+    var spec = new jasmineUnderTest.Spec(
+      {queueableFn: { fn: function() {} }
+    });
+
+    expect(spec.onException('foo')).toEqual(true);
+  });
+
+  it("should return true when handling a pending spec exception", function() {
+    var spec = new jasmineUnderTest.Spec(
+      {queueableFn: { fn: function() {} }
+      });
+    var error = new Error(jasmineUnderTest.Spec.pendingSpecExceptionMessage);
+
+    expect(spec.onException(error)).toEqual(true);
+  });
+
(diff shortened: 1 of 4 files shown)


spec/core/SpecSpec.js
```diff
@@ -357,6 +357,23 @@ describe("Spec", function() {
     }]);
   });
 
+  it("should return true when handling an exception", function() {
```
sgravrock: I'd have preferred an integration test that tells the overall story, like the various Env integration tests, but I haven't been able to reproduce the overall behavior reliably enough to do that.

src/core/Spec.js
```diff
@@ -78,11 +78,11 @@ getJasmineRequireObj().Spec = function(j$) {
   Spec.prototype.onException = function onException(e) {
     if (Spec.isPendingSpecException(e)) {
       this.pend(extractCustomPendingMessage(e));
-      return;
+      return true;
     }
 
     if (e instanceof j$.errors.ExpectationFailed) {
-      return;
+      return false;
```
slackersoft: I think that a spec should probably always return `true` to denote it handled any exception. The only case I can think of where we might want it to propagate up the suite chain would be a disabled spec.

spec/core/SpecSpec.js
slackersoft: I think the unit test is a bit nicer in this case since it is a bit easier to note which cases cause `true` or `false`.

sgravrock:
#1352 prevents this scenario from happing in the first place. So I think it makes sense to close this PR if #1352 or something like it gets merged.

PR #1356, opened 2017-05-12
Support for promises and and ES2017 async/await in before/it/after

sgravrock:
This is similar to #1270 but updated to account for recent changes to QueueRunner. I also added support for async functions, since that was nearly free once promises were supported.

### DIFF
--- a/spec/helpers/asyncAwait.js
+++ b/spec/helpers/asyncAwait.js
@@ -1,0 +1,27 @@
+(function(env) {
+  function getAsyncCtor() {
+    try {
+      eval("var func = async function(){};");
+    } catch (e) {
+      return null;
+    }
+
+    return Object.getPrototypeOf(func).constructor;
+  }
+
+  function hasAsyncAwaitSupport() {
+    return getAsyncCtor() !== null;
+  }
+
+  env.makeAsyncAwaitFunction = function() {
+    var AsyncFunction = getAsyncCtor();
+    return new AsyncFunction("");
+  };
+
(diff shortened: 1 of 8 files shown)


spec/helpers/asyncAwait.js
```diff
@@ -0,0 +1,27 @@
+(function(env) {
+  function getAsyncCtor() {
+    try {
+      eval("var func = async function(){};");
```
sgravrock: This is nasty but I haven't been able to find a way to do this that neither requires eval nor breaks in environments that don't support async.

src/core/Env.js
```diff
@@ -319,6 +319,12 @@ getJasmineRequireObj().Env = function(j$) {
       }
     };
 
+    var ensureIsFunctionOrAsync = function(fn, caller) {
```
sgravrock: I'm not completely wild about the naming here or in base.js, but I think that most Jasmine code that asks "is this thing a function?" isn't prepared to handle async functions. So it's probably best for j$.isFunction and ensureIsFunction to continue to reject async functions.

NicBright:
This is great. What about documentation? Can't find it :(

slackersoft:
The API docs have a bit about it here: https://jasmine.github.io/api/edge/global.html#implementationCallback. We're still looking into how best to format all of the documentation for Jasmine. The big long pages like introduction.html are pretty and can read nicely, but they need to be broken apart a bit to be more useful.

PR #1362, opened 2017-05-25
Fail when one of the arguments passed into toBeCloseTo matcher is null

beatrichartz:
Currently, `toBeCloseTo` passes when one of the arguments passed into it is `null`. This is due to the quirk that `9 - null` evaluates to `9`, and also `null - 9` evaluates to `-9`. Whether that's well defined in the language is probably a separate discussion. The matcher should fail though since it can't be said whether `null` is close to any other value.

### DIFF
--- a/spec/core/matchers/toBeCloseToSpec.js
+++ b/spec/core/matchers/toBeCloseToSpec.js
@@ -29,6 +29,22 @@ describe("toBeCloseTo", function() {
     expect(result.pass).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


slackersoft:
I think it would be a bit nicer of an error to just have `toBeCloseTo` throw an `Error` with a message about which argument was null, instead of simply failing. Since the `null` is probably accidental, this should help people find the real issue quicker.

Thanks!

beatrichartz:
@slackersoft ok, how does the updated code look to you?

PR #1375, opened 2017-06-22
Make toEqual matcher report the difference between arrays with different length

kiewic:
The purpose of this change, is to improve the readability of specs that fail because two arrays have different length. Instead of printing the regular error message:

```bash
Expected $.values = [ 'foo', 'bar', 'baz', 'fizz', 'buzz', 'bazz', 'fur', 'bliz', 'pugg', 'bizz', 'fixx' ] to equal [ 'foo', 'bar', 'baz', 'fizz', 'buzz', 'bazz', 'fur', 'pugg', 'bizz' ].
```

The error message will include only the differences between the two arrays:

```bash
Expected $.values[7] = 'bliz' to equal 'pugg'.
Expected $.values[8] = 'pugg' to equal 'bizz'.
Expected $.values[9] = 'bizz' to equal undefined.
Expected $.values[10] = 'fixx' to equal undefined.
```

Please take a look.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -2529,15 +2529,12 @@ getJasmineRequireObj().matchersUtil = function(j$) {
     // Recursively compare objects and arrays.
     // Compare array lengths to determine if a deep comparison is necessary.
     if (className == '[object Array]') {
-      size = a.length;
-      if (size !== b.length) {
-        diffBuilder.record(a, b);
-        return false;
-      }
+      var aLength = a.length;
+      var bLength = b.length;
 
-      for (i = 0; i < size; i++) {
+      for (i = 0; i < aLength || i < bLength; i++) {
         diffBuilder.withPath(i, function() {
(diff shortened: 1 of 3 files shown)


sgravrock:
This looks like a potentially useful change, but I'm seeing a couple of issues with the PR as it stands now:

* It breaks some of the existing tests.
* It treats different-length arrays as equal if the extra elements are all `undefined`. For instance, `expect([1, 2, undefined]).toEqual([1, 2])` should fail but this PR makes it pass.

I'd be happy to review a version of this that passes all tests and changes only the message produced by the matcher, not its pass/fail behavior. Thanks for using Jasmine!

kiewic:
Hi @sgravrock, thank you for looking at my PR. I put back the array length comparison, and added a test to cover the case you described. The errors messages now include a line like `Expected $.length = 1 to equal 2.` to clarify the error, although this may be redundant. I'm happy to hear your suggestions.

Also reverted changes to *jasmine.js* as described in the instructions.

Please take a look again.

Issue #1377, opened 2017-06-23
Uncaught ReferenceError: jasmineRequire is not defined


wood1986:
I found the example on https://github.com/jasmine/jasmine#installation can be done by the webpack with html-webpack-plugin. Therefore I try to wire them up instead of hardcoding the script

```
// webpack.config.js

const path = require("path"),
      webpack = require("webpack"),
      HtmlWebpackPlugin = require('html-webpack-plugin'),
      ExtractTextPlugin = require("extract-text-webpack-plugin"),
      jasmineCore = require("jasmine-core");

module.exports = {
  "devServer": {
    "compress": true,
    "host": "0.0.0.0",
    "disableHostCheck": true,
    "clientLogLevel": "error"
  },
  "entry": {
    "client": path.resolve(__dirname, "src/pages/Spec/Spec.web.tsx")
  },
  "module": {
    "rules": [
      {
        "test": /\.spec\.tsx?$/,
        "enforce": "pre",
        "loader": "tslint-loader",
        "options": {
          "configuration": require("./tslint.json"),
          "tsConfigFile": "tsconfig.json",
          "fix": true
        }
      },
      {
        "test": /\.tsx?$/,
        "exclude": /node_modules/,
        "loader": "awesome-typescript-loader"
      },
      {
        "test": /jasmine\.css$/,
        "use": ExtractTextPlugin.extract({
          "fallback": "style-loader",
          "use": ["css-loader"]
        })
      },
      {
        "test": /\.scss$/,
        "use": "null-loader"
      }
    ]
  },
  "resolve": {
    "alias": {
      "jasmine": path.resolve(jasmineCore.files.path, "jasmine.js"),
      "jasmine-core": path.resolve(jasmineCore.files.path)
    },
    "extensions": [".ts", ".tsx", ".js", ".json", ".jsx"]
  },
  "output": {
    "filename": "[name].spec.js",
    "libraryTarget": "umd",
    "path": path.resolve(__dirname, `spec`),
  },
  "plugins": [
    new HtmlWebpackPlugin({
      "minify": {
        "removeComments": true,
        "collapseWhitespace": true
      },
      "filename": "index.html",
      "template": path.resolve(__dirname, "src/pages/Spec/Spec.ejs"),
      "favicon": path.resolve(jasmineCore.files.imagesDir, "jasmine_favicon.png")
    }),
    new ExtractTextPlugin("[name]-[contenthash].css"),
  ],
  "target": "web"
};

```

[...]

PR #1381, opened 2017-06-24
fix the environment issue and add a jasmine-core.standalone.js

wood1986:
#1377

### DIFF
--- a/lib/jasmine-core.standalone.js
+++ b/lib/jasmine-core.standalone.js
@@ -1,0 +1,3 @@
+require("./jasmine-core/jasmine.js");
+require("./jasmine-core/jasmine-html.js");
+require("./jasmine-core/boot.js");
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -23,7 +23,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 var getJasmineRequireObj = (function (jasmineGlobal) {
   var jasmineRequire;
 
-  if (typeof module !== 'undefined' && module.exports && typeof exports !== 'undefined') {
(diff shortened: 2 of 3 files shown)


wood1986:
Here is the bug example https://github.com/wood1986/jasmine-bug-example

sgravrock:
I'm seeing a few issues with this pull request as it stands now, particularly the lack of tests, but before we go into that I think it's worth talking about what it's for. From the issue that you linked to I understand that you're trying to include Jasmine via Webpack. Can you give us some information on why you want to do that rather than just including the Jasmine JS files via script tags in your spec runner HTML? There's an ongoing labor cost to supporting additional environments like Webpack, and I want to make sure we're spending future maintainer time wisely.

wood1986:
1. The problem is the standalone setting can be done by the `webpack`. Then you don't need to hardcode the version in your tailor-made html to run the standalone page.

2. As I try to wire them up the standalone setting and `webpack`, it gives the #1377 but it can be fixed by using `script-loader`. I think this is a hack not the correct fix.

3. I believe the the root cause of #1377 is that the existing environment checking is not enough

4. It just cost you 30 seconds or less to update `jasmine-core.standalone.js`. Actually you do not need to supporting the additional environment.

wood1986:
any update?

slackersoft:
I still don't think this is a Jasmine issue, so much as a CommonJS/WebPack concern, considering that [gulp-jasmine-browser](https://github.com/jasmine/gulp-jasmine-browser) work [with webpack](https://github.com/jasmine/gulp-jasmine-browser/#usage-with-webpack) without any modifications to Jasmine itself.

Additionally, you've named your new boot file `jasmine.standalone.js` but it has nothing to do with the standalone-ness of the Jasmine environment, but rather the usage of CommonJS and WebPack.

I don't think this is something we're looking to change in Jasmine at this time. Thanks for using Jasmine!

PR #1385, opened 2017-06-27
Add matcher aliases "toBeTruey" & "toBeFalsey" [sic]

matthewadams:
I've noticed many people trip over the `toBeFalsy` matcher's spelling, plus `toBeFalsy` & `toBeTruthy` are inconsistent with respect to `false` & `true`, respectively.

This PR normalizes the mapping from boolean value to adjective (`true` -> `toBeTruey` & `false` -> `toBeFalsey`) by adding matcher aliases.

Also see https://twitter.com/matthewadams12/status/872469510745468928.  :)

### DIFF
--- a/spec/core/matchers/toBeFalsySpec.js
+++ b/spec/core/matchers/toBeFalsySpec.js
@@ -35,4 +35,8 @@ describe("toBeFalsy", function() {
     result = matcher.compare({});
     expect(result.pass).toBe(false);
   });
+
+  it("has 'falsy' aliases", function() {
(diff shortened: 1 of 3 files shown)


matthewadams:
Doc PR:  https://github.com/jasmine/jasmine.github.io/pull/73

slackersoft:
[Truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) and [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) are the generally accepted spelling in Javascript for things that make an `if` succeed or fail respectively. To this end, I don't think this is something we're looking to change at this point in time.

I'm going to close this. Thanks for using Jasmine!

PR #1386, opened 2017-06-27
Add Promise checking to eq

### LINKED ISSUE #1314: Mocks' toHaveBeenCalledWith method behaves incorrectly with promise arguments
Calling a mock with a promise will make any later assertions on that mock's calls with other promises it was not called with fail.

In other words, calling a mock with a promise will make that mock behave as if it were called with any other promises, even when it was not called with another promise.

 - jasmine and jasmine-core version 2.5.2
 - Running with Karma through Chrome 56.0.2924.87

Failing reproduction spec:
```
    it('verifies a method was not called with promise2 as a parameter', () => {
        const promise1 = new Promise(resolve => resolve({
            data: 'foo'
        }))
        const promise2 = new Promise(resolve => resolve({
            data: 'bar'
        }))

        const someMock = jasmine.createSpy('something')
        someMock(promise1)

        expect(someMock).not.toHaveBeenCalledWith(promise2)
    })
```

In Chrome, running promise1 === promise2 and promise1 == promise2 returns false for both cases.

Please let me know if additional info would help :)

sderickson:
Fixes #1314 

Tested on:
* Mac Chrome
* Mac Firefox
* Mac Safari
* Virtualbox IE11
* Virtualbox Edge
* Node 7

Both before and after running `grunt buildDistribution`.

All files are lint free.

### DIFF
--- a/src/core/base.js
+++ b/src/core/base.js
@@ -75,6 +75,10 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return obj.nodeType > 0;
   };
 
+  j$.isPromise = function(obj) {
+    return (typeof Promise !== 'undefined') && obj.constructor === Promise;
+  };
+
   j$.fnNameFor = function(func) {
     if (func.name) {
       return func.name;
--- a/spec/core/matchers/matchersUtilSpec.js
+++ b/spec/core/matchers/matchersUtilSpec.js
@@ -166,6 +166,16 @@ describe("matchersUtil", function() {
 
       expect(jasmineUnderTest.matchersUtil.equals(a,b)).toBe(true);
     });
+    
+    it("passes for equivalent Promises (GitHub issue #1314)", function() {
+      if (typeof Promise === 'undefined') { return; }
+
+      var p1 = new Promise(function () {}),
+        p2 = new Promise(function () {});
+
+      expect(jasmineUnderTest.matchersUtil.equals(p1, p1)).toBe(true);
+      expect(jasmineUnderTest.matchersUtil.equals(p1, p2)).toBe(false);
+    });
 
     describe("when running in a browser", function() {
       function isNotRunningInBrowser() {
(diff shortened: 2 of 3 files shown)


src/core/base.js
```diff
@@ -75,6 +75,10 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return obj.nodeType > 0;
   };
 
+  j$.isPromise = function(obj) {
+    return (typeof Promise !== 'undefined') && obj.constructor === Promise;
```
matthewadams: Would a more portable way to do this be `obj.constructor.name === 'Promise'`?  Line 79 as written would only support native `Promise`s.  I'm not sure it's necessary yet, but have you considered allowing users to tell Jasmine which Promise library you're using via something like `jasmine.setPromiseLib(require('bluebird'))` or similar?

sderickson: I tend to avoid basing logic on `obj.constructor.name` since it can break in production when they're modified by build systems. In a production build using uglify I tested, bluebird's constructor name ended up `"o"`. Figured the global Promise would have more consistent behavior. What do you think?

I did consider being able to set a Promise library, but went for the simplest implementation to start off. I'd be happy to add that configuration option, though.

slackersoft: I would prefer to not even see `Promise` in here if we can help it. Possibly checking for a `then` function on the object when it doesn't have any keys or something. This would implicitly allow other promise libraries, without the user having to configure anything in Jasmine to make it work.

Does that seem like it would make sense?

sgravrock: After talking this over with @slackersoft I'm going to go ahead and merge this PR. The drawback to checking for `then` is that it might incorrectly treat a non-promise object that has a `then` method as a promise. It would be nice to be able to extend this behavior to promise libraries like Bluebird, and I'd be happy to review a pull request that does that. But even if we only handle native promises, that's an improvement.

PR #1394, opened 2017-07-06
Update jasmine-html.js

DanAtShenTech:
Fix "Uncaught TypeError: Cannot read property 'env' of undefined at new HtmlReporter (jasmine-html.js:38)"  that results when call to new HtmlReporter() is made without defining options.env.

### DIFF
--- a/lib/jasmine-core/jasmine-html.js
+++ b/lib/jasmine-core/jasmine-html.js
@@ -35,7 +35,7 @@ jasmineRequire.HtmlReporter = function(j$) {
   };
 
   function HtmlReporter(options) {
-    var env = options.env || {},
+    var env = "env" in options : options.env ? {},
       getContainer = options.getContainer,
       createElement = options.createElement,
       createTextNode = options.createTextNode,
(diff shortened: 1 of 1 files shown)


DanAtShenTech:
I don't know why this commit is failing.  I only proposed one tiny, very straightforward change.

jaapz:
The [build fails](https://travis-ci.org/jasmine/jasmine/jobs/250916963) because you used double quotes while the linter wants you to use single quotes.
Also, isn't the `in` operator an ES6 feature? I think jasmine isn't currently using any ES6 features.

sgravrock:
It would help to know a little more about why you want to make this change. `HtmlReporter` already allows omitting `options.env` to support easier unit testing of the methods that don't use it, but it needs an `env` for any real use. The error message that you quoted sounds more like the result of omitting the `options` parameter altogether, and I'm not sure that's something we want to support.

I'd also suggest taking a look at the [contributor's guide](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md). It includes some pull request guidelines and also explains how to run the same tests and linters that are run by the Travis build. 

Thanks for using Jasmine!

slackersoft:
Closing due to inactivity

Issue #1317, opened 2017-04-17
Calling toThrowError with no arguments errors via assuming the undefined argument is a regex


tsclaus:
The current behavior of `toThrowError` assumes that if the first argument isn't a string or a function, then it must be a regex; it subsequently calls `.test` on the undefined first argument and throws the error `TypeError: expected.test is not a function` (error partially suppressed by Protractor wrapping Jasmine).

The behavior should change to `.toThrowError()` being identical to `.toThrowError(Error)`.

I am running Jasmine 2.4.1 (The 2.5 docs do not include an argumentless example so I think this carries through to 2.5) in node/protractor.

slackersoft:
That seems reasonable. I would be happy to review a pull request to treat `toThrowError` without arguments the same as if it had been called with just the `Error` constructor.

slackersoft:
Upon further inspection, it looks like this already works in Jasmine. The following spec passes:

```javascript
it('checks errors thrown', function() {
  expect(function() {
    throw new TypeError();
  }).toThrowError();

  expect(function() {
    throw new TypeError();
  }).toThrowError(Error);
});
```

So it seems like this is already resolved. Thanks for using Jasmine!

PR #1397, opened 2017-07-15
Fix for Jasmine's Issue #1317

Zagagy:
Proposed fix for issue #1317 - Calling toThrowError with no arguments errors via assuming the undefined argument is a regex

This fix adds another case for the getMatcher function, tol handle arguments list of size 1.


slackersoft:
Looks like this has become a duplicate of #1398 Closing.

PR #1398, opened 2017-07-15
Patch 2

Zagagy:
Proposed fix for issue #1317 - Calling toThrowError with no arguments errors via assuming the undefined argument is a regex

This fix adds another case for the getMatcher function, tol handle arguments list of size 1.

### DIFF
--- a/spec/core/matchers/toThrowErrorSpec.js
+++ b/spec/core/matchers/toThrowErrorSpec.js
@@ -121,10 +121,10 @@ describe("toThrowError", function() {
       },
       result;
 
-    result = matcher.compare(fn);
+    result = matcher.compare(fn, TypeError, "");
 
     expect(result.pass).toBe(true);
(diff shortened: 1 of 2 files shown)


slackersoft:
Since it doesn't look like the root cause for the original issue is still a problem, and I'm not totally sure how this PR actually fixes the issue, I'm going to close it.

Sorry for the confusion. Thanks for using Jasmine!

Zagagy:
Hi,
The original problem was that there wasnt a possibility of calling the
"toThrowError" without any arguments due to the check of 2 arguments or more
that is what this small fix came to solve, just overloading the
toThrowError Function, how come this minor fix doesnt solve the issue?

2017-07-21 7:01 GMT+03:00 Gregg Van Hove <notifications@github.com>:

slackersoft:
As [discussed](https://github.com/jasmine/jasmine/issues/1317#issuecomment-316598186) in #1317 it looks like the underlying issue is already solved in master and `toThrowError` can be called without any arguments.

PR #1400, opened 2017-07-25
User context

darthjee:
In order to allow better extendability within the user cotext, this PR wraps it in a class with no methods or attributes that will behave just like ```Object``` but still allow for user to write helpers that extend it without affecting the behavior of ```Object``` objects

## Description
- [x] New UserContext class
- [x] Split extract extend method from within clone
- [x] Use clonedSharedUserContext when a clone of user context is needed

## Motivation and Context
While I was trying to write a method that acts just like let for ruby, I discovered that I couldn't easily extend the user context without ruining the behavior of Object class

## How Has This Been Tested?
It has mainly been tested through specs and by extending it in a spec on a separated project

## Types of changes

## Checklist:
- [x] Tested on linux with Grunt
- [x] Tested on Linux Firefox (rake jasmine)
- [x] Tested on Linux Chrome (rake jasmine)
- [ ] Tested on windows internet explorer (rake jasmine)


src/core/Suite.js
```diff
@@ -147,15 +147,18 @@ getJasmineRequireObj().Suite = function(j$) {
...
-        clonedObj[prop] = obj[prop];
+        newObj[prop] = obj[prop];
       }
     }
 
-    return clonedObj;
+    return newObj;
+  }
+
+  function clone(obj) {
```
slackersoft: If the `clone` function is no longer used within Suite, we should just remove it.

src/core/Suite.js
```diff
@@ -88,14 +88,14 @@ getJasmineRequireObj().Suite = function(j$) {
...
-      this.sharedContext = this.parentSuite ? clone(this.parentSuite.sharedUserContext()) : {};
+      this.sharedContext = this.parentSuite ? this.parentSuite.clonedSharedUserContext() : new j$.UserContext();
     }
 
     return this.sharedContext;
   };
 
   Suite.prototype.clonedSharedUserContext = function() {
-    return clone(this.sharedUserContext());
+    return extend(new j$.UserContext(), this.sharedUserContext());
```
slackersoft: Now that there is a place to put it, could this be a method on the `UserContext` constructor, so Suite would call something like:

```javascript
j$.UserContext.fromExisting(this.sharedUserContext())
```

And remove all of the `extend` stuff from Suite.

darthjee: makes sense ... I didnt want to put in the prototype, but like this would work, will do that later

:+1:

src/core/Suite.js
darthjee: :+1:

darthjee:
@slackersoft after your review, I added more commits for estra specs and changed QueueRunner to never initialize userContext as an empty object.

I am not being able to run the tests on windows, but will have access to a real windows machine on the weekend so I can try running it there

Issue #1180, opened 2016-08-13
Improve default error messages for typed arrays


reiner-dolp:
For typed arrays, only a comma-separated list of array elements is output, which can cause confusion.

**For example:**

``` js
expect(new Uint32Array([1,3,2,5,4,8,])).toEqual(new Uint8Array([1,3,2,5,4,8,]));
```

outputs

```
Error: Expected 1,3,2,5,4,8 to equal 1,3,2,5,4,8.
```

Gerg:
How would you suggest we pretty print typed arrays? Ideally we could find a solution that doesn't affect regular arrays (since they are far more common).

reiner-dolp:
How about one of these: 
```
if(is_typed_array()) {
msg = "Error: Expected array 1,3,2,5,4,8 of type uint32 to equal array 1,3,2,5,4,8 of type uint8"
msg = "Error: Expected array type uint32 to equal uint8" // <-- hides values until type matches
msg = "Error: Expected Uint32Array 1,3,2,5,4,8 to equal Uint8Array 1,3,2,5,4,8" // <-- my fav
}
```

Gerg:
I also like "Error: Expected Uint32Array 1,3,2,5,4,8 to equal Uint8Array 1,3,2,5,4,8" as long as it doesn't show the type for regular arrays. For example, we **don't** want "Error: Expected Array 1,3,2,5,4,8 to equal Array 11,13,12".

We'd be happy to review a pull request to accomplish this.

mezoistvan:
Hi! I'm using Jasmine as my daily driver, and I'd like to start working on this issue if that's ok!

reiner-dolp:
I am not working on it. Would appreciate!

PR #1404, opened 2017-08-01
Improved error message for typed arrays

ksvitkovsky:
Typed arrays presentation now contains info about their type.

## Description
- added helper method to check if value is typed array - since we know which types there might be, method simply checks each;
- added clause in PrettyPrinter' format method that catches typed arrays;

## Motivation and Context
Without this changes (requested in #1180) in cases when toEqual compares two arrays of different types but equal elements error message is unclear, because no type data is given.

## How Has This Been Tested?
- added helper function to check that typed arrays are supported in current environment;
- added test case that compares two arrays of different types;

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/matchers/toEqualSpec.js
+++ b/spec/core/matchers/toEqualSpec.js
@@ -216,7 +216,7 @@ describe("toEqual", function() {
 
     var actual = new Uint32Array([1, 2, 3]),
       expected = new Uint16Array([1, 2, 3]),
-      message = "Expected Uint32Array 1,2,3 to equal Uint16Array 1,2,3.";
+      message = "Expected Uint32Array [ 1, 2, 3 ] to equal Uint16Array [ 1, 2, 3 ].";
 
     expect(compareEquals(actual, expected).message).toEqual(message);
   });
--- a/spec/core/matchers/toEqualSpec.js
+++ b/spec/core/matchers/toEqualSpec.js
@@ -211,6 +211,16 @@ describe("toEqual", function() {
     expect(compareEquals(actual, expected).message).toEqual(message);
   });
 
+  it("reports mismatches between arrays of different types", function() {
+    jasmine.getEnv().requireFunctioningTypedArrays();
+
+    var actual = new Uint32Array([1, 2, 3]),
+      expected = new Uint16Array([1, 2, 3]),
+      message = "Expected Uint32Array 1,2,3 to equal Uint16Array 1,2,3.";
+
+    expect(compareEquals(actual, expected).message).toEqual(message);
+  });
+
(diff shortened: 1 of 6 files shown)


src/core/PrettyPrinter.js
```diff
@@ -210,6 +212,13 @@ getJasmineRequireObj().pp = function(j$) {
     this.append(' })');
   };
 
+  StringPrettyPrinter.prototype.emitTypedArray = function(obj) {
+    var constructorName = j$.fnNameFor(obj.constructor);
+    var stringifiedArray = Array.prototype.join.call(obj);
```
slackersoft: Typed arrays should also respect the `MAX_PRETTY_PRINT_ARRAY_LENGTH` in the same way that normal arrays do.

spec/core/matchers/toEqualSpec.js
```diff
@@ -211,6 +211,16 @@ describe("toEqual", function() {
     expect(compareEquals(actual, expected).message).toEqual(message);
   });
 
+  it("reports mismatches between arrays of different types", function() {
+    jasmine.getEnv().requireFunctioningTypedArrays();
+
+    var actual = new Uint32Array([1, 2, 3]),
+      expected = new Uint16Array([1, 2, 3]),
+      message = "Expected Uint32Array 1,2,3 to equal Uint16Array 1,2,3.";
```
slackersoft: This should probably change with the pretty printer to look more like a how Jasmine currently prints arrays

ksvitkovsky:
@slackersoft got it! now message should look like `Expected Uint32Array [ 1, 2, 3, ... ] to equal Uint16Array [ 1, 2, 3, ... ].` right?

By the way, I got a question. 
Apparently, in PhantomJS some standard class constructors (typed arrays, ArrayBuffer, and probably others) have `toString` presentation that looks like `[object ClassNameConstructor]` rather than `function ClassName() { [native code] }` that leads jasmine.fnNameFor to mistakenly return `<anonymous>` as a function name. That in its turn leads travis build to fail on PhantomJS.

I suggest to skip the test for PhantomJS completely and raise an issue to improve fnNameFor. Does that sounds right?

slackersoft:
That message sounds good for the pretty printer. I'm not totally sure what we want to do with `fnNameFor` in phantomjs. I think it probably makes sense to try to fix it as part of this PR, since it's the functionality of pretty printing typed arrays that requires the change.

I would be happy to have `fnNameFor` detect the `[object ClassNameConstructor]` and either:
- return just `ClassName`, lopping off the `Constructor` portion
- return `ClassNameConstructor` and have the pretty printer tests check for phantom and change the assertion

Thanks!

Issue #1402, opened 2017-08-01
Sets & Maps should be compared irrespective of order


theefer:
## Expected Behavior

ES6 Sets and Maps are most often used as unordered sets and maps, and I would expect the following tests to pass:

``` javascript
var setA = new Set([3, 6]);
var setB = new Set([6, 3]);
expect(setA).toEqual(setB); // pass

var mapA = new Map([['a', 3], [6, 1]]);
var mapB = new Map([[6, 1], ['a', 3]]);
expect(mapA).toEqual(mapB); // pass
```

## Current Behavior

Jasmine matchers would fail the tests above, since the insertion order is considered for equality.

I would consider this behaviour a bug, since it's surprising to require the same order. As an analogy, Objects also retain insertion order, but Jasmine matches ignore the insertion order when comparing Objects.

## Possible Solution

Can we change the tests and implementation to ignore insertion order for Sets and Maps, and thus match the expected behaviour?

If not, why not? Is there an alternative approach to providing order-agnostic matchers?

## Suite that reproduces the behavior (for bugs)

The behaviour described above is enforced by [the tests](https://github.com/jasmine/jasmine/blob/master/spec/core/matchers/matchersUtilSpec.js#L410).

slackersoft:
I believe the insertion order was left in the equality comparison simply because it was easier to implement at the time. I would be happy to review a pull request to change the equality comparison to ignore insertion order for Maps and Sets.

Thanks for using Jasmine!

theefer:
Great thanks, I've started working on a PR to make the comparison order-agnostic!

PR #1406, opened 2017-08-02
Rewrite ES6 Set and Map comparison to ignore insertion order

theefer:
## Description

Changes the ES6 Set and Map equality comparison algorithm to consider the data structures as unordered, rather than requiring the insertion order to be respected.

For instances, the follow tests now pass:

``` javascript
var setA = new Set([3, 6]);
var setB = new Set([6, 3]);
expect(setA).toEqual(setB); // pass

var mapA = new Map([['a', 3], [6, 1]]);
var mapB = new Map([[6, 1], ['a', 3]]);
expect(mapA).toEqual(mapB); // pass
```

## Motivation and Context

The new behaviour reflects the usual assumption that Sets and Maps are not ordered. The same assumption already applies to Objects, whose properties are technically ordered but that order is not taking into consideration when comparing objects in Jasmine.

Fixes https://github.com/jasmine/jasmine/issues/1402 (Sets & Maps should be compared irrespective of order).

## How Has This Been Tested?

Some more tests were added or fixed to reflect the new behaviour.

## Types of changes

Although it is meant as a "bug fix", it is a breaking change for specific tests that would previously assert that two Sets/Maps are *not* equal due to in difference in order, for example:

``` javascript
expect(new Set([1, 2])).not.toEqual(new Set([2, 1])); // passed before but not after
```

Any existing positive Map/Set equality comparison test that previously passed is still expected to pass:

``` javascript
expect(new Set([1, 2])).toEqual(new Set([1, 2])); // passed before and after
```

## Checklist:

### DIFF
--- a/src/core/matchers/matchersUtil.js
+++ b/src/core/matchers/matchersUtil.js
@@ -239,21 +239,66 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       if (!result) {
         return false;
       }
-    } else if (className == '[object Set]' || className == '[object Map]') {
+    } else if (className == '[object Map]') {
       if (a.size != b.size) {
         diffBuilder.record(a, b);
         return false;
       }
-      var iterA = a.entries(), iterB = b.entries();
-      var valA, valB;
-      do {
-        valA = iterA.next();
-        valB = iterB.next();
-        if (!eq(valA.value, valB.value, aStack, bStack, customTesters, j$.NullDiffBuilder())) {
-          diffBuilder.record(a, b);
-          return false;
+
+      // For both sets of keys, check they map to equal values in both maps
+      var mapKeys = [a.keys(), b.keys()];
+      var mapIter, mapKeyIt, mapKey;
+      for (i = 0; result && i < mapKeys.length; i++) {
+        mapIter = mapKeys[i];
+        mapKeyIt = mapIter.next();
+        while (result && !mapKeyIt.done) {
+          mapKey = mapKeyIt.value;
+          result = eq(a.get(mapKey), b.get(mapKey), aStack, bStack, customTesters, j$.NullDiffBuilder());
+          mapKeyIt = mapIter.next();
         }
-      } while (!valA.done && !valB.done);
+      }
+
+      if (!result) {
+        diffBuilder.record(a, b);
+        return false;
+      }
+    } else if (className == '[object Set]') {
+      if (a.size != b.size) {
+        diffBuilder.record(a, b);
+        return false;
+      }
+
+      // For both sets, check they are all contained in the other set
+      var setPairs = [[a, b], [b, a]];
+      var baseIter, baseValueIt, baseValue;
+      var otherSet, otherIter, otherValueIt, otherValue, found;
+      for (i = 0; result && i < setPairs.length; i++) {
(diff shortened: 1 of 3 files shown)


src/core/matchers/matchersUtil.js
```diff
@@ -239,21 +239,66 @@ getJasmineRequireObj().matchersUtil = function(j$) {
...
+
+      // For both sets of keys, check they map to equal values in both maps
+      var mapKeys = [a.keys(), b.keys()];
+      var mapIter, mapKeyIt, mapKey;
+      for (i = 0; result && i < mapKeys.length; i++) {
+        mapIter = mapKeys[i];
+        mapKeyIt = mapIter.next();
+        while (result && !mapKeyIt.done) {
+          mapKey = mapKeyIt.value;
+          result = eq(a.get(mapKey), b.get(mapKey), aStack, bStack, customTesters, j$.NullDiffBuilder());
```
slackersoft: I think we have a nice opportunity here to get better diff descriptions for maps with mismatched values. We should be able to use `diffBuilder.withPath()` similar to what is done for objects (https://github.com/jasmine/jasmine/pull/1406/files#diff-4b1cbd917e62d27b2645d46dd0d01be3R336) instead of just using a `NullDiffBuilder`.

slackersoft: This will certainly change the message that is output and thus the tests as well.

theefer: I did actually give `diffBuilder.withPath()` while writing this and the output looked odd to me (the `$` was throwing me off), but reading more of the tests now I can see that's the expected error syntax.

However IIUC the ObjectPath used by DiffBuilder only supports strings or numbers as path components, but Map keys can actually be anything, so even assuming we extend ObjectPath, I'm not sure how readable the path would be if a key is a complex object?

I guess there are different options:
1. Just report that the Map/Set are different, as per the current code (noop).
2. Use `diffBuilder.withPath()` iff keys are numbers or strings, and fallback to 1. otherwise.
3. Always `diffBuilder.withPath()` and somehow extend it with some clever syntax to represent paths with objects as keys.
4. Extend the `diffBuilder` to be a lot more expressive in the type of errors it supports (e.g. it'd be cool if we could get errors like "Expected Set([1,2]) to contain 3.", or "Expected Map key Object({ x: [3] }) to be 42.", or some nice visual diffing with highlighting).

You probably have more ideas or things on your roadmap, so let me know what you think is best for this PR.

theefer:
Thanks for the review @slackersoft, could you please advise on the preferred option to report diff paths when keys are not strings or numbers (see my reply above)? Thanks!

slackersoft:
We'll just leave the diffBuilder stuff for now. At some point I think we may want to figure out a better error message, but that can wait.

theefer:
Awesome thanks! If you don't mind me asking, do you have a typical release schedule that this would fall into, or do you have an idea of when it might get released?

slackersoft:
We don't really have a specified release cadence. I tend to aim for every couple of months or so where possible. No one wants to be updating their test framework all the time. The actual timing can vary widely based on how much time I actually have to work on Jasmine.

Since 2.7 just came out, this stuff will be waiting for 2.8.

theefer:
OK, thanks for the reply!

Issue #1295, opened 2017-03-22
Feature Request: Nested 'it' blocks should throw a useful error


juliemr:
Version 2.5.3 and previous, all environments.

Nested `it` blocks are unsupported and meaningless, but sometimes test writers mess up and make them anyway (for example, if they have helper functions). It would be nice if Jasmine threw an error when this happens, instead of silently passing. For example, something like:

```js
describe('nested its', function() {
  it('should throw some sort of error', function() {
    it('erroneous nested it', function() {
      expect(true).toEqual(false);
    });
  });
});
```

Should throw something like `unexpected "it" statement inside a test`.

slackersoft:
This seems reasonable as we've moved to some nicer error messages for cases like this. I'd be happy to review a pull request to have `it` throw an error if Jasmine is already executing another `it`.

Goodwine:
Note that the problem is not only nested `it`s, but also nested `beforeEach`, `afterEach`, etc.

ksvitkovsky:
@slackersoft I started to work on the issue, should I keep the issue just about nested `it` statements or expand that to describes/befores/afters that are being used not on top-level of `describe` function?

slackersoft:
To clarify, anything can be nested inside of a `describe` (including another `describe`), but nothing should be nested inside of an `it`, `beforeEach`, `afterEach`, `beforeAll`, or `afterAll`. Possibly the best way to block this is just to have all of these top-level interface methods throw an error if they are called while the suite is executing.

sgravrock:
This was released in 2.9.0. Thanks, @ksvitkovsky!

PR #1411, opened 2017-08-09
Throw an error on methods in it/befores/afters

ksvitkovsky:
## Description
Added check on each method call to ensure that no suite is currently executing.

## Motivation and Context
As described in #1295 methods nested in it/before/after are not supported and jasmine does not explicitly tell that if user have them.

## How Has This Been Tested?
Added new test.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/Env.js
+++ b/src/core/Env.js
@@ -384,7 +384,8 @@ getJasmineRequireObj().Env = function(j$) {
       }
     };
 
-    function ensureIsNotNested(method, runnable) {
+    function ensureIsNotNested(method) {
+      var runnable = currentRunnable();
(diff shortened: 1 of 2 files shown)


src/core/Env.js
```diff
@@ -327,6 +327,12 @@ getJasmineRequireObj().Env = function(j$) {
       }
     };
 
+    function ensureIsNotNested(method, currentSuite) {
```
slackersoft: I think it's a bit confusing here to cover the outer `currentSuite` which is a function, especially if all of the callers are calling the same thing as a parameter. Additionally, we should probably be using `currentRunnable` for consistency unless there is a reason to only check for `currentSuite` instead.

PR #1427, opened 2017-10-11
Fix #1426 clearTimeout not correctly clearing a timeout

### LINKED ISSUE #1426: fake clock does not allow one timer to clear another when they are scheduled for the same time
The following test should pass, but it fails:

```js
it('one timer can clear another even if they expire concurrently', () => {
  jasmine.clock().install();

  let timerId2;

  setTimeout(() => {
    console.log('timer1');
    expect(timerId2).not.toBeUndefined();
    clearTimeout(timerId2);
  }, 100);

  timerId2 = setTimeout(() => {
    console.log('timer2');
    fail();
  }, 100);

  jasmine.clock().tick(1000);

  jasmine.clock().uninstall();
});
```

I can see an argument that the order of these two might not be guaranteed, or shouldn't be relied upon, so maybe timer2 could fire first so timer1 can't clear it. However, the log output shows that timer1 does in fact fire first, so timer2 was cleared, and should not fire.

Tested using 2.8.0.

leahciMic:
## Description

`clearTimeout` was not correctly handling the case of clearing a timeout that is also scheduled to run at the same tick.

This fix adds a `deletedKeys` array that is checked whilst we are running the scheduled functions for the current clock tick. If a function exists in `deletedKeys` it will not run. `deletedKeys` is then reset to an empty array.

## Motivation and Context

Fixes #1426 

## How Has This Been Tested?

Tests were added to confirm the issue presented in #1426 and changes were made to ensure they pass and no other tests break.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/DelayedFunctionScheduler.js
+++ b/src/core/DelayedFunctionScheduler.js
@@ -129,9 +129,7 @@ getJasmineRequireObj().DelayedFunctionScheduler = function() {
 
         currentTime = newCurrentTime;
 
-        var funcsToRun = scheduledFunctions[currentTime].sort(function (a, b) {
-          return a.millis > b.millis;
-        });
+        var funcsToRun = scheduledFunctions[currentTime];
 
         delete scheduledFunctions[currentTime];
 
--- a/src/core/DelayedFunctionScheduler.js
+++ b/src/core/DelayedFunctionScheduler.js
@@ -5,6 +5,7 @@ getJasmineRequireObj().DelayedFunctionScheduler = function() {
     var scheduledFunctions = {};
     var currentTime = 0;
     var delayedFnCount = 0;
+    var deletedKeys = [];
 
     self.tick = function(millis, tickDate) {
       millis = millis || 0;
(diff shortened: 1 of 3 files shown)


src/core/DelayedFunctionScheduler.js
```diff
@@ -126,7 +129,10 @@ getJasmineRequireObj().DelayedFunctionScheduler = function() {
 
         currentTime = newCurrentTime;
 
-        var funcsToRun = scheduledFunctions[currentTime];
+        var funcsToRun = scheduledFunctions[currentTime].sort(function (a, b) {
```
sgravrock: JavaScript's `sort` is not guaranteed to be stable, so this potentially causes functions that are scheduled for the same time to run in the wrong order. I think that's why the tests [fail in PhantomJS](https://travis-ci.org/jasmine/jasmine/jobs/286512581) even though they pass in other browsers.

I see that that the same test fails in other browsers if this `sort` is removed. If that's the only reason for the `sort`, a better approach might be to change the test to explicitly schedule the one function slightly after the other.

leahciMic: @sgravrock Oh I think you're right :) I'll fix this up today. Sorry I fired off this PR late at night and sort of forgot about it.

PR #1435, opened 2017-10-24
Add support for jasmine.any(Symbol).

### LINKED ISSUE #1431: jasmine.any(Symbol) does not match Symbol instances
## Expected Behavior
`symbolObj instanceof Symbol` actually returns `false`, so perhaps this is working as intended (even if a bit non-intuitive). But I'd expect a symbol instance to be treated as equal to `jasmine.any(Symbol)`.

## Current Behavior
Symbol instances are not equal to `jasmine.any(Symbol)`.

## Possible Solution
We could special-case Symbols in the equality checker to work with `.any(Symbol)`.

https://codepen.io/anon/pen/rGoXMz

## Your Environment

* Version used: 2.8.0
* Environment name and version (e.g. Chrome 39, node.js 5.4): Chrome 61
* Operating System and version (desktop or mobile): Linux

voithos:
## Description
The original asymmetric matcher code for Any did not work with symbols
since `symbolInstance instanceof Symbol` is actually `false` (but,
`symbolInstance.constructor` is `Symbol). This simply adds an extra
clause that explicitly checks for symbol (if available) like the other
primitive types.

Also added some missing specs for other types, like Map, Set, etc.

Fixes #1431.

## Motivation and Context
Before this change `jasmine.any(Symbol)` would not match `Symbol` instances.

## How Has This Been Tested?
Executed tests in-browser as well as in node, before and after `grunt buildDistribution`. Importantly, have not run the tests in IE yet.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/asymmetric_equality/AnySpec.js
+++ b/spec/core/asymmetric_equality/AnySpec.js
@@ -29,6 +29,38 @@ describe("Any", function() {
     expect(any.asymmetricMatch(true)).toBe(true);
   });
 
+  it("matches a Map", function() {
+    jasmine.getEnv().requireFunctioningMaps();
+
+    var any = new jasmineUnderTest.Any(Map);
+
+    expect(any.asymmetricMatch(new Map())).toBe(true);
+  });
+
+  it("matches a Set", function() {
+    jasmine.getEnv().requireFunctioningSets();
+
+    var any = new jasmineUnderTest.Any(Set);
+
+    expect(any.asymmetricMatch(new Set())).toBe(true);
+  });
+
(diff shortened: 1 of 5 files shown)


voithos:
One thing to note: this triggers a JSHint warning about an invalid `typeof` value `'symbol'` - but `'symbol'` is actually valid as of ES6. I tried updating `grunt-contrib-jshint` locally to v1.1.0, but the warning still remains (I think, because the default target version is ES3 or ES5, so the warning is still valid for older versions of ECMAScript, but JSHint isn't able to see that the offending line is guarded).

PR #1446, opened 2017-11-02
Add toHaveLength matcher

emilos:
## Description

Thanks for the lib :+1:

I've added an additional matcher called `toHaveLength` which can be helpful for arrays and strings.

## Motivation and Context

Minor utility for length checking. In my use case it would also makes it easier to switch from `jest` to `jasmine`.

The matcher could be extended to support any object that has a length property, I didn't add it because I'm not sure it's that would be a common use case.

## How Has This Been Tested?

`npm t -s`

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/spec/core/matchers/toHaveLengthSpec.js
+++ b/spec/core/matchers/toHaveLengthSpec.js
@@ -1,0 +1,32 @@
+describe("toHaveLength", function() {
+  it("fails for invalid lengths", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveLength(),
+      result;
+
+    result = matcher.compare(['foo'], 0);
+    expect(result.pass).toBe(false);
+
+    result = matcher.compare('hello', 4);
+    expect(result.pass).toBe(false);
+  });
+
+  it("succeeds for valid lengths", function() {
+    var matcher = jasmineUnderTest.matchers.toHaveLength(),
+      result;
+
+    result = matcher.compare('foo bar baz', 11);
+    expect(result.pass).toBe(true);
+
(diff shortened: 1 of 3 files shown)


sgravrock:
Can you help me understand the benefit of this matcher over using `toEqual` on the length, besides Jest compatibility?

emilos:
1)

```js
expect(user.id).toEqual(1)
expect(user.guid.length).toEqual(32)
```

vs

```js
expect(user.id).toEqual(1)
expect(user.guid).toHaveLength(32)
```

2)

```js
let { id, guid } = user
expect(id).toEqual(1)
expect(guid.length).toEqual(32)
```

```js
let { id, guid } = user
expect(id).toEqual(1)
expect(guid).toHaveLength(32)
```

It's not a big difference, I find the second one easier a little bit faster to understand at a quick glance.

slackersoft:
Jasmine does support [custom matchers](https://jasmine.github.io/edge/custom_matcher.html) if you want to add this globally to your suite to aid in conversion for now. I'm not sure a `toHaveLength` matcher is something we're looking for in Jasmine core at this time. It's a bit specific, and easily implementable in a suite that feels the need for it.

emilos:
Thanks, that's fine. I think that the lib would be a little bit easier to adopt in more places if the core api would be a little bit broader, but that's a very opinionated measure :). There are matchers available like `toBeCloseTo` or `toBeGreaterThanOrEqual` which imo have the same or least probability of usage than the one presented above.

Thanks again, closing then :)

PR #1448, opened 2017-11-10
Allowed async functions to be passed into SpyStrategy#callFake

mrlannigan:
## Description
Allow async functions as stubs for spies. Also included updates to run the tests in v4, v8, v9 node versions.

## Motivation and Context
A spy's callFake would not accept an `async function`. This change proposes that it should indeed allow AsyncFunctions.

## How Has This Been Tested?
By adding additional specs to cover this new functionality. This creates minimal impact to any existing code.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -92,23 +92,22 @@ describe("SpyStrategy", function() {
     expect(returnValue).toEqual(67);
   });
 
-  it("allows a fake async function to be called instead", async function(done) {
-    try {
-      var originalFn = jasmine.createSpy("original"),
-          fakeFn = jasmine.createSpy("fake").and.callFake(async function () { return 67; }),
-          spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn}),
-          returnValue;
+  it("allows a fake async function to be called instead", function(done) {
+    jasmine.getEnv().requireAsyncAwait();
+    var originalFn = jasmine.createSpy("original"),
+        fakeFn = jasmine.createSpy("fake").and.callFake(eval("async () => { return 67; }")),
+        spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn}),
+        returnValue;
 
-      spyStrategy.callFake(fakeFn);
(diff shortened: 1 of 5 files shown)


DaRizat:
:+1:

mrlannigan:
Question for the maintainers. How far back in node versions do you all wish to continue to support?

sgravrock:
We'd like Jasmine 2.x to continue supporting v0.12.18 for now. We also support IE 8-11, none of which support `async`. You should be able to guard your spec with `jasmine.getEnv().requireAsyncAwait()` so that it only runs on platforms that have that feature. There are some examples in EnvSpec.js.

spec/core/SpyStrategySpec.js
```diff
@@ -92,16 +92,40 @@ describe("SpyStrategy", function() {
     expect(returnValue).toEqual(67);
   });
 
+  it("allows a fake async function to be called instead", async function(done) {
```
sgravrock: `async` here is unnecessary because the spec function takes a callback, and it'll break on platforms that don't support it. In general, asynchronous specs should do any one of: take a callback, explicitly return a promise, or be declared `async`. But in Jasmine's own test suite we need to stick to the callback style.

spec/core/SpyStrategySpec.js
```diff
@@ -92,16 +92,40 @@ describe("SpyStrategy", function() {
...
+    try {
+      var originalFn = jasmine.createSpy("original"),
+          fakeFn = jasmine.createSpy("fake").and.callFake(async function () { return 67; }),
```
sgravrock: Using `async` here is fine if you guard the spec with `jasmine.getEnv().requireAsyncAwait()` so that this part won't execute on platforms that don't support it.

clok:
👍

mrlannigan:
@sgravrock Sounds good. I made changes to ensure this spec can run and be ignored in an env that doesn't support async/await. I also added a change to the travis build to run the specs in various major node versions. I do believe this change will satisfy https://www.pivotaltracker.com/story/show/152685778

PR #1460, opened 2017-12-08
Added matchers: truthy, falsy, empty and notEmpty

sjolicoeur:
Added Matchers to enable testing against:
- truthy values
- falsy values
- empty values
- notEmpty values

## Motivation and Context
We wanted to be able to check in react if props were `falsy` and we could not use `any`  so we created truthy as the opposite. The same went for `empty` which is a more specific case of `falsy` in some cases and would return true for cases like empty arrays or objects, which are truthy.

## How Has This Been Tested?
Included tests see PR

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/asymmetric_equality/AsymmetricEqualitySpec.js
+++ b/spec/core/asymmetric_equality/AsymmetricEqualitySpec.js
@@ -1,0 +1,40 @@
+describe("Fuzzy Equalities", function() {
+  describe('Truthy/Falsy equalities', function() {
+    it('allows the test to pass', function() {
+      expect({
+        name: "Rex",
+        id: null,
+        description: undefined,
+        age: 42,
+        tags: [],
+      }).toEqual({
+        name: jasmineUnderTest.truthy(),
+        id: jasmineUnderTest.falsy(),
+        description: jasmineUnderTest.falsy(),
+        age: jasmineUnderTest.truthy(),
+        tags: jasmineUnderTest.truthy()
+      })
+    })
+  });
+
+  if (typeof Set !== 'undefined') {
+    describe('empty/notEmpty equalities', function () {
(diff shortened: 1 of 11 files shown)


slackersoft:
Can you help me understand a bit more of the use case for these matchers? Typically, we think of asymmetric equality matchers as things you would need when matching arguments when calling a `Spy`. 

If you want to check the properties on an object, you should be able to use the `toBeTruthy` and `toBeFalsy` matchers for those concepts.

Jasmine doesn't have a matcher for `toBeEmpty`, but I think I would rather see that as a matcher than asymmetric equality. Considering the parameter matcher problem, emptiness seems to general when verifying calls on a spy, I probably want to know that I passed an empty `Array` and not a `Set`, `Map`, `object`, or `string`

rdy:
Hey @slackersoft I'd like to chime in on this. I think the reasoning for having truthy and falsy is reasonable. Very often I like to use object equality because the generated error message on what is wrong is better than testing the property directly. For example:
```
var foo ={bar: null};

expect(foo.bar).toTruthy();
// expected null to be truthy

expect(foo).toEqual({bar: jasmine.truthy()});
// expected {bar: null} to be {bar: <truthy> }

```

In the second case the error message is far more readable as I don't need to think about which property on the object is the problem, it is part of the error message. This is mostly useful in cases where we are using jasmine.objectContaining where this kind of pattern really shines.

Having the asymmetric equality on empty and not empty I can see being more controversial. This is usually used in something in react where you are testing props on a react component. I can't think of many cases where you might want to use not empty but I asked @sjolicoeur to add it for completeness since without it you wouldn't be able to make certain kinds of assertions.

I think that some other members of the apps manager team  on CF can speak up to why this makes sense in their use case but I wanted to at least describe the reasoning for this PR.

Issue #1469, opened 2017-12-19
Fake setTimeout/setInterval is missing Node methods


chris--young:
I'm running into a problem trying to write tests for a Node project that uses [the brakes module](https://github.com/awolden/brakes) since it uses [Node's Timeout.unref() method](https://nodejs.org/dist/latest-v9.x/docs/api/timers.html#timers_timeout_unref) and Jasmine's fake setInterval returns a number.

## Expected Behavior

Running test shouldn't cause uncaught exceptions

## Current Behavior

Tests cause this uncaught exception `TypeError: this._spinningInterval.unref is not a function`

## Possible Solution

Return a fake Timeout object if we're using Node. I don't mind doing this myself if people are cool with it.

## Context

I'm trying to switch our circuit breaker implementation, the new module we want to use seems incompatible with Jasmine

## Environment
* Version used: 2.8.0
* Environment name and version (e.g. Chrome 39, node.js 5.4): Node
* Operating System and version (desktop or mobile): OS X
* Link to your project: https://github.com/Tinder/sharded-redis-client

PR #1470, opened 2017-12-20
Use timeout objects when in node

chris--young:
Fixes: https://github.com/jasmine/jasmine/issues/1469

## Description
When run in Node, setTimeout and setInterval should return an instance of Timeout instead of a number

## Motivation and Context
A Node dependency that used `Timeout.unref()` was unable to be tested with Jasmine

## How Has This Been Tested?
Specs pass in both browser and Node, fixes problem in my project

## Types of changes

## Checklist:

### DIFF
--- a/src/core/Clock.js
+++ b/src/core/Clock.js
@@ -1,4 +1,7 @@
 getJasmineRequireObj().Clock = function() {
+
+  var NODE_JS = typeof process !== 'undefined' && process.versions && typeof process.versions.node === 'string';
+
   /**
    * _Note:_ Do not construct this directly, Jasmine will make one during booting. You can get the current clock with {@link jasmine.clock}.
    * @class Clock
@@ -22,6 +25,7 @@ getJasmineRequireObj().Clock = function() {
       delayedFunctionScheduler,
       timer;
 
+    self.FakeTimeout = FakeTimeout;
 
     /**
      * Install the mock clock over the built-in methods.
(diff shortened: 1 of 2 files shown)


sgravrock:
This makes sense overall, and it's something we'd like to have in Jasmine. There are some things I'd like to see changed. I'll add some review comments.

You marked this as a breaking change. Could you elaborate? I can see that it changes the return type of mocked `setTimeout` and friends when running in Node, but I'm having a hard time picturing a test suite that both relies on those functions returning numbers and works with code under test that actually runs on Node.

src/core/Clock.js
```diff
@@ -1,3 +1,9 @@
+try {
+  var NODE_JS = typeof process.versions.node === 'string';
```
sgravrock: Browsers don't have file scope, so this creates a global variable when run in a browser. It also throws and catches an exception. We'd like to avoid doing that as much as possible because it makes it harder for users to debug their own code using exception breakpoints. Something like `typeof process !== 'undefined' && process.versions && typeof process.versions.node === 'string'` would work better than try/catch. (Here and in ClockSpec.js as well.)

src/core/Clock.js
```diff
@@ -167,3 +177,18 @@ getJasmineRequireObj().Clock = function() {
 
   return Clock;
 };
+
+/**
+ * Stubs Node.js Timeout class
+ */
+function Timeout(id) {
```
sgravrock: This leaks a `timeout` global when run in a browser.

sgravrock: In addition to not making it a global, giving it a name like `StubTimeout` would also help with debugging.

src/core/Clock.js
```diff
@@ -167,3 +177,18 @@ getJasmineRequireObj().Clock = function() {
...
+  this.id = id;
```
sgravrock: I'd rather not give these fake `Timeout` objects an `id` property that the real ones don't have. It looks like all we're doing with it is passing it back to `delayedFunctionScheduler.removeFunctionWithId`. A better approach might be to use the `Timeout` object itself as a timeout key, using the optional 5th argument to `delayedFunctionScheduler.scheduleFunction`. Does that make sense?

chris--young:
> You marked this as a breaking change. Could you elaborate? I can see that it changes the return type of mocked setTimeout and friends when running in Node, but I'm having a hard time picturing a test suite that both relies on those functions returning numbers and works with code under test that actually runs on Node.

Yea this is cause I figured the different return type could maybe cause problems for some poorly written tests. I guess it's not really breaking.

sgravrock:
Looks good. Can you squash your three commits into one? Other than that I think this is ready to merge.

PR #1473, opened 2018-01-04
Do not execute pending specs

### LINKED ISSUE #1265: Using --grep will call every defined beforeAll() in your suite/spec
Hello!

Essentially when users use the --grep option to delineate which tests run (which is a fantastic feature), it will run every beforeAll() in the file where matches are found, resulting in many superfluous beforeAll()s being triggered. This does not happen with beforeEach().

I reported this issue within the Protractor project, but it was tagged as possibly as issue with Jasmine (and not Protractor).

For much more detail (including full code examples), please see the original issue: https://github.com/angular/protractor/issues/3952

chrismwendt:
## Description

This changes the implementation of `isExecutable` so that pending specs and suites without any executable specs are considered non-executable.

## Motivation and Context

Previously, suites would needlessly execute beforeAll and afterAll when all of the specs were pending. This was causing lots of setup/teardown to occur for no reason.

Fixes #1265

#1225 is similar, but was only concerned with **disabled** suites. In contrast, this change is concerned with **enabled** suites that have no executable specs or suites inside.

## How Has This Been Tested?

- This repo's test suite
- Monkey patching Jasmine in an app that uses Protractor

## Types of changes

This changes the conditions under which `specDone` executes, and probably has other side effects I'm not aware of at the moment.

## Checklist:

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -1400,7 +1400,7 @@ describe("Env integration", function() {
         status: 'passed'
       }));
 
-      expect(reporter.specDone).toHaveBeenCalledWith(jasmine.objectContaining({
+      expect(reporter.specDone).not.toHaveBeenCalledWith(jasmine.objectContaining({
         description: "with an x'ed spec",
         status: 'pending'
       }));
@@ -1410,7 +1410,7 @@ describe("Env integration", function() {
         status: 'failed'
       }));
 
-      expect(reporter.specDone).toHaveBeenCalledWith(jasmine.objectContaining({
+      expect(reporter.specDone).not.toHaveBeenCalledWith(jasmine.objectContaining({
         description: 'is pending',
         status: 'pending'
       }));
--- a/src/core/Suite.js
+++ b/src/core/Suite.js
@@ -82,7 +82,7 @@ getJasmineRequireObj().Suite = function(j$) {
   };
 
   Suite.prototype.isExecutable = function() {
-    return !this.markedPending;
+    return !this.markedPending && this.children.some(function(child) { return child.isExecutable(); });
   };
 
   Suite.prototype.canBeReentered = function() {
--- a/spec/core/SpecSpec.js
+++ b/spec/core/SpecSpec.js
@@ -422,13 +422,13 @@ describe("Spec", function() {
     expect(spec.isExecutable()).toBe(false);
   });
 
-  it("should be executable when pending", function() {
+  it("should not be executable when pending", function() {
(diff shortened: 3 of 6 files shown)


spec/core/integration/EnvSpec.js
```diff
@@ -1400,7 +1400,7 @@ describe("Env integration", function() {
         status: 'passed'
       }));
 
-      expect(reporter.specDone).toHaveBeenCalledWith(jasmine.objectContaining({
+      expect(reporter.specDone).not.toHaveBeenCalledWith(jasmine.objectContaining({
```
sgravrock: Does this mean that we'd no longer be reporting pending specs? That would be a problem. We're looking to improve reporting of pending specs (see e.g. #1418), not remove it.

src/core/Suite.js
```diff
@@ -82,7 +82,7 @@ getJasmineRequireObj().Suite = function(j$) {
   };
 
   Suite.prototype.isExecutable = function() {
-    return !this.markedPending;
+    return !this.markedPending && this.children.some(function(child) { return child.isExecutable(); });
```
sgravrock: I don't think that all of the browsers that we support have `.some`.

spec/core/integration/EnvSpec.js
chrismwendt: Currently, yes. However, change could probably be amended such that `specDone` is still called for pending specs, but `beforeAll` is not. I am not very familiar with this code, but perhaps you would be able to make such an improvement.

slackersoft: I like the idea of making sure we don't run setup code that we don't need, but we definitely don't want to merge this pull request if it is going to break reporting for pending specs.

eileenzheng:
i really hope this will fix #1265! It was closed as a duplicate of #1175 that was fixed with #1225, but it really isn't. I have jasmine 2.8.0 and with that #1175 is fixed but #1265 is still happening. cc @slackersoft

slackersoft:
@eileenzheng as [I mentioned](https://github.com/jasmine/jasmine/issues/1265#issuecomment-302806048) in #1265, the `--grep` flag isn't part of Jasmine itself, so we don't know how it is calling into Jasmine to accomplish its task, or if it is even using a new enough version of Jasmine to have that fix.

sgravrock:
I'm closing this since the underlying issue was fixed in 3.0. Feel free to open a new issue or PR if there's anything that we missed.

anushreetp:
I am working with jasmine 3.1.0 and still facing the same issue.
beforeAll and afterAll of pending specs are running.

PR #1498, opened 2018-02-08
allow adding a deprecation object

UziTech:
## Description

Allow `jasmine.getEnv().deprecated() to take an error object instead of a string

## Motivation and Context

Trying to add a deprecation with an error object so I can use it's stack instead of the the `deprecationWarning.stack` starting at `env.deprecated`

## How Has This Been Tested?

added integration test

## Types of changes

## Checklist:

I tried running `grunt buildDistribution` but it gave me the error `Warning: not found: bundle Use --force to continue.`

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -2349,11 +2349,11 @@ describe("Env integration", function() {
     var env = new jasmineUnderTest.Env(),
       exceptionFormatter = new jasmineUnderTest.ExceptionFormatter(),
       reporter = jasmine.createSpyObj('reporter', ['jasmineDone', 'suiteDone', 'specDone']),
-      topLevelError = new Error('top level deprecation'),
-      suiteLevelError = new Error('suite level deprecation'),
-      specLevelError = new Error('spec level deprecation');
+      topLevelError, suiteLevelError, specLevelError;
 
-
+      try { throw new Error('top level deprecation') } catch (err) { topLevelError = err; }
+      try { throw new Error('suite level deprecation') } catch (err) { suiteLevelError = err; }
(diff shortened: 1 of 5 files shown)


sgravrock:
PhantomJS doesn't set the `stack` property of error objects until they're thrown. The error objects in the code under test are thrown but the ones in the test are not, so the latter's `stack` values are null and the comparison fails.

slackersoft:
Additionally, Travis-CI only runs the Node.js and PhantomJS branches because the real browsers all require SauceLabs and thus secure environment variables, and are thus as disabled as we can make them for PRs.

I would also like to hear a bit more about why you need this functionality. The `deprecated` method isn't really intended for usage outside of Jasmine-Core as it more a way to mark certain features of Jasmine as deprecated.

UziTech:
I maintain [`atom-jasmine2-test-runner`](https://github.com/UziTech/atom-jasmine2-test-runner) and have been looking for a way to list [deprecations](https://github.com/UziTech/atom-jasmine2-test-runner/blob/master/lib/spec-helper/deprecations.js#L50) in Atom without showing them as errors.

I just thought it would be useful to combine the `deprecationWarnings` in Jasmine with the deprecations in Atom. Atom already has a way to list deprecations but when I use `env.deprecated` the stack trace points to the `env.deprecated` call instead of the deprecation in Atom.

UziTech:
Any chance we can backport this to 2.99.x?

slackersoft:
The whole point of the 2.99 release(s) is to help people upgrade to 3.0. They aren't really intended for long term usage. I'm inclined to just leave it in 3.0 and try to help everyone update to a version we're going to support with further development. Sorry.

UziTech:
Sounds good. Thanks for the response.

PR #1516, opened 2018-02-23
Display error properties for failed specs

jbunton-atlassian:
## Description
If a spec fails due to an uncaught exception or unhandled promise rejection then Jasmine currently prints the error message and stack. This is missing potentially very helpful information that may be in other custom properties on the error.

Happy to make any changes you suggest. Thanks for taking a look :)

## Motivation and Context
For example imagine a test like this:
```
async function makeHttpRequest(...) {
  // do stuff
  const error = new Error('Unexpected response from server!');
  error.statusCode = ...;
  error.body = ...;
  throw error;
}

it('talks to the server', async () => {
  const response = await makeHttpRequest(...);
  expect(response.body).toEqual('ok');
});
```

Having the `statusCode` and `body` properties printed in the spec failure text greatly simplifies debugging. Currently I have to catch the error and insert a console.log, or use a debugger. This is especially useful for sporadic test failures.

## How Has This Been Tested?
I ran the tests on my Arch Linux machine in the following environments:
- Firefox 58.0.2 (`bundle exec rake jasmine`)
- Chrome 64.0.3282.167 (`bundle exec rake jasmine`)
- Node v9.5.0 (`npm test`)

All the tests passed except for one failure in Firefox. Note that this test also fails for me on master, so this is not caused by my changes.
- `Env integration > with a mock clock > should wait a custom interval before reporting async functions that fail to call done`

## Types of changes

## Checklist:

### DIFF
--- a/src/core/ExceptionFormatter.js
+++ b/src/core/ExceptionFormatter.js
@@ -29,12 +29,16 @@ getJasmineRequireObj().ExceptionFormatter = function(j$) {
 
       var stackTrace = new j$.StackTrace(error.stack);
       var lines = filterJasmine(stackTrace);
+      var result = '';
 
       if (stackTrace.message) {
         lines.unshift(stackTrace.message);
       }
 
-      return lines.join('\n');
+      result += formatProperties(error);
+      result += lines.join('\n');
+
+      return result;
     };
 
     function filterJasmine(stackTrace) {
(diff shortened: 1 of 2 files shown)


src/core/ExceptionFormatter.js
```diff
@@ -51,6 +55,32 @@ getJasmineRequireObj().ExceptionFormatter = function(j$) {
...
+        if (ignored.includes(prop)) {
+          continue;
+        }
+        result[prop] = error[prop];
+        empty = false;
+      }
+
+      if (!empty) {
+        try {
+          return 'error properties: ' + JSON.stringify(result, null, 2) + '\n';
```
slackersoft: I think we probably just want to use Jasmine's own pretty printer here, which will still print something out even with a cycle in the object.

jbunton-atlassian:
Thanks @slackersoft.

I've changed to use `j$.pp()` as you suggested. This meant I could just remove the cycle test, so that's nice.

Is there any way to configure the pretty printer to use more than one line for complex objects?

slackersoft:
Jasmine's pretty printer doesn't have any formatting options to it at this time, but I think it's still the way to go so that everything gets formatted similarly from Jasmine.

jbunton-atlassian:
Thanks!

PR #1567, opened 2018-06-12
Improve Jasmine timeout error message

ikonst:
We're addressing two issues:
1) When an async spec times out, Jasmine always reports an error saying the async callback was not invoked within the default timeout interval. This can be misleading when a custom timeout is passed as the 3rd argument to `it`, `beforeAll`, etc. 
2) Jasmine fails to report the timeout interval that was in effect, which can be valuable since the interval can be custom, and the default timeout too can be altered within the test code.

With this change, we make a distinction between custom timeouts and default timeouts, and we surface what was the actual timeout within the error message, e.g.
```
Error: Timeout - Async callback was not invoked within custom timeout (of 20ms)
```
or:
```
Error: Timeout - Async callback was not invoked within jasmine.DEFAULT_TIMEOUT_INTERVAL (of 100ms)
```


ikonst:
👀 @slackersoft

slackersoft:
Merged

Issue #1442, opened 2017-10-28
extend createSpyObj to support properties


Plondrein:
## Expected Behavior
Currently i cant create nicely spy object full of methods, but what if i need to mix some properties and methods?

## Current Behavior
currently createSpyObj creates only methods 

## Possible Solution
another array argument 'propertyNames' just like ;methodNames' could solve problem.
Properties would return default values until i use spyOnProperty.

## Context
That's my typescript,  i hope it could be less ugly:

```typescript
let mockRouter: jasmine.SpyObj<Router>;

mockRouter = jasmine.createSpyObj('mockRouter', ['navigateByUrl']);

mockRouter = {
  ...mockRouter,
  events: Observable.of(...routerEvents),
} as jasmine.SpyObj<Router>;
```

sgravrock:
That seems like a nice feature to have. I'd be happy to review a pull request that adds it.

ninjamyst:
This would be very helpful as we have many objects we want to mock that have getter properties.

sergiojoker11:
Is this achievable? This feature would make code more readable... and our life easier in turn.

sergiojoker11:
https://github.com/jasmine/jasmine/pull/1569

felpel:
@sergiojoker11 any updates on your end to complete your current PR? Would be very neat to have this feature out of the box, my test classes would be cleaner.

sergiojoker11:
@felpel Yes, I am planning on addressing the comments later this week. I swear I will!

szabolcssimonyi:
any progress on this? this feature would be very usefull

slackersoft:
@szabolcssimonyi  You can see the current state of @sergiojoker11's pull request #1569.

MarcinusX:
Is it possible to achieve the same result until the feature is released?

benelliott:
@MarcinusX there are workarounds in the original post and my comment above.

MarcinusX:
@benelliott Oh! I understand now! I thought that `events` was something specific to the issue but it is just overriding the field. Thank you!

rcavezza:
Does this pull request need someone to come in and finish it? This is something I can use in my codebase and it looks like this is close to the finish line, but fell through the cracks a few months ago.

sergiojoker11:
@rcavezza Yes, it does need someone to finish it. By the time when the code review was done, I didn't understand the proposed changes and didn't have the time. So, please, feel free to take it over and finish it.

[...]

PR #1569, opened 2018-06-18
Create spy obj with also a properties name object/array

sergiojoker11:
## Description

It enlarges the power of the method ´createSpyObj´ by adding an extra parameter that represents the properties to be spied.

## Motivation and Context

The motivation is to improve the readability of the unit tests and provide the capability of spying the dependencies(methods as well as properties) of the code under test in one go - with one liner.

This feature was marked as ´ready to work´ here: https://github.com/jasmine/jasmine/issues/1442

## How Has This Been Tested?

Some more tests have been added to the spec to test this new feature. 
The existing ones have been updated where it made sense do it so. Obviously, they all pass.

## Types of changes

## Checklist:

*Note*: No updates have been carried out on the documentation as I could not find any reference to it and I am not very familiar with JS projects. Nonetheless, I am more than happy to do it - it should not take very long.

### DIFF
--- a/spec/core/SpySpec.js
+++ b/spec/core/SpySpec.js
@@ -140,7 +140,15 @@ describe('Spies', function () {
       expect(spyObj.property2.and.identity).toEqual('BaseName.property2');
     });
 
-    it("should allow you to omit the baseName", function() {
+    it("should allow you to omit the baseName without propertyNames parameter", function() {
+      var spyObj = env.createSpyObj(['method1', 'method2']);
+
+      expect(spyObj).toEqual({ method1: jasmine.any(Function), method2: jasmine.any(Function)});
+      expect(spyObj.method1.and.identity).toEqual('unknown.method1');
+      expect(spyObj.method2.and.identity).toEqual('unknown.method2')
+    });
+
+    it("should allow you to omit the baseName with propertyNames parameter", function() {
       var spyObj = env.createSpyObj(['method1', 'method2'], ['property1', 'property2']);
 
       expect(spyObj).toEqual({ method1: jasmine.any(Function), method2: jasmine.any(Function), property1: jasmine.any(Function), property2: jasmine.any(Function)});
--- a/spec/core/SpySpec.js
+++ b/spec/core/SpySpec.js
@@ -77,7 +77,7 @@ describe('Spies', function () {
 
       for (var arity = 0; arity < functions.length; arity++) {
         var someFunction = functions[arity],
-            spy = env.createSpy(someFunction.name, someFunction);
+          spy = env.createSpy(someFunction.name, someFunction);
 
         expect(spy.length).toEqual(arity);
       }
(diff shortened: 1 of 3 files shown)


spec/core/SpySpec.js
```diff
@@ -116,12 +132,22 @@ describe('Spies', function () {
...
+      var spyObj = env.createSpyObj('BaseName', ['method1', 'method2'], ['property1', 'property2']);
+
+      expect(spyObj).toEqual({ method1: jasmine.any(Function), method2: jasmine.any(Function), property1: jasmine.any(Function), property2: jasmine.any(Function)});
+      expect(spyObj.property1.and.identity).toEqual('BaseName.property1');
+      expect(spyObj.property2.and.identity).toEqual('BaseName.property2');
+    });
+
     it("should allow you to omit the baseName", function() {
-      var spyObj = env.createSpyObj(['method1', 'method2']);
+      var spyObj = env.createSpyObj(['method1', 'method2'], ['property1', 'property2']);
```
slackersoft: I would like to see this test added as a separate test to ensure that the previous functionality of only providing methods, without basename or properties, still works correctly.

src/core/SpyFactory.js
```diff
@@ -1,45 +1,61 @@
...
-  return SpyFactory;
+    function SpyFactory(getCustomStrategies) {
+        var self = this;
+
+        this.createSpy = function(name, originalFn) {
+            return j$.Spy(name, originalFn, getCustomStrategies());
+        };
+
+        this.createSpyObj = function(baseName, methodNames, propertyNames) {
+            var baseNameOrMethodNamesAreCollections = j$.isObject_(baseName) || j$.isArray_(baseName) || j$.isObject_(propertyNames) || j$.isArray_(propertyNames);
```
slackersoft: It looks like this variable now also has some knowledge about the state of the propertyNames as well. I'd like to see it renamed or lose that knowledge.

src/core/SpyFactory.js
```diff
@@ -1,45 +1,61 @@
...
+                        obj[key] = self.createSpy(baseName + '.' + key);
+                        obj[key].and.returnValue(methodNames[key]);
+                        spiesWereSet = true;
+                    }
+                }
+            }
+
+            if (j$.isArray_(propertyNames)) {
+                for (var j = 0; j< propertyNames.length; j++) {
+                    obj[propertyNames[j]] = self.createSpy(baseName + '.' + propertyNames[j]);
```
slackersoft: I think this will create spy functions on the resulting object and not spy properties. Take a look at [SpyRegistry.js](https://github.com/jasmine/jasmine/blob/516fe6875959dffde38cd512f1b19af9a862f98c/src/core/SpyRegistry.js#L68) for how property spies are currently installed.

src/core/SpyFactory.js
sergiojoker11: This was a mistake, it should have been methodNames insteadof propertyNames

sergiojoker11: This wasn't even needed. A change in the line 13 does the trick.

spec/core/SpySpec.js
sergiojoker11: Done

src/core/SpyFactory.js
adriangodong: I'm trying to port this change on my code and managed to hit the same issue.

The spied property needs to be redefined using `Object.defineProperty` call. I believe you should be able to use `spyOnProperty` call to do this.

Afterwards, we need to expose the spy itself. Properties are not called like methods, so we can't use the same name. Contrast this with method spies: `spyObj.method` will return the spy, while `spyObj.method()` will call the spy. We want `spyObj.prop` to call the spy, and something else to return the spy.

To resolve this, we can store property spies into a dictionary and expose that. The end goal can be something like:
```
var propertySpies = { };

for (var j = 0; j< propertyNames.length; j++) {
  Object.defineProperty(obj, propertyNames[j], {
    configurable: true,
    get() { return undefined; }
  });

  propertySpies[propertyNames[j]] = spyOnProperty(obj, propertyNames[j]);
}

obj.propertySpies = propertySpies
```

And the following code should work:
```
var spiedObj = jasmine.createSpyObj("name", ["method"], ["prop"]);
spiedObj.propertySpies.prop.and.returnValue(1);
spiedObj.prop === 1; // true
```

slackersoft: I don't think you need to use `spyOn` or `spyOnProperty` when building up a spy object. To set up a property spy correctly, you will have to use `Object.defineProperty`, and probably 2 spies (one for `get` and one for `set`). I would err on the side of passing the spies directly in the `defineProperty` call and not calling `spyOnProperty` afterwards unless you have to.

`spyOn` and `spyOnProperty` do some extra bookkeeping to clean up spies, and that isn't needed (or wanted) for a one-off object like that created by `createSpyObj`.

epaitz:
What is the status of this open pull request?

Issue #1421, opened 2017-09-17
[Improvement] Spy on all functions in a class/object


yxliang01:
This is similar to issue #909. But, instead of just being able to have an object with spies on multiple functions with one function call, I would like to have a method like `spyOnAllMethods(Instance of some class/object)` which returns a spy object with all the functions of the provided instance are spies of the original functions.

Thanks!

slackersoft:
I think it might make sense to have an easier way to create a test double for an object. We would be happy to review a pull request that creates something like a `createDouble` that takes a constructor function and returns a new Jasmine spy object for all of the functions on an instance of that.

mjeanroy:
Hi,

It's certainly not perfect, but I already implemented this kind of utility function in a small library: take a look at the description [here](https://github.com/mjeanroy/jasmine-utils#utils-functions).

Not sure to have the time soon, but I would be happy to share it to jasmine core.

yxliang01:
@mjeanroy Nice one! This library looks fantastic. :)

bdirito:
I doesnt *spy* on them but 
``
proxy = new Proxy(
  () => {
    return proxy;
  },
  {
    get: function() {
      return proxy;
    }
  }
)
```
allows the mocking of anything. ie you can `mock.call().it.all.the().way.down()`

mohanrao:
Any plans to improve jasmine like the methods in jasmine-utils developed by @mjeanroy ?

aeisenberg:
I had some free time and added this functionality in the PR above. Is anyone available to take a look at the PR and provide feedback?

PR #1581, opened 2018-07-19
Add spyOnAll function

aeisenberg:
## Description

This function will spy on all writable and configurable properties of
an object that is passed in. It can be used like this:

    spyOnAll(obj);

## Motivation and Context
This commit addresses https://github.com/jasmine/jasmine/issues/1421

## How Has This Been Tested?
I have included specs for spyOnAll. I have run the tests in the browser through `bundle exec rake jasmine`.  The browsers I tested are chrome, firefox, safari, and vivaldi (latest versions of each).

## Types of changes

## Checklist:

I am not sure how to update the documentation. After I submit this PR, I will figure this out.

### DIFF
--- a/spec/core/SpyRegistrySpec.js
+++ b/spec/core/SpyRegistrySpec.js
@@ -214,6 +214,97 @@ describe("SpyRegistry", function() {
     });
   });
 
+  describe("#spyOnAllFunctions", function() {
+    it("checks for the existence of the object", function() {
+      var spyRegistry = new jasmineUnderTest.SpyRegistry();
+      expect(function() {
+        spyRegistry.spyOnAllFunctions(void 0);
+      }).toThrowError(/spyOnAllFunctions could not find an object to spy upon/);
+    });
+
+    it("overrides all writable and configurable functions of the object", function() {
+      var spyRegistry = new jasmineUnderTest.SpyRegistry({createSpy: function() {
+        return 'I am a spy';
+      }});
+      var createNoop = function() { return function() { /**/}; };
+      var noop1 = createNoop();
+      var noop2 = createNoop();
+      var noop3 = createNoop();
+      var noop4 = createNoop();
+      var noop5 = createNoop();
+
+      var parent = {
+        notSpied1: noop1
+      };
+      var subject = Object.create(parent);
+      Object.defineProperty(subject, 'spied1', {
+        value: noop1,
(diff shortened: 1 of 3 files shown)


aeisenberg:
I have added proper jsdoc. And it looks like that is all I need to do to update the documentation. For that reason, I am going to check that box in the PR checklist above.

adriangodong:
How do I access the spy on the spied property?

aeisenberg:
It would be something like this:

```
let obj = { a: () => true };
spyOnAll(obj);
obj.a();
expect(obj.a).toHaveBeenCalled();
```

Or maybe I'm misunderstanding your question.

adriangodong:
That’s for a function. How about for a property? If ‘a’ is a property, how do we differentiate between getting the property value vs. the spy?

aeisenberg:
That's a good question. I have never used spies on non-functions before. Perhaps I should restrict `spyOnAll` to functions only. (and perhaps rename to `spyOnAllFunctions`)

And then there could be an equivalent `spyOnAllProperties` to be implemented at a later date.

aeisenberg:
I'll update this PR so that it provides spyOnAllFunctions and then create
another one for spyOnAllProperties.

aeisenberg:
Updated the commit so that it provides spyOnAllFunctions. Do you want me to update the PR text as well? If and when this goes through, I'll add spyOnAllProperties.

steenburgh:
I don't know a lot about developing for Jasmine, but I noticed that I cannot use spyOnAllFunctions outside of a spyRegistry. Was that intended behavior? Currently, 
```
it("does a thing", function() {
    spyOnAllFunctions(someThingToSpyOn);
});
```
throws `env.spyOnAllFunctions` is not a function.

Adding the following to `Env.js` seems to fix my problem.
```
this.spyOnAllFunctions = {
  return spyRegistry.spyOnAllFunctions.apply(spyRegistry, arguments);
}
```

dougbacelar:
+1 !

Took me a long time to find this PR, excited for this. 

Unfortunately, I am not currently on the latest version, but very glad with the addition.

ShinDarth:
is there a way to use this with classes instead of objects?

aeisenberg:
For a class named A, I believe that calling spyOnAll(A.prototype) should
work the way you expect.

PR #1593, opened 2018-08-19
[WIP] Fix context cloning proccess

darthjee:
whenever a new context is generated (through the use of `UserContext.fromExisting`, if a property is also an Object, properties may leak between examples)

This PR tries to fix this issue

## Description
This PR will try to, anytime the context is being created from an existing one, once a property is an object, this will be also duplicated

### Problems
  an object might be an instance of a simple structure or might be something else with a different constructor which might pose a problem for clonning ... I am still thinking about the best way to solve it

## Motivation and Context
sometimes the object stored is being modified by specs and this is being shared between specs, when it should not

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/UserContextSpec.js
+++ b/spec/core/UserContextSpec.js
@@ -50,5 +50,85 @@ describe("UserContext", function() {
       });
     });
   });
-});
 
+  describe('context generation', function() {
+    beforeAll(function() {
+      this.id = 1;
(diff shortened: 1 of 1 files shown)


darthjee:
After checking something, I believe that the problem was actually the way the specs were written, as variables to the context should be set inside a beforeEach

PR #1599, opened 2018-09-06
Optimized clearTimeout cpu usage

Havunen:
Hi

I optimized CPU usage Jasmine uses when running tests. See screen shots for before after

**Before**
before

**After**
after

Its mostly about not calling `clearTimeout` when timeoutId is not number. This PR reduces `clearTimeout` cost by 50%, or 10% of total time. I believe rewriting `QueueRunner` to pure functional form could benefit also, but that is larger task to do.


piecioshka:
@Havunen  👏

Issue #1595, opened 2018-08-23
implement toBeRejectedWith(exception) method


dkrantsberg:
jasmine now has toBeRejected() method to check if promise rejects, however it doesn't check for specific exception being thrown, which limits its use.  
It would be very helpful  to have toBeRejectedWith(exception) method which would be used like this: 

```
await expectAsync(aPromise).toBeRejectedWith(expectedError);
```
or to be even more flexible: 
```
await expectAsync(aPromise).toBeRejectedWith({message: 'expected error message'});
await expectAsync(aPromise).toBeRejectedWith({code: 21});
```

PR #1600, opened 2018-09-09
WIP: implementation of toBeRejectedWith()

thodorisbais:
Hello,

It's been a while since I worked with this codebase, so, having seen #1595 , I thought it would be a good idea to take a better look at it.

I'm pretty sure there are mistakes, so, I'm just marking this as a WIP, hoping for your comments and help.

_So, the very first question is "is the implementation right? Am I missing something?"; after I have a clear view, more tests will be added of course. _

Regards,
TB

### DIFF
--- a/spec/core/AsyncExpectationSpec.js
+++ b/spec/core/AsyncExpectationSpec.js
@@ -178,40 +178,52 @@ describe('AsyncExpectation', function() {
       });
     });
 
-    it('builds its message correctly when negated', function() {
+  });
+
+  describe('#toBeRejectedWith', function() {
+    it('passes if the promise is resolved to the expected value', function() {
       jasmine.getEnv().requirePromises();
 
-      var addExpectationResult = jasmine.createSpy('addExpectationResult'),
-        expectation = jasmineUnderTest.AsyncExpectation.factory({
-          util: jasmineUnderTest.matchersUtil,
-          actual: Promise.resolve(true),
-          addExpectationResult: addExpectationResult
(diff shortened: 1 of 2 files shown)


thodorisbais:
Any news on this one?

spec/core/AsyncExpectationSpec.js
```diff
@@ -178,40 +178,52 @@ describe('AsyncExpectation', function() {
...
     });
 
-    it('builds its message correctly when negated', function() {
+  });
+
+  describe('#toBeRejectedWith', function() {
+    it('passes if the promise is resolved to the expected value', function() {
       jasmine.getEnv().requirePromises();
 
+      var actual = Promise.resolve({foo: 42});
```
slackersoft: I'm pretty sure this should be `Promise.reject` in order to make the matcher pass under the appropriate circumstances. Should probably also change the test description to match

spec/core/AsyncExpectationSpec.js
```diff
@@ -178,40 +178,52 @@ describe('AsyncExpectation', function() {
...
+          actual: actual
+        });
       });
     });
 
-    it('supports custom equality testers', function() {
+    it('fails if the promise is resolved to a different value', function() {
       jasmine.getEnv().requirePromises();
 
+      var actual = Promise.resolve({foo: 17});
```
slackersoft: If we're trying to check the value comparison here, this should also be a `Promise.reject`.

Issue #1594, opened 2018-08-23
Improve exception formatter


giniedp:
## Expected Behavior
If an expectation fails because of an ErrorEvent it should print at least the message of that ErrorEvent object.

## Current Behavior
The printed message is simply `[object ErrorEvent] thrown`

This is because the ErrorEvent object does not have a `name` property, so it drops into the else case in the ExceptionFormatter here https://github.com/jasmine/jasmine/blob/master/src/core/ExceptionFormatter.js#L8

## Possible Solution
The exception formatter could be changed, so that it can handle the ErrorEvent object which is described here https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

## Suite that reproduces the behavior (for bugs)
https://stackblitz.com/edit/jasmine-error-event

slackersoft:
This sounds like a nice addition. I would be happy to review a pull request to add support for printing the `message` for a thrown object that doesn't also have a `name` property.

Thanks for using Jasmine!

armand2099:
It will be great if the script line could be printed when an exception fails instead of:
Failure: Expected true to be false.
Currently I'm using:
expect(btn.enabled).toBe(false, "[TS:" + Error().lineNumber + "] Button should be disabled");

nitobuendia:
Sent PR #1601 to cover @giniedp proposal.

PR #1601, opened 2018-09-11
Print error message when available

nitobuendia:
## Description
With the current set up, error message is only printed out when error name is available.
The new set up ensures error message is always printed when available.

## Motivation and Context
There are situations where the ErrorEvent object does not have a `name` property, but the message is still relevant. For more details see #1594

## How Has This Been Tested?
To be done.

## Types of changes

## Checklist:
- [X] My change does _not_ require updates of documentation.

### DIFF
--- a/spec/core/ExceptionFormatterSpec.js
+++ b/spec/core/ExceptionFormatterSpec.js
@@ -40,8 +40,8 @@ describe("ExceptionFormatter", function() {
     it('formats thrown exceptions with message but no name', function() {
       var unnamedError = {message: 'This is an unnamed error message.'};
 
-      exceptionFormatter = new jasmineUnderTest.ExceptionFormatter(),
-      message = exceptionFormatter.message(unnamedError);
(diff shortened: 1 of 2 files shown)


nitobuendia:
Potential proposal for testing:

Extend 3x classes from Error:
- One NamedError where this.name = 'NamedError';
- One UnnamedError where this.name = null;
- One EmptyError where this.name = null; and this.message = null.

Raise each of them and check the "return' output:
- First one should contain NamedError and passed message.
- Second one should contain only message.
- Third one should be showing the stringified version of the object.


ETA: Thursday.

nitobuendia:
Added tests. For the three proposals:
- NamedError type check was significantly covered by other tests.
- Added tests for UnnamedError and EmptyError.
- The EmptyError test is similar to "formats thrown exceptions that aren't errors" but a slightly different flavour of it.

The PR is now ready for review and merge.

Issue #1573, opened 2018-07-02
jitSpy = spyOn(window, 'Function').and.callThrough(); breaks Jasmine


mhevery:
Imagine you would like to spy on `new Function()` calls. This may be accomplished in your tests by something like so:
```
const mySpy = spyOn(window, 'Function').and.callThrough();
```

This works great as expected except if one would like to test with clock. In that case it fails with NullPointerException. The crux of the problem is code which contains `Function.prototype.apply.apply:
- https://github.com/jasmine/jasmine/blob/74287c578cb7a9c5a775fba849f8e7c71f6c8194/src/core/Clock.js#L90-L96
- https://github.com/jasmine/jasmine/blob/8bb0e2d494eff37fa19cf1d50066b8cf225f603a/src/core/QueueRunner.js#L54-L60
- https://github.com/jasmine/jasmine/blob/9ee7b4ee0b60a734f6f4beb7087fc2e766a42d0d/src/core/ClearStack.js#L44-L46

With spy enabled on `Function` `Function.prototype.apply.apply` raises NPE because `Function.prototype.apply` is not defined. 

The issue is that `spyOn` replaces native `Function` with an instance of `Function` which no longer has the `prototype`.

## Suggested Fix.

In all cases where `Function.prototype.apply.apply` is invoked, replace it with.
```
// Place the next line at top of the file in the outer most context so that `Function.prototype`
// can be captured early.
const Function_prototype = Function.prototype;


function someFunction() {
  // Now even if `Function` gets replace in `spyOn` the code works fine 
  // Since we are working with the original Function.prototype
  Function_prototype.apply.apply(...);
}
```

## Simple workaround in tests
For users of `jasmine` which may run into this problem the following line fixes the issues in your test.
```
const mySpy = spyOn(ɵglobal, 'Function').and.callThrough();
// Patch prototype due to spying on the Function
Function.prototype.apply = Function.constructor.prototype.apply;
```

nitobuendia:
Clock.js also uses Array.prototype.

https://github.com/jasmine/jasmine/blob/74287c578cb7a9c5a775fba849f8e7c71f6c8194/src/core/Clock.js#L168-L168

Even if the `const Function_prototype = Function.prototype;` is applied, using `spyOn(window, 'Array');` would make jasmine fail again. Trying to capture all possible window Objects like this would be tedious and not a very good practice to be populating the global space or even namespace.

----

**Approach 1**

[...]

PR #1605, opened 2018-09-15
Import prototype methods from original object before spying

nitobuendia:
Potential fix for #1573.

## Description
Spied objects copy the prototype of their original objects.

## Motivation and Context
When spying on global objects such as Function or Array, in combination with methods which relies on these objects (e.g. jasmine.clock.install()) the tests fails or Jasmine throws errors.

## How Has This Been Tested?
[See comment](https://github.com/jasmine/jasmine/issues/1573#issuecomment-421585251) on #1573

## Types of changes

## Checklist:
- [x] Tests cannot be added as they need to rely on "window" object.

### DIFF
--- a/src/core/Spy.js
+++ b/src/core/Spy.js
@@ -1,4 +1,6 @@
 getJasmineRequireObj().Spy = function (j$) {
+  // Allows copying objects even when spyOn(window, 'Object'); happened.
+  var copyObject = Object.assign;
 
   var nextOrder = (function() {
     var order = 0;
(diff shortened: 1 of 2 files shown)


slackersoft:
Thanks for taking a look at this. I'm afraid there are a few issues with this PR as it stands right now:

* You haven't provided any tests for this functionality. Given the implementation settled upon, it should be reasonably easy to setup a constructor that you can spy on in tests to ensure that this functionality keeps working going forward.
* Since we're looking to fix an existing issue with expecting functions to be available on global objects, I would rather save off the `assign` function itself rather than the `Object` global
* `requireCore` is not the correct place to save off things from the global environment, that should probably happen in somewhere like `util` where we can provide a helper function, because
* Jasmine can't rely on `Object.assign` being available. It's great if it is available, but we need a fallback in case Jasmine is being run in Internet Explorer or some other supported environment that doesn't provide that.

Please also see [my comments](https://github.com/jasmine/jasmine/issues/1573#issuecomment-428398563) in #1573 about how approaches for this issue.

slackersoft:
Closing this since we haven't seen any updates since initial feedback

PR #1613, opened 2018-10-18
add promiseLike support for non global / browser Promise API implementations

codymikol:
## Description

This adds an inPromiseLike method that will test to see if a given object has promise like qualities. Then this helper replaces current isPromise implementations.

## Motivation and Context

This fixes: #1612

Some libraries have wrappers for the Promise API or roll their own implementation. Currently jasmine strictly requires the usage of the browser / nodejs global PromiseAPI or else it will throw an error when using assertAsync that you must pass a promise.

This solves this problem by adding a new method isPromiseLike that can determine if a method has promise like qualities. This way we can handle wrapped / custom Promise API implementations.

## How Has This Been Tested?

The isPromiseLike method has unit tests and unit tests have also been added for isPromise. Integration tests were already in place for the assertAsync functionality and with the new isPromiseLike method implemented they are all still passing.

I ran into this problem while trying out the assertAsync functionality for the first time. I noticed that I was always getting an error that assertAsync requires a promise. After making changes to jasmine, I ran the buildDistribution grunt task and replaced my version of jasmine-core with the new build I had created. Everything seemed to work perfectly with these changes. 

There was one gotcha that was specific to angularJS for anyone reading this in the future. For angularJS you have to wrap your returned promise in $rootScope.$apply() or else angularJS will not know to trigger a digest.

ex. 

```javascript
return $rootScope.$apply(expectAsync($q.resolve('Hello!').toBeResolvedTo('Hello!'));
```

## Types of changes

## Checklist:


src/core/matchers/matchersUtil.js
```diff
@@ -193,9 +193,9 @@ getJasmineRequireObj().matchersUtil = function(j$) {
       return false;
     }
 
-    var aIsPromise = j$.isPromise(a);
```
slackersoft: I don't think we want to move Jasmine's equality away from doing an explicit constructor check at this point. It is possible for non-Promises to have a `then` function available, and we don't want to break deep equality for anyone relying on that functionality.

spec/core/UtilSpec.js
```diff
@@ -31,6 +31,36 @@ describe("jasmineUnderTest.util", function() {
...
+
+    it("should return false for promise like objects", function () {
+      var promiseLikeObject = function () {this.then = function () {};};
+      var p = new promiseLikeObject();
+      expect(jasmineUnderTest.isPromise(p)).toBe(false);
+    });
+
+  });
+
+  describe("isPromiseLike", function () {
```
slackersoft: We should also have some negative tests for `isPromiseLike` as well. Something like checking for a `then` that isn't actually a function, and that `null` and `undefined` continue to work.

spec/core/UtilSpec.js
```diff
@@ -31,6 +31,38 @@ describe("jasmineUnderTest.util", function() {
     });
   });
 
+  describe("isPromise", function () {
+
+    it("should return true when passed a native promise", function () {
+      if (typeof Promise === 'undefined') { return; }
```
slackersoft: Let's use the `requirePromises` function supplied by Jasmine's promise helpers, for a bit more visibility on when this test is being skipped.

codymikol:
Thanks for the review, I'll try and make these changes tonight :chicken:

spec/core/UtilSpec.js
codymikol: Interestingly I added tests for isPromise as well and found a bug. Currently the functionality is

```javascript
    return typeof jasmineGlobal.Promise !== 'undefined' && obj && obj.constructor === jasmineGlobal.Promise;
```

this totally looks like it should work, but if obj is null or undefined, that value (null or undefined) will be returned.

Not super sure why this happens, but if you evaluate

```javascript
1 === 1 && null && 1 === 1
```

you will get null.

I'm changing this method to 

```javascript
return typeof jasmineGlobal.Promise !== 'undefined' && !!obj && obj.constructor === jasmineGlobal.Promise;
```

and now it works as expected.

codymikol:
I believe I have tackled everything in your review, let me know if you'd like any additional changes 🐔

gkalpak:
For anyone trying to use this with AngularJS, `return $rootScope.$apply(expectAsync($q.resolve('Hello!').toBeResolvedTo('Hello!'))` is NOT correct.
See https://github.com/angular/angular.js/issues/16744#issuecomment-433428714.

PR #1632, opened 2018-12-10
When catching a global error in Node.js, print the type of error

jbunton-atlassian:
## Description
`GlobalErrors.js` now adds a new `jasmineMessage` property to errors. This is picked up by `ExceptionFormatter.js` and used in preference to the regular `message` property in order to provide a more detailed message.

## Motivation and Context
Currently when Jasmine catches an unhandled promise rejection or an uncaught exception it fails the test with that `Error`. This is unclear, it is not possible to distinguish between the different errors below.

```
it('global uncaught exception', (done) => {
    setTimeout(() => { throw new Error('bad'); }, 0);
    setTimeout(done, 10);
});

it('regular uncaught exception', () => {
    throw new Error('bad');
});

it('global unhandled rejection', (done) => {
    setTimeout(() => Promise.reject(new Error('bad')), 0);
    setTimeout(done, 10);
});

it('regular unhandled rejection', () => {
    return Promise.reject(new Error('bad'));
});
```

## How Has This Been Tested?
I have run the full test suite in the following environments:
- Node.js 11
- Firefox 63
- Chrome 70

Additionally I ran the four specs above in Node.js and they produced this output. I have trimmed the stack traces as they are not relevant.
```
1) global uncaught exception
  Message:
    Uncaught exception: Error: bad
  Stack: <trimmed>

2) regular uncaught exception
  Message:
    Error: bad
  Stack: <trimmed>

3) global unhandled rejection
  Message:
    Unhandled promise rejection: Error: bad
  Stack: <trimmed>

4) regular unhandled rejection
  Message:
    Error: bad
  Stack: <trimmed>
```

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/GlobalErrorsSpec.js
+++ b/spec/core/GlobalErrorsSpec.js
@@ -98,9 +98,11 @@ describe("GlobalErrors", function() {
     errors.pushListener(handler);
 
     var addedListener = fakeGlobal.process.on.calls.argsFor(0)[1];
-    addedListener(new Error('bar'));
+    const expectedError = new Error('bar');
+    addedListener(expectedError);
 
-    expect(handler).toHaveBeenCalledWith(new Error('bar'));
+    expect(handler).toHaveBeenCalledWith(expectedError);
+    expect(expectedError.jasmineMessage).toBe('Uncaught exception: Error: bar');
 
     errors.uninstall();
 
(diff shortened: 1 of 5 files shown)


slackersoft:
Can you give me a bit more information on what the current functionality you're seeing is compared to the behavior you'd like to see? I'm not clear on exactly what the issue is.

Thanks.

jbunton-atlassian:
Hi,

The problem is that if I have a test failure with an error printed I cannot tell how Jasmine caught this error. Knowing how the error fired is crucial information. For example if I'm looking for an unhandled promise rejection I'll look for code missing a `.catch()`, or perhaps yielding the event loop before adding the catch handler.

To be concrete when I run the four specs from my first comment against Jasmine 3.3.0 I get this output. It's impossible to distinguish the global errors from the "regular" exceptional/rejection cases.

NB: It actually is possible to distinguish them, the global error cases are logged twice because of the additional call to `self.fail(err)` in `QueueRunner.js`. However this isn't very clear to the user :)

```
1) global uncaught exception
  Message:
    Error: bad
  Stack: <trimmed>
  Message:
    Failed: bad
  Stack: <trimmed>

2) regular uncaught exception
  Message:
    Error: bad
  Stack: <trimmed>

3) global unhandled rejection
  Message:
    Error: bad
  Stack: <trimmed>
  Message:
    Failed: bad
  Stack: <trimmed>

4) regular unhandled rejection
  Message:
    Error: bad
  Stack: <trimmed>
```

Issue #1634, opened 2018-12-12
Feature request: Include (custom) expectation message in both passed and failed expectations


GCHQDeveloper911:
I appreciate this is something of a duplicate of #1127, but the response there of "Jasmine doesn't expect you to really print messages for expectations that were met" doesn't really help.

I'm guessing that many users like me will have tests that produce multiple expectations. Some might argue this is not good practice - and they may have a point - but the fact is, that's what many people will do. Furthermore, sometimes these varied expectations can include somewhat "dynamically" generated actual / expected values - it would be massively beneficial to be able to see the actual values used in the successfully executed expectation. 

Beyond this, it would be useful to have a custom message for reporting purposes. A stream of "Passed. Passed. Passed. Passed." is meaningless to interested parties. Something that says "Passed - found 3 doughnuts left after having 4 before that, because we ate one" would be far more useful / meaningful.

Clearly a number of users have looked into what would be required to support this sort of thing. Can we revisit this as an idea? If someone submitted a PR that offered it, would the PR immediately be rejected?

slackersoft:
See [my comment](https://github.com/jasmine/jasmine/issues/971#issuecomment-156576842) in the #971 (referenced from #1127) about some of the difficulties related to this. The main issue is that a given matcher doesn't know whether the user has inserted `.not` into the call chain, and as such is unable to provide the message you suggest. There are also some memory usage concerns around passed expectations and how much memory they should be expected to take up from other members of the community, so I'm hesitant about adding more information back into these.

Hope this helps. Thanks for using Jasmine!

GCHQDeveloper911:
I've made a few simple changes to improve the reporting - I don't believe there will be any impact on memory usage - with your approval I'll submit a PR and let you take a look. I've not done anything earth shattering(!), but hopefully something that you and others would agree makes sense / could prove beneficial. FWIW, We're using webdriverio, and I'm going to be submitting a PR there for some similar changes, but there is (unless I'm very much mistaken!) no interdependency between my changes there, and here.

slackersoft:
If you want to report additional context around an expectation, you should also take a look at the new [`withContext`](https://jasmine.github.io/api/edge/matchers#withContext) matcher helper. Additionally, Jasmine doesn't report (and doesn't expect you to report) on expectations that were successful. This includes expectations that use `.not` in them.

[...]

PR #1645, opened 2019-01-10
Updated reporting to support specified messages in custom matchers an…

GCHQDeveloper911:
Updated reporting to support specified messages in custom matchers and output message with passed expectations (not just 'Passed') Note that custom matchers would need to be named along lines of 'toEqualBecause' or 'toBeGreaterThanBecause'.

## Description
Previously, passed expectations were only reported as 'Passed'. This gave no meaningful indication of what comparison had been done, leaving the user unsure as to what had passed, or when. Now the same sort of output as is seen on Failed expectations is shown. I have also added 'Because' support for custom comparisons with bespoke messages. For example: expect(1).toEqual(0,'well this is silly') Would produce 'Expected 1 to equal 0 because well this is silly' in the output.

## Motivation and Context
For tests with numerous expectations, this gives added the user scope to add clarity as to what happened in the test, both good and bad!

## How Has This Been Tested?
Using standard and custom toEqual (toEqualBecause), and toBeGreaterThan (toBeGreaterThanBecause) I created a test with a number of expectations, some designed to fail, some to pass, and also added negated versions. All seemed fine.

## Types of changes

## Checklist:
- [X] I have read the **CONTRIBUTING** document - not found?

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -3438,33 +3438,19 @@ getJasmineRequireObj().matchersUtil = function(j$) {
         actual = args[2],
         englishyPredicate = matcherName.replace(/[A-Z]/g, function(s) { return ' ' + s.toLowerCase(); }),
         customMatcher = matcherName.indexOf('Because') > 0 ? true : false;
-      var expected, customMsg;
+      var expected = args[3],
+        customMsg = args[4] || '';
         if (customMatcher){
-          expected = [];
-          expected.push(args[3]);
-          customMsg = args[4] || '';
-          englishyPredicate = englishyPredicate.replace(' because', '');
+          englishyPredicate = englishyPredicate.replace(' because', '')
           }
-        else{
-          expected = args.slice(3);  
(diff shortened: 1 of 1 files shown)


slackersoft:
See [my comment](https://github.com/jasmine/jasmine/issues/1634#issuecomment-453337068) in #1634 about the tradeoffs involved in a feature like this.

This pull request is also not ready for inclusion in Jasmine, and indicates that you didn't fully read the [CONTRIBUTING](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md) document.

* You have changed `jasmine.js` which is a generated file
* You haven't changed and of the `src` files that would generate your changes in the next version of Jasmine
* You haven't provided any tests for the new functionality

Even if these issues are fixed, I'm not sure this is a feature we're looking for in Jasmine itself. It is especially suspect to check for a trailing `Because` on a matcher now that Jasmine supports [withContext](https://jasmine.github.io/api/edge/matchers#withContext) as a mechanism to provide that extra information.

PR #1654, opened 2019-02-19
Fix statefulness bugs for test reexecution

SLaks:
## Description

 - Don't dispose top-level suite, since it's never recreated.
 - Allow beforeAndAfterFns() to be called multiple times.

## Motivation and Context

This lets me run jasmine.getEnv().execute() in /debug.html to rerun all tests without reloading the page.

NOTE: This means that spies creates in the top-level suite will never be torn down (to allow reruns with them).
That is bad practice; don't do that.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -916,30 +916,6 @@ describe("Env integration", function() {
     env.execute();
   });
 
-  it('removes a spy from the top suite after the run is complete', function(done) {
-    var env = new jasmineUnderTest.Env(),
-      originalFoo = function() {},
-      testObj = {
-        foo: originalFoo
-      };
-
-    env.beforeAll(function() {
-      env.spyOn(testObj, 'foo');
-    });
-
-    env.it('spec', function() {
-      expect(jasmineUnderTest.isSpy(testObj.foo)).toBe(true);
-    });
-
-    env.addReporter({
-      jasmineDone: function() {
-        expect(jasmineUnderTest.isSpy(testObj.foo)).toBe(false);
-        done();
-      }
-    });
-
-    env.execute();
-  });
 
   it("Mock clock can be installed and used in tests", function(done) {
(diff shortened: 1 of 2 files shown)


slackersoft:
Jasmine hasn't really ever intended to support re-executing an existing `Env` and the 3.0 release officially broke that. This is a memory optimization to remove all functions that Jasmine has completed execution on to free up closures. We have also put in work to make the top level suite work as similarly to any other suite in Jasmine, to eliminate a class of bug report where some things don't work when you do them this way. This isn't about which way is the "right" or "wrong" way to test your code, but making sure that the testing framework behaves consistently in different contexts.

Additionally, I don't think think that this would actually allow multiple executions of a given `Env` because all of the child suites and specs will have removed their own references to the functions that would actually execute the suite.

SLaks:
What is that a memory optimization for?
If you don't support re-executing tests, there is no point in keeping a Jasmine execution context around after tests have finished.

For anything but the top-level suite, reexecution must already be supported, because it must be possible to execute two specs from the same suite.

slackersoft:
> What is that a memory optimization for?

If you initialize a large object that is saved by reference in the closure, `delete`ing the function will allow the garbage collector to reclaim that space so your Jasmine suite doesn't end up consuming all available memory on your system.

> If you don't support re-executing tests, there is no point in keeping a Jasmine execution context around after tests have finished.

I'm not sure what "Jasmine execution context" you're referring to here. Jasmine currently installs itself globally and has a singleton reference to the `Env` that manages your suite, but cleans up all of those internals as the suite executes.

> For anything but the top-level suite, reexecution must already be supported, because it must be possible to execute two specs from the same suite.

Re-entry into the same suite is not the same thing of re-execution of the suite. Jasmine knows which specs and suites are children of a suite and will only delete things that it is sure have been fully executed.

slackersoft:
Closing this since we haven't seen any updates since initial feedback

PR #1657, opened 2019-02-23
add execution time to Spec and Suite

wood1986:
## Description

Add the execution time for each Spec and Suite

image

## Motivation and Context

People can easily identify which Spec/Suite takes ages to complete. Based on this information, people can tweak the test

## How Has This Been Tested?

It is just a proposal. If the maintainers agree, I will finish them

## Types of changes

## Checklist:

If the maintainers agree, I will update them

### DIFF
--- a/lib/jasmine-core/jasmine-html.js
+++ b/lib/jasmine-core/jasmine-html.js
@@ -308,7 +308,8 @@ jasmineRequire.HtmlReporter = function(j$) {
         if (resultNode.type === 'suite') {
           var suiteListNode = createDom('ul', {className: 'jasmine-suite', id: 'suite-' + resultNode.result.id},
             createDom('li', {className: 'jasmine-suite-detail jasmine-' + resultNode.result.status},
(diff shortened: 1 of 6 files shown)


slackersoft:
Thanks for taking some time too look at this. Unfortunately there are some issues with the PR as it stands right now.

1. You have not supplied any tests
1. You have modified the generated files `jasmine.js` and `jasmine-html.js`

The `result` returned from a Spec or Suite should be a static object and not have any functions on it. I would like to see the `duration` on the result without the start/stop times and have it just be a number in milliseconds.

I would also like to see the timing functionality moved out of `Env` (there's enough cruft in there already) and probably not have it be duplicated in both Spec and Suite. QueueRunner seems like a reasonable choice, but it may have too many responsibilities right now as well.

wood1986:
I do not think `QueueRunner` is the right place because it is being called from different class `Env`, `TreeProcessor`, `Spec`... The only place I can drop the timestamp for the duration is in `QueueRunner` is `setTimeout` and `clearTimeout`. Each method in `QueueRunner` is being called multiple times and I am unable to do it here

PR #1662, opened 2019-03-08
Fix #1651: add support for nodejs experimental-modules stacktraces

### LINKED ISSUE #1651: Support nodejs experimental-modules
## Are you creating an issue in the correct repository?

The "bug" (unwanted feature) is at jasmine.js#89.5

## Expected Behavior
We should be able to use the "ecmascript" experimental modules of nodejs (https://nodejs.org/dist/latest-v11.x/docs/api/esm.html), 

## Current Behavior
we receive an error:

> Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /mnt/c/Users/jhn/src/kiosk/tests/server/basic-test.mjs
>     at Object.Module._extensions..mjs (internal/modules/cjs/loader.js:724:11)
>     at Module.load (internal/modules/cjs/loader.js:599:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:530:3)
>     at Module.require (internal/modules/cjs/loader.js:637:17)
>     at require (internal/modules/cjs/helpers.js:22:18)
>     at /mnt/c/Users/jhn/src/kiosk/node_modules/jasmine/lib/jasmine.js:89:5
>     at Array.forEach (<anonymous>)
>     at Jasmine.loadSpecs (/mnt/c/Users/jhn/src/kiosk/node_modules/jasmine/lib/jasmine.js:88:18)
>     at Jasmine.execute (/mnt/c/Users/jhn/src/kiosk/node_modules/jasmine/lib/jasmine.js:257:8)

## Possible Solution
Jasmine should use "import" instead of "require" when loading esm modules. But that is not that easy, since "import" is asynchronous, while require is synchronous.

- Create a .jsm file (can be any test, even a simple one)
launch nearly as usual:
```
node --experimental-modules node_modules/.bin/jasmine
```

## Context
This is blocking to use jasmine in this new (experimental) context.

## Your Environment

* Version used: 3.3.1
* Environment name and version (e.g. Chrome 39, node.js 5.4): nodejs v10.15.1
* Operating System and version (desktop or mobile): ubuntu
* Link to your project:

jehon:
Add support for nodejs --experimental-modules.

## Description
When using node --experimental-modules, the stackstrace are not string anymore, they are an array of "CallSites".
> See https://nodejs.org/api/esm.html

## Motivation and Context
Jasmine was failing with something like [...] .split('\n') [...]

## How Has This Been Tested?
This is difficult to test with the current setup, because you need to restart node with --node-modules.
I did test that by running tests on my own project.

## Types of changes

## Checklist:
- [?] My code follows the code style of this project.
- [no] I have added tests to cover my changes.
- [yes] All new and existing tests passed.

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -6463,9 +6463,21 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
 
 getJasmineRequireObj().StackTrace = function(j$) {
   function StackTrace(error) {
-    var lines = error.stack
-      .split('\n')
-      .filter(function(line) { return line !== ''; });
+    var lines = [ 'unknown result!' ];
+    if (Array.isArray(error.stack)) {
+      /** 
+       * node --experimental-modules (v11.9 at least)
+       * esm stackstraces are are composed of an array of "CallSite"
+       * 
+       * TODO: See later if this behavior need to be adapted again
+       */
+      lines = error.stack.map(function(site) { return site.toString(); });
+    } else {
+      /* normal handling */
+      lines = error.stack
+        .split('\n')
+        .filter(function(line) { return line !== ''; });
+    }
 
     var extractResult = extractMessage(error.message, lines);
 
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -6463,21 +6463,9 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
 
 getJasmineRequireObj().StackTrace = function(j$) {
   function StackTrace(error) {
-    var lines = [ 'unknown result!' ];
-    if (Array.isArray(error.stack)) {
-      /** 
-       * node --experimental-modules (v11.9 at least)
-       * esm stackstraces are are composed of an array of "CallSite"
-       * 
-       * TODO: See later if this behavior need to be adapted again
-       */
-      lines = error.stack.map(function(site) { return site.toString(); });
-    } else {
-      /* normal handling */
-      lines = error.stack
-        .split('\n')
-        .filter(function(line) { return line !== ''; });
-    }
+    var lines = error.stack
+      .split('\n')
+      .filter(function(line) { return line !== ''; });
 
     var extractResult = extractMessage(error.message, lines);
 
--- a/src/core/StackTrace.js
+++ b/src/core/StackTrace.js
@@ -1,8 +1,21 @@
 getJasmineRequireObj().StackTrace = function(j$) {
   function StackTrace(error) {
-    var lines = error.stack
-      .split('\n')
-      .filter(function(line) { return line !== ''; });
+
+    var lines = [ 'unknown result!' ];
+    if (typeof(error.stack) == 'string') {
+      /* normal handling */
+      lines = error.stack
+        .split('\n')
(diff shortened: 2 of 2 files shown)


slackersoft:
Please take a look at the [CONTRIBUTING](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md) document. This pull request right now, doesn't include any tests for the new functionality, or any code changes in `src`, but only changes the generated `jasmine.js` file. As such any merge of this will be overwritten without any kind of alert by the next `grunt buildDistribution`. Please add the code to the appropriate source file and we will generate `jasmine.js` when the PR is merged.

Additionally, Jasmine has its own functions to detect "Arrayness" (`j$.isArray_`) that should be used for browser compatibility.

Thanks for taking a look at this. Hope this helps. Thanks for using Jasmine!

jehon:
Waw, a lot more complex than expected...
How could I test this feature, that requires node to run with "--experimental-modules" startup flag?

slackersoft:
Jasmine's test suite is mostly unit tests right now, so a test that calls `StackTrace` with a thing that looks like the new stack traces is probably for now. I would also be happy to see another entry in the travis build matrix that runs a suite with the flag set, but that will probably be more fiddly to get correct.

jehon:
Hello,

In fact, this is about not having regressions when using "experimental modules".

To see that in action:

`node --experimental-modules ./node_modules/.bin/grunt jshint execSpecsInNode`

should succeed.

This is what it does.

Do we need to add that into travis build?

slackersoft:
If we want to add an entry to the travis-ci matrix you should be able to add one with `NODE_OPTIONS="--experimental-modules"` and the appropriate node version. I'm not sure it will actually fail though, when I run `NODE_OPTIONS="--experimental-modules" npm test` on my machine with node v10.13.0 everything executes just fine (although the tests are green right now).

For the unit tests, remember that the thing being formatted doesn't have to actually _be_ `CallSite` objects just provide enough functionality that Jasmine's `StackTrace` can interact with them like they are.

Hope this helps.

jehon:
Hello

This work is too huge for me to take it now. If someone want to take over, please feel free to do so.

Jean

sgravrock:
@jehon Thanks for your contribution. Can you provide more information on how to reproduce the problem that this PR solves? I've been unable to coax a stack trace that isn't a string out of Node. Here's what I've done so far:

1. Patched `jasmine-npm` to load specs using `import()`.
2. Wrote a spec that throws an exception, and put it in a spec file with a name ending in `.mjs`.
3. Ran Jasmine under Node 10.16.3, 11.9.0, 12.18.3, 13.5.0 using `--experimental-modules`.

In all cases Jasmine is able to correctly format and filter the stack trace.

jehon:
Thanks for all your trials.

It has been a long time, and I didn't see the problem coming back since then. I do remember that, in the time I made that PR, exception handling was still under work in nodejs part. Perhaps it has been solved from they part.

I propose that we close this PR. If the problem arise back, someone could find it back and use it if necessary?

sgravrock:
> I do remember that, in the time I made that PR, exception handling was still under work in nodejs part. Perhaps it has been solved from they part.

That was my guess as well. I'll go ahead and close this. Thanks again.

PR #1678, opened 2019-04-12
Handling nodejs assert exceptions

apla:
## Description

Currently Jasmine ignores nodejs assertions, which already contains information about `operator` (`matcher` in jasmine terms), `actual` and `expected` fields.

## Motivation and Context

I have lot of code which already use constructs like`assert.equal(value, expected)` and most other testing tools (jest, mocha) already supports this.

## How Has This Been Tested?

First commit have failing test, second one fixing it by passing `actual` and `expected` fields. Also I want to implement mapping between nodejs `operator` and jasmine `matcher`, but unsure where to put it — ExpectationResult or ExceptionFormatter.

## Types of changes

## Checklist:


spec/core/ExpectationResultSpec.js
```diff
@@ -73,4 +73,30 @@ describe("buildExpectationResult", function() {
     var result = jasmineUnderTest.buildExpectationResult({actual: 'some-value'});
     expect(result.actual).toBe('some-value');
   });
+
+  it("handles nodejs assertions", function() {
+    var assert = require('assert');
```
slackersoft: These specs also need to run in browsers, so we need to first check that we're in node.js where `require` exists before running this test.

spec/core/ExpectationResultSpec.js
```diff
@@ -73,4 +73,30 @@ describe("buildExpectationResult", function() {
...
+    var expectedValue = false;
+    try { assert.equal(value, expectedValue) } catch(e) { error = e; }
+
+    expect(Object.keys(error)).toContain(
+      'code', 'actual', 'expected', 'operator'
+    );
+    expect(error.code).toEqual('ERR_ASSERTION');
+    expect(error.expected).toEqual(expectedValue);
+    expect(error.actual).toEqual(value);
+    expect(error.operator).toEqual('==');
```
slackersoft: This block looks like we're testing the `assert` functionality itself and not necessarily how Jasmine handles it. Is this necessary?

spec/core/ExpectationResultSpec.js
```diff
@@ -73,4 +73,30 @@ describe("buildExpectationResult", function() {
...
+
+    var result = jasmineUnderTest.buildExpectationResult({
+      passed: false,
+      matcherName: '',
+      expected: '',
+      actual: '',
+      error: error
+    });
+    expect(result.actual).toEqual(value);
```
slackersoft: I would want to see the above checks for the `error` fields here for the `result` to ensure we're copying them over properly.

spec/core/ExpectationResultSpec.js
apla: I'm not sure how to do it in Jasmine way. Do you have an example? Because I can check for `process`|`window`, `require` availability and so on.

slackersoft: Take a look at [matchersUtilSpec](https://github.com/jasmine/jasmine/blob/7fb53dcdfa2dcf83682361a3cf160717be8a0d93/spec/core/matchers/matchersUtilSpec.js#L178) for an example of the opposite. The right thing to check for here might just be the presence of `require` with a `typeof` check.

spec/core/ExpectationResultSpec.js
apla: Actually, we're checking for undocumented, but fairly stable internals of third party library. If something will ever change in nodejs assertion library, we will know.

slackersoft: I think I would still rather not test the runtime here, but this is fair safeguarding for undocumented behavior.

We definitely need to check that Jasmine is putting values in the correct place on the `result` below.

apla:
Added requested checks

slackersoft:
Merged, but github can't find it because of squashing...

PR #1686, opened 2019-04-25
Add toBeRejectedWithError matcher

### LINKED ISSUE #1625: Add a `toBeRejectedWithError` async matcher
## Expected Behavior

I want to test to use the `toThrowError` utility matcher on an asynchronous operation which rejects with a given error.

## Current Behavior

API is not available - but [`eq`](https://github.com/jasmine/jasmine/blob/master/src/core/matchers/matchersUtil.js#L111-L117) and `objectContaining` works as interim solutions.

```
await expectAsync(boom()).toBeRejectedWith(new TypeError("BOOM"))
await expectAsync(boom()).toBeRejectedWith(jasmine.objectContaining({ message: "BOOM" }))
```

## Possible Solution

Jasmine already has `toThrowError` besides `toThrow` to make it easier matching against errors with various overloads.

```
expect(foo).toThrowError("foo bar baz");
expect(foo).toThrowError(/bar/);
expect(foo).toThrowError(TypeError);
expect(foo).toThrowError(TypeError, "foo bar baz");
```

Having `toBeRejectedWithError` besides `toBeRejectedWith` with the same overloads would be quite intuitive.

It seems that would require extracting some code from `toThrowError` first.

I wonder whether that extracted code could live in a `jasmine.error` asymmetric matcher instead though (meaning that ultimately `toThrowError` could be deprecated).

megahertz:
## Description
Allow to use async match similar to `toThrowError`, like

`await expectAsync(aPromise).toBeRejectedWithError(MyCustomError, /Error message/);`

## Motivation and Context
Fixes #1625 

## How Has This Been Tested?
specs are included, tested in node, ie8

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/matchers/async/toBeRejectedWithErrorSpec.js
+++ b/spec/core/matchers/async/toBeRejectedWithErrorSpec.js
@@ -1,0 +1,100 @@
+describe('#toBeRejectedWithError', function () {
+  it('passes when Error type matches', function () {
+    jasmine.getEnv().requirePromises();
+
+    var matcher = jasmineUnderTest.asyncMatchers.toBeRejectedWithError(jasmineUnderTest.matchersUtil),
+      actual = Promise.reject(new TypeError('foo'));
+
+    return matcher.compare(actual, TypeError).then(function (result) {
+      expect(result).toEqual(jasmine.objectContaining({ pass: true }));
+    });
+  });
+
+  it('passes when Error type and message matches', function () {
+    jasmine.getEnv().requirePromises();
+
+    var matcher = jasmineUnderTest.asyncMatchers.toBeRejectedWithError(jasmineUnderTest.matchersUtil),
+      actual = Promise.reject(new TypeError('foo'));
+
+    return matcher.compare(actual, TypeError, 'foo').then(function (result) {
+      expect(result).toEqual(jasmine.objectContaining({ pass: true }));
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


slackersoft:
This looks pretty good; all of the cases look correct, but it looks like some of the messages used by `.not` aren't quite right. For example:

```javascript
it('reject Error', async function() {
  await expectAsync(Promise.reject(new Error(''))).not.toBeRejectedWithError();
});
```
Produces the message: `"Expected a promise to be rejected with Error: undefined."` but should say: `"Expected a promise not to be rejected with Error, but it was."`. The general rule is when `pass` is `true`, write the failure message you'd want displayed if you used `.not` in the expectation.

A smaller change: I'd like to see the message tell the user more about what the actual rejection was. Right now:

```javascript
it('reject', async function() {
  await expectAsync(Promise.reject()).toBeRejectedWithError();
});
```
Produces: `"Expected a promise to be rejected with Error: undefined but it was not rejected with Error."`. I'd like to see something more like the comparable `toThrowError` version: `"Expected function to throw an Error, but it threw undefined."`

I tend to prefer error messages that tell the use what _did_ happen as they usually make it easier to figure out what went wrong.

megahertz:
Thank you for the feedback. I'll try to make this changes on this week.

megahertz:
@slackersoft Looks better now?

Issue #1648, opened 2019-02-07
spyOn() breaks Jasmine with when returning rejected Promise


pe8ter:
When using this pattern in a spec

```javascript
spyOn(foo, 'bar').and.returnValue(Promise.reject('NO!'));
```

the entire test run fails even if the spied function never gets called.

This is similar to https://github.com/jasmine/jasmine/issues/1590, but different enough that I thought it warranted a separate item (in my case my spy creates correctly). Feel free to close this if you think it is too similar.

## Expected Behavior
The test run should pass because both the spec and tested code do not actually throw any exceptions or await on rejected Promises.

## Current Behavior
The spec succeeds but the test run fails (the Node.js process exits with code 1). When I use jasmine-console-reporter module it reports `An error was thrown in an afterAll`.

## Suite that reproduces the behavior

Here is the complete test file. Notice that the spied function is in a branch that cannot get called: 

```javascript
const foo = {
    bar: async () => 0
}

const baz = {
    quux: async () => {
        if (false) {
            await foo.bar();
        }
    }
};

describe('some suite', () => {
    it('some spec', async () => {
        const spy = spyOn(foo, 'bar').and.returnValue(Promise.reject('NO!'));
        await baz.quux();
        expect(spy).not.toHaveBeenCalled();
    });
});
```

## Context
We have a test where we need to verify that one code path isn't taken given some conditions. To make the test fail in a controlled manner, we added a spy with this rejected Promise pattern if the tests hits this undesired code path.

## Your Environment

* jasmine: **v3.3.1**
* jasmine-core: **v3.3.0**
* Environment: **Node.js v8.15.0**
* OS: **macOS Mojave, Version 10.14.3 18D42**

sgravrock:
Simply creating a rejected promise is enough to trigger Node's unhandled rejection event if the rejection is never handled, even if you don't do anything with the promise. Jasmine then turns the unhandled rejection into a failure, because unhandled rejections *almost* always indicate that something went wrong with the test or the code under test. You can see that more clearly with this stripped-down test, which produces the same failure:

```
it('fails', () => {
    Promise.reject('NO!');
});
```

If I understand your use case correctly, you don't want the unhandled promise to cause a failure unless `foo.bar` is called. There are a couple of ways to do that. I think the cleanest is to create the promise only if the function is called:

```
it('some spec', async () => {
    const spy = spyOn(foo, 'bar').and.callFake(() => Promise.reject('NO!'));
    await baz.quux();
[...]

PR #1688, opened 2019-05-01
Add resolveValue and rejectValue spy strategies

elliot-nelson:
## Description
This is a feedback-only PR (no specs yet) that adds some additional spy strategies to core.

### `.and.resolveValue`

Return a resolved promise for the specified value.

### `.and.rejectValue`

Return a rejected promise for the specified value.

### Promise: end user experience

I've been looking through my various jasmine helper files and I think for least surprise as an end user, either of these two approaches would be "what I expect":

```
// in a helpers.js file, etc.:
jasmine.setPromiseLibrary(require('bluebird'));

// or:
jasmine.getEnv().configure({
    promiseLibrary: require('bluebird')
});
```

Could it be more like `addMatchers`, where it has to happen in a beforeEach?  Sure, it definitely could, although I didn't go that route here (see next section).

### Promise: internal experience

I saw your (Greg's) notes on my other PR.  This is my attempt at a compromise :).

I'm not sure it's worth option-plumbing `Env > SpyFactory > Spy` just for providing a pointer to a useable Promise.  I see a Promise as being more of a globally available utility; more like `j$.isPromiseLike` than the options passed to the Expectation factory.

Based on that reasoning, this implementation:

- Internal usage of promises is intended to be through `j$.getPromise()`.  This is where we pick up global promise if it exists.
- End users can _configure_ a custom promise lib through Env, so, Env also provides a `getPromise()` (this only returns a lib if one was configured).
- Promise-checking (i.e., determining whether a promise lib is available) is the responsibility of the code using it.  In this case, the individual spy strategies check and generate custom messages if it isn't available.

### Alternative: option-plumbing

If you're convinced plumbing for Env > SpyFactory > Spy is the right approach, then it's about equal work to go the "global Promise impl" route like I have, or to make it a runnable per-spec setting instead.  Per-spec feels a little odd to me, but, it perhaps does make sense for large repos that are testing multiple components (perhaps client & server components that use different promise libraries).

## Motivation and Context

The spy strategies above are the ones I find myself missing the most, and now that custom spy strategies are a thing individual projects can solve the problems one-off -- but putting these strategies that almost any project could use in core makes tests more portable and gives an "accepted" syntax.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/EnvSpec.js
+++ b/spec/core/EnvSpec.js
@@ -46,6 +46,23 @@ describe("Env", function() {
     }));
   });
 
+  it('can configure a custom promise library', function() {
+    var myLibrary = { resolve: jasmine.createSpy(), reject: jasmine.createSpy() };
+    expect(env.getPromise()).toBeUndefined();
+
+    env.configure({ promiseLibrary: myLibrary });
+    expect(env.getPromise()).toBe(myLibrary);
+  });
+
+  it('fails to configure a custom promise library if library is invalid', function() {
+    var myLibrary = {};
+    expect(env.getPromise()).toBeUndefined();
+
+    expect(function() {
+      env.configure({ promiseLibrary: myLibrary });
+    }).toThrowError('Custom promise library missing `resolve`/`reject` functions');
+  });
+
   it('defaults to multiple failures for specs', function() {
     spyOn(jasmineUnderTest, 'Spec');
       env.it('bar', function() {});
--- a/spec/core/EnvSpec.js
+++ b/spec/core/EnvSpec.js
@@ -46,21 +46,19 @@ describe("Env", function() {
     }));
   });
 
-  it('can configure a custom promise library', function() {
-    var myLibrary = { resolve: jasmine.createSpy(), reject: jasmine.createSpy() };
-    expect(env.getPromise()).toBeUndefined();
+  describe('promise library', function() {
+    it('can be configured with a custom library', function() {
+      var myLibrary = { resolve: jasmine.createSpy(), reject: jasmine.createSpy() };
+      env.configure({ Promise: myLibrary });
+    });
 
-    env.configure({ promiseLibrary: myLibrary });
-    expect(env.getPromise()).toBe(myLibrary);
-  });
+    it('cannot be configured with an invalid promise library', function() {
+      var myLibrary = {};
 
-  it('fails to configure a custom promise library if library is invalid', function() {
-    var myLibrary = {};
-    expect(env.getPromise()).toBeUndefined();
-
-    expect(function() {
-      env.configure({ promiseLibrary: myLibrary });
-    }).toThrowError('Custom promise library missing `resolve`/`reject` functions');
+      expect(function() {
+        env.configure({ Promise: myLibrary });
+      }).toThrowError('Custom promise library missing `resolve`/`reject` functions');
+    });
   });
 
   it('defaults to multiple failures for specs', function() {
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -115,7 +115,8 @@ describe("SpyStrategy", function() {
       jasmine.getEnv().requirePromises();
 
       var originalFn = jasmine.createSpy("original"),
-          spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn});
+          getPromise = function() { return Promise; },
+          spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn, getPromise: getPromise});
 
       spyStrategy.resolveValue(37);
       spyStrategy.exec().then(function (returnValue) {
@@ -128,11 +129,9 @@ describe("SpyStrategy", function() {
       var originalFn = jasmine.createSpy("original"),
           spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn});
 
-      spyOn(jasmineUnderTest, 'getPromise');
-
       expect(function() {
         spyStrategy.resolveValue(37);
-      }).toThrowError('resolveValue requires global Promise, or a `promiseLibrary` configured with `jasmine.getEnv().configure()`');
+      }).toThrowError('resolveValue requires global Promise, or `Promise` configured with `jasmine.getEnv().configure()`');
     });
   });
 
@@ -141,7 +140,8 @@ describe("SpyStrategy", function() {
       jasmine.getEnv().requirePromises();
 
       var originalFn = jasmine.createSpy("original"),
-          spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn});
+          getPromise = function() { return Promise; },
+          spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn, getPromise: getPromise});
 
       spyStrategy.rejectValue(new Error('oops'));
       spyStrategy.exec().then(done.fail).catch(function (error) {
@@ -154,11 +154,9 @@ describe("SpyStrategy", function() {
       var originalFn = jasmine.createSpy("original"),
           spyStrategy = new jasmineUnderTest.SpyStrategy({fn: originalFn});
 
-      spyOn(jasmineUnderTest, 'getPromise');
-
       expect(function() {
         spyStrategy.rejectValue(new Error('oops'));
-      }).toThrowError('rejectValue requires global Promise, or a `promiseLibrary` configured with `jasmine.getEnv().configure()`');
+      }).toThrowError('rejectValue requires global Promise, or `Promise` configured with `jasmine.getEnv().configure()`');
     });
   });
 
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -150,6 +150,20 @@ describe("SpyStrategy", function() {
       }).catch(done.fail);
     });
 
+    it("allows a non-Error to be rejected, wrapping it in an exception when executed", function(done) {
+      jasmine.getEnv().requirePromises();
+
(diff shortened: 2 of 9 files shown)


sgravrock:
`and.rejectValue` might help fewer people run into problems like #1648.

slackersoft:
I'm not sure what the use case for multiple fake implementations is. At the point where you're telling Jasmine which code to run, you can already do different things on different invocations.

I can definitely see the need for `and.resolveValue` or `and.rejectValue`, and possibly the multiple call variants as well. The Promise stuff gets tricky due to Jasmine's current browser support matrix. We do actually still support IE 10 & 11, which don't even have built-in Promise, so Jasmine would need to provide an intelligible message when the Promise-based strategies are used but there is _no_ default implementation for Promise available, as you have here.

I think it makes more sense to have this be an explicit configuration call or setting the in `config()` function. `jasmine.getGlobal()` is really intended for internal-to-Jasmine usage to find whatever the global object happens to be (`window` in a browser, `global` in node.js, etc.).

elliot-nelson:
I'll pitch the multiple fakes: testing retry functions.  Seems to come up frequently for me for some reason.

```
request.get.and.callFakes(
    () => Promise.reject(new Error("404")),
    () => Promise.reject(new Error("404")),
    () => Promise.resolve("OK")
);

expect(await myModule.getWithRetry()).toEqual("404");
```

Yes, you can implement mocks that check `calls.count` directly, but it gets tiresome for complicated tests and is not as readable as above (imo).  And you can't use my suggested Promise-based strategies because you need a mix of both resolves and rejects.

slackersoft:
Ok, that use case makes sense. I might actually also look at baking some better knowledge of multiple calls into the higher level spy strategy such that you could do something like:

```javascript
request.get.and.rejectValue(new Error('404'))
    .andThen.rejectValue(new Error('404'))
    .andThen.resolveValue('OK')
```

or:

```javascript
request.get.and.do(
  (call) => call.rejectValue(new Error('404')),
  (call) => call.rejectValue(new Error('404')),
  (call) => call.resolveValue('OK')
)
```

The first is probably going to require more work to the SpyStrategy to make it work, but the interface feels a little cleaner to me. The latter would "_just_" be a new strategy that changes the strategy after each spy call. I like either of these as longer term solutions as they allow all the other spy strategies to get the multiple variant form without extra effort.

I also understand if you don't want to take this on at the moment and this might be something I look for in Jasmine 4.0.

elliot-nelson:
Actually, I like that a lot, and may propose a solution in a different PR.  (I think this PR is basically going to become "let's get some promise-based strategies", which is a separate thing.)

elliot-nelson:
I'm still working through the specifics, but I think the implementation really depends on what behavior you want.  Especially with regards to the current spy interface which _allows_ stuff like:

    if (spy.and.plan !== myMock) { spy.and.plan = myMock; }

I would want to be careful not to break that not-documented-but-OK behavior.  I think the correct solution is to change SpyStrategy's core guts to be a `planned` queue.  Then `andThen` can return a proxy that ensures new plans are added to the queue, whereas the default SpyStrategy behavior is to _set_ the queue to `[plan]`.

(We can turn `plan` into a property, to ensure that the not-documented behavior I mentioned earlier still holds true -- for example, directly assigning `plan` should be equivalent to setting the plan qeueue to `[plan]`.)

slackersoft:
I think I would actually be more inclined to _remove_ the ability to set `plan` directly (leaving the getter). Reaching into the spy strategy to set one of its properties shouldn't really be necessary now that you can provide custom spy strategies. Note that we shouldn't remove the setter immediately per your concern but probably use `Object.defineProperty` to mark it as deprecated if anyone is still using it.

elliot-nelson:
Updates:
- Updated body + title.
- Trimmed PR to only 2 new promise-based strategies, with the necessary promise handling.
- Added unit tests.

src/core/Env.js
```diff
@@ -193,6 +211,10 @@ getJasmineRequireObj().Env = function(j$) {
       }
     };
 
+    this.getPromise = function() {
+      return config.promiseLibrary;
+    };
+
```
slackersoft: This implementation of `getPromise` will only return the _configured_ library, and will never try to load up a default one already present in the execution environment (i.e. `window.Promise` or `global.Promise`).

src/core/base.js
```diff
@@ -45,6 +45,10 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return env;
   };
 
+  j$.getPromise = function() {
+    return j$.getEnv().getPromise() || j$.getGlobal().Promise;
+  };
+
```
slackersoft: I don't think that the configured promise library needs to be (or should be) exposed back outside of Jasmine itself. If Jasmine has been told to use whichever library I specify, I shouldn't need to check up on it to make sure.

spec/core/SpyStrategySpec.js
```diff
@@ -110,6 +110,62 @@ describe("SpyStrategy", function() {
     })
   });
 
+  describe("#resolveValue", function() {
+    it("allows a resolved promise to be returned", function(done) {
+      if (!jasmineUnderTest.getPromise()) {
+        pending('Environment does not support promises.');
```
slackersoft: Try using [`requirePromises`](https://github.com/jasmine/jasmine/blob/110c092c9ece00cd6c882bd573b911c35ff73478/spec/helpers/promises.js) for consistent behavior with other specs.

spec/core/EnvSpec.js
```diff
@@ -46,6 +46,23 @@ describe("Env", function() {
...
+  it('fails to configure a custom promise library if library is invalid', function() {
+    var myLibrary = {};
+    expect(env.getPromise()).toBeUndefined();
+
+    expect(function() {
+      env.configure({ promiseLibrary: myLibrary });
+    }).toThrowError('Custom promise library missing `resolve`/`reject` functions');
+  });
+
   it('defaults to multiple failures for specs', function() {
```
slackersoft: I think I might rather see these as integration specs that ensure the correct custom functions are called instead of checking that the configured Promise library is accessible. This also allows `getPromise` to not need to be publicly available.

spec/core/SpyStrategySpec.js
```diff
@@ -110,6 +110,62 @@ describe("SpyStrategy", function() {
...
+      expect(function() {
+        spyStrategy.resolveValue(37);
+      }).toThrowError('resolveValue is unavailable because the environment does not support promises.');
+    });
+  });
+
+  describe("#rejectValue", function() {
+    it("allows a rejected promise to be returned", function(done) {
+      if (!jasmineUnderTest.getPromise()) {
+        pending('Environment does not support promises.');
```
slackersoft: See note above about `requirePromises`.

src/core/SpyStrategy.js
```diff
@@ -84,6 +84,44 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
...
+   * Tell the spy to return a promise resolving to the specified value when invoked.
+   * @name SpyStrategy#resolveValue
+   * @function
+   * @param {*} value The value to return.
+   */
+  SpyStrategy.prototype.resolveValue = function(value) {
+    var Promise = j$.getPromise();
+
+    if (!Promise) {
+      throw new Error('resolveValue is unavailable because the environment does not support promises.');
```
slackersoft: Let's add a note in here for how to configure a custom library, as long as it doesn't get too wordy.

src/core/SpyStrategy.js
```diff
@@ -84,6 +84,44 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
...
+   * Tell the spy to return a promise rejecting with the specified value when invoked.
+   * @name SpyStrategy#rejectValue
+   * @function
+   * @param {*} value The value to return.
+   */
+  SpyStrategy.prototype.rejectValue = function(value) {
+    var Promise = j$.getPromise();
+
+    if (!Promise) {
+      throw new Error('rejectValue is unavailable because the environment does not support promises.');
```
slackersoft: Let's add a note in here for how to configure a custom library, as long as it doesn't get too wordy.

elliot-nelson:
Thanks for all your assistance.  I've made your suggested changes to the specs & "no promise" failure messages (open to additional thoughts on the error message wording).

I'm confused by your other comments though and want to check my assumptions:

I don't think it's even possible (right now) to allow the user to configure `promiseLibrary` through `jasmine.getEnv()`, and have that value be accessible to `SpyStrategy`, without exposing it _somewhere_ to the end user.  Either in `jasmine.getEnv()` or `jasmine`.  So I'm not sure how to follow through on those comments.

(The only examples I see in the code currently are where Env has local vars and passes them directly down through constructors.  This would accomplish your goal but would require plumbing through SpyFactory etc.  I'm willing to do that if we want it totally invisible, I just want to make sure I'm understanding properly before I tackle that...)

slackersoft:
For hiding the Promise library, I was thinking of something like wiring the library through to the SpyFactory and on to the particular strategies involved. I'm open to some discussion on this and I promise I'm not trying to make your life harder. The goal is to not expand Jasmine's public interface unless we really need to. Obviously the expansion is necessary to allow the library to be _set_, but I would like to explore keeping it private after that.

slackersoft:
I'm definitely not wedded to the `resolveValue`/`rejectValue` naming. I would tend to stay away from just `resolve`/`reject` just to ease on naming overloads. `resolveWith`/`rejectWith` sound great.

elliot-nelson:
Updates:
  - `resolveValue` -> `resolveWith`
  - `rejectValue` -> `rejectWith`

(I'm liking these a lot after writing some sample tests; "spy on method and resolve with value" & "spy on method and reject with error" are imo a good match for the rest of jasmine's semi-conversational API.)

PR #1689, opened 2019-05-02
Add support for ArrayBuffers to matcherUtil.equals.

### LINKED ISSUE #1687: Arraybuffer does not work correctly with toEqual
## Expected Behavior
Two arraybuffers are compared with their content in jasmine

## Current Behavior
Two different ArrayBuffers with different values but same length compare equal in jasmine.

Example passes even though it should not:
```javascript
describe('Arraybuffer', () => {
  it('equal different Arraybber', () => {
    const buffer1 = new ArrayBuffer(4);
    const buffer2 = new ArrayBuffer(4);
    const array = new Uint8Array(buffer1);
    array[0] = 1;
    expect(buffer1).toEqual(buffer2);
  });
});
```

dankurka:
Fixes #1687

## Description
Added support for using ArrayBuffers in matcherUtils.eq.

## Motivation and Context
ArrayBuffers with same length but different content were comparing as equal when they should not.

## How Has This Been Tested?
Ran tests locally.

## Types of changes

## Checklist:

One issue I am seeing with the change is that I am now referring to Uint8Array (after a guard on ArrayBuffer) but that makes jshint unhappy. I'd like some guidance on how to proceed with that.

### DIFF
--- a/spec/core/matchers/matchersUtilSpec.js
+++ b/spec/core/matchers/matchersUtilSpec.js
@@ -164,7 +164,7 @@ describe("matchersUtil", function() {
 
       expect(jasmineUnderTest.matchersUtil.equals(a,b)).toBe(true);
     });
-    
+
     it("passes for equivalent Promises (GitHub issue #1314)", function() {
       if (typeof Promise === 'undefined') { return; }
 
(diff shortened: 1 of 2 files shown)


src/core/matchers/matchersUtil.js
```diff
@@ -167,6 +168,36 @@ getJasmineRequireObj().matchersUtil = function(j$) {
         }
         return result;
       // RegExps are compared by their source patterns and flags.
```
slackersoft: The regexp comment should stay with the relevant `case` statement.

src/core/matchers/matchersUtil.js
```diff
@@ -167,6 +168,36 @@ getJasmineRequireObj().matchersUtil = function(j$) {
...
+      case '[object ArrayBuffer]':
+        // If we have an instance of ArrayBuffer the Uint8Array ctor
+        // will be defined as well
+        var arrayA = new Uint8Array(a);
+        var arrayB = new Uint8Array(b);
+        var arrayALength = arrayA.length;
```
slackersoft: I'd like to see the common code paths here pulled out so the existing `Array` code and this code stay in sync.

spec/core/matchers/matchersUtilSpec.js
```diff
@@ -524,6 +524,20 @@ describe("matchersUtil", function() {
       expect(jasmineUnderTest.matchersUtil.equals(mapA, mapB)).toBe(false);
     });
 
+    it("passes for ArrayBuffers with same length and content", function() {
```
slackersoft: These specs should do something similar to the existing Typed Array specs in pretty printer so they don't break in an environment where `ArrayBuffer` isn't defined.

perezd:
Anything preventing this from getting merged in?

sgravrock:
Merged via #1891. Thank you.

PR #1690, opened 2019-05-02
Provide access to Promise via jasmine.getEnv().Promise

elliot-nelson:
## Description
Provide basic access to "the current Promise library" via `jasmine.getEnv().Promise`.  This is a breakout/preparation for https://github.com/jasmine/jasmine/pull/1688, which would make use of a Promise library if it exists.

Here's a quick summary of what we want:

```
// In a browser with no Promise support

expect(jasmine.getEnv().Promise).toBe(undefined);

// In a browser or node env with Promise support

expect(jasmine.getEnv().Promise).toBe(Promise);

// In any environment, where user provides a custom Promise implementation

jasmine.getEnv().configure({ promise: Bluebird });
expect(jasmine.getEnv().Promise).toBe(Bluebird);

// Finally, if a different library overrides the global Promise and we haven't set up
// a custom one, Jasmine should respect that.

expect(jasmine.getEnv().Promise).toBe(undefined);
window.Promise = Bluebird;
expect(jasmine.getEnv().Promise).toBe(Bluebird);
```

## Motivation and Context
This PR sets up the underlying "how do we get a Promise" implementation for other Jasmine functions, and ensures there's a consistent way to use one.

## How Has This Been Tested?
- Currently, unit tested in node only
- Can perform additional manual testing if reviewers are comfortable with the approach

## Types of changes

## Checklist:

> I think the ability to customize the Promise library probably warrants a doc update, although I'm not sure the ability to `configure({ })` is currently documented at all.  Open to any suggestions here.

### DIFF
--- a/spec/core/EnvSpec.js
+++ b/spec/core/EnvSpec.js
@@ -233,4 +233,55 @@ describe("Env", function() {
       expect(globalErrors.install).toHaveBeenCalled();
     });
   });
+
+  describe("#Promise", function() {
+    it("defaults to global Promise, if available", function() {
+      var examplePromise = jasmine.createSpyObj('Promise', ['resolve', 'reject']);
+      var exampleGlobal = { Promise: examplePromise };
+      var exampleEnv = new jasmineUnderTest.Env({ global: exampleGlobal });
+      expect(exampleEnv.Promise).toBe(examplePromise);
+    });
+
+    it("defaults to undefined if no global Promise is available", function() {
+      var exampleGlobal = {};
+      var exampleEnv = new jasmineUnderTest.Env({ global: exampleGlobal });
+      expect(exampleEnv.Promise).toBeUndefined();
+    });
+
+    it("defaults to global Promise, if available", function() {
+      var examplePromise = jasmine.createSpyObj('Promise', ['resolve', 'reject']);
+      var exampleGlobal = { Promise: examplePromise };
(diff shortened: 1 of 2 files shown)


slackersoft:
I would rather see something like adding a configurable Promise library to Jasmine be part of actually using it in the rest of the code. I think it makes more sense and ensures the interface works with the code that actually needs it. So please pull this type of work into #1688. This also allow the two PRs to not have dependencies on each other.

I'm not sure that a raw `Env.Promise` is quite the right way to share this state throughout Jasmine. Currently `getEnv()` is for use _outside_ of Jasmine and not currently used within the library itself. I'm also wary of expanding the public interface for something that is intended for internal use. Most of the rest of the runtime defined things that are shared around the whole suite are also intended to be spec/suite specific, so I don't think we want to follow the `runnableResources` pattern. But this might be something we can inject into the `Expectation.factory` similar to how `matchersUtil` is.

elliot-nelson:
@slackersoft Thanks for your reviews and comments, very useful!

Takeaways:
 - I'll close this PR, and the functionality will show up in #1688 in the near future.
 - I'm not 100% clear on what interface (both internal and external) you'd prefer -- but we can figure that out later.  Mostly I want to make sure there is a clear, easy way for the jasmine _user_ to ask Jasmine to only use Bluebird when it creates promises (or `when`, or whatever your library of choice is).

PR #1691, opened 2019-05-04
Introduce .andThen spy strategy series

elliot-nelson:
## Description
Propose an implementation of `.andThen`, which is a new spy strategy wrapper that allows the user to specify a series of different strategies.

```
mySpy.and.throwError('No dice')
    .andThen.returnValue('OK')
    .andThen.stub();

mySpy();    // => throws error
mySpy();    // => returns 'OK'
mySpy();    // => returns undefined
```

### Changes to existing public interfaces

- Anywhere the user can say `.and`, they can now say `.andThen` instead, to enqueue "future" plans instead of assigning the current plan.

- Interacting directly with `spy.and.plan` behaves as expected, although assigning to it now issues a deprecation warning.

### New public interfaces

Ideally I'd like there to be no new surface area at all, but that's tough in Javascript :).

- New publicly accessible keys:
  - `spy.and._applyPlan`
  - `spy.and._planned`
  - `spy.and.plannedProxy`
  - `spy.and.deprecated`

Keys `_applyPlan` and `_planned` are explicitly implementation details (shame on anyone who relies on those in a test).  I didn't underscore `plannedProxy` or `deprecated` currently, but probably could if we wanted to be real sticklers about tell peopling "do not rely on the existence of these".

(The end user can get ahold of `deprecated` from their `getEnv()` anyway, so there's no real reason for them to care about it here.)

### Caveats/negative interactions

This functionality makes an existing bug worse.

Today, spy strategy assignment chains into the base spy and not the `withArgs` context, meaning:

```
spy.withArgs(1).and.returnValue('a').and.returnValue('b');

spy(1);    // => 'a' (user expects 'b')
spy();     // => 'b'
```

Today you could argue that this is non-problem because there's no reason for the end user to do this, but that changes with `.andThen` chains:

```
spy.withArgs(1).and.returnValue('a').andThen.returnValue('b');

spy(1);    // => 'a'
spy(1);    // => 'a' (user expects 'b')
spy();     // => undefined
spy();     // => 'b'
```

I think this is just an existing problem with `withArgs()` that should be fixed, so I didn't tackle it in this PR.

### Alternatives
While writing these I did think of some potential changes we could make...

- If you want to consider `.and.then` instead of `.andThen`, that would isolate all the changes in the SpyStrategy (could remove any impact on the StrategyDispatcher).  Not sure which is better from a user perspective.  I suppose you could also bat around `.and.next` etc.  There is a danger in _any_ method just called "then" being mistaken for a Promise I suppose.  (Could interactive very negatively with some popular linting rules.)

- If we _want_ to hide more implementation detail in SpyStrategy, it's possible to wrap up some stuff (like _planned) into local state in the constructor and expose it only through methods we create in the constructor, instead of prototype methods.  (My take is you've moved away from that, as I remember SpyStrategy used to be super locked down and is now more open.)

- If we _don't_ care to deprecate `spy.and.plan = `, there is an alternative to creating an `.applyPlan` method.  (My first version of this functionality actually just used plan assignment everywhere, and override the plannedProxy's behavior so that `plannedProxy.plan =` enqueues instead of overwriting.)  This approach actually touches a lot _less_ of SpyStrategy, but you wouldn't want to include the deprecation warning in that case.  Also, having an assignment that doesn't assign probably breaks the rule of least surprise.

### Testing
Open to feedback on testing approach here!

I added some unit tests to `SpyStrategySpec`, for basic lines-of-code coverage.  I think the more useful/important unit tests are in `SpySpec`, though, as I would treat these as the "end user spec" -- these describe how the overall interface behaves for the end user.

## Motivation and Context
Provide a standardized way to specify a _series of behaviors_ to execute in series for a spy, without having to implement custom helpers or provide multiple `returnValues`-style variants for every strategy.

## How Has This Been Tested?
- Currently unit tested in node

## Types of changes

> Added a non-breaking deprecation warning in one instance.

## Checklist:

### DIFF
--- a/spec/core/SpySpec.js
+++ b/spec/core/SpySpec.js
@@ -199,6 +199,17 @@ describe('Spies', function () {
       expect(spy()).toEqual(3);
     });
 
+    it("can execute plans in series after being set with separate calls to andThen", function() {
+      var spy = env.createSpy('foo');
+      spy.and.returnValue(1);
+      spy.andThen.callFake(function() { return 1 + 1; });
+      spy.andThen.returnValue(3);
+
+      expect(spy()).toEqual(1);
+      expect(spy()).toEqual(2);
+      expect(spy()).toEqual(3);
+    });
+
     it("can execute plans that raise errors in series", function() {
       var spy = env.createSpy('foo');
       spy.and.throwError('Error A')
@@ -223,11 +234,10 @@ describe('Spies', function () {
       expect(spy()).toEqual(2);
     });
 
-    it("adds a future plan even if there is no plan set yet", function() {
+    it("sets the current plan if no plan was configured yet", function() {
       var spy = env.createSpy('foo');
       spy.andThen.returnValue(1);
 
-      expect(spy()).toBeUndefined();
       expect(spy()).toEqual(1);
     });
 
--- a/spec/core/SpySpec.js
+++ b/spec/core/SpySpec.js
@@ -186,4 +186,82 @@ describe('Spies', function () {
       expect(function() { spy('baz', {qux: 42}); }).toThrowError('Spy \'foo\' received a call with arguments [ \'baz\', Object({ qux: 42 }) ] but all configured strategies specify other arguments.');
     });
   });
+
+  describe("with future plans", function() {
+    it("can execute plans in series", function() {
+      var spy = env.createSpy('foo');
+      spy.and.returnValue(1)
+         .andThen.callFake(function() { return 1 + 1; })
+         .andThen.returnValue(3);
+
+      expect(spy()).toEqual(1);
+      expect(spy()).toEqual(2);
+      expect(spy()).toEqual(3);
+    });
+
+    it("can execute plans that raise errors in series", function() {
+      var spy = env.createSpy('foo');
+      spy.and.throwError('Error A')
+         .andThen.throwError('Error B');
+
+      expect(function() {
+        spy();
+      }).toThrowError('Error A');
+      expect(function() {
+        spy();
+      }).toThrowError('Error B');
+    });
+
+    it("retains the most recently configured plan", function() {
+      var spy = env.createSpy('foo');
+      spy.and.returnValue(1)
+         .andThen.returnValue(2);
+
+      expect(spy()).toEqual(1);
+      expect(spy()).toEqual(2);
+      expect(spy()).toEqual(2);
+      expect(spy()).toEqual(2);
+    });
+
+    it("adds a future plan even if there is no plan set yet", function() {
+      var spy = env.createSpy('foo');
+      spy.andThen.returnValue(1);
+
+      expect(spy()).toBeUndefined();
+      expect(spy()).toEqual(1);
+    });
+
+    it("allows external access to the current plan", function () {
+      var a = function() { },
+          b = function() { },
+          c = function() { };
+      var spy = env.createSpy('foo');
+      spy.and.callFake(a).andThen.callFake(b).andThen.callFake(c);
+
+      expect(spy.and.plan).toBe(a);
+      spy();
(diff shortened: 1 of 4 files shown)


spec/core/SpySpec.js
```diff
@@ -186,4 +186,82 @@ describe('Spies', function () {
...
+      expect(spy()).toEqual(2);
+      expect(spy()).toEqual(2);
+    });
+
+    it("adds a future plan even if there is no plan set yet", function() {
+      var spy = env.createSpy('foo');
+      spy.andThen.returnValue(1);
+
+      expect(spy()).toBeUndefined();
+      expect(spy()).toEqual(1);
```
slackersoft: This seems surprising to me as a user. I think I would expect using `andThen` first to behave as if I had simply used `and`

elliot-nelson:
Changes:
- `spy.andThen` sets current plan if no plan was configured yet.

Regarding your block comment: the extra context is really useful, thanks for the detail.  I wasn't sure if there was a suggested change in there or not (for example, to try and hide more of the additional implementation this PR adds?).

elliot-nelson:
The more I think about this change, the less sure I am I like it.

"Chaining" statements together is tricky if there's not an intentional decision on what is in the chain context.  The `withArgs()` thing is, I feel, perhaps even a blocker for this work; because I don't think the issues can be _fixed_ without a breaking change, which means it's a bug that will sit there forever.

```
// In my opinion, this needs to work as expected:
spy.withArgs(1).and.returnValue(2).andThen.returnValue(3);

// ... but without breaking this!
var spy = spyOn(something).withArgs(1).and.returnValue(2);
spy.and.returnValue(3);
```

With the current proposed syntax, it's impossible to both return a useful chain and return the original spy.  So maybe the proposed syntax is wrong.

I wonder what you think of perhaps solving in a "jQuery-esque" way and explicitly adding both chain start and chain ending functions, which would fix this problem _and_ the withArgs one.

```
spy.withArgs(1).and.chain().returnValue(1).returnValue(2).returnValue(3);
spy.withArgs(2).and.chain().returnValue('a').returnValue('b').throwError();

spy.withArgs(1).and.chain().returnValue(1).returnValue(2).end()
   .withArgs(2).and.returnValue(3).end()
   .withArgs(3).and.chain().rejectValue("error1").rejectValue("error2").end();
```

The big restriction imo is the requirement that any chain should be able to start with `var spy = spyOn().........`.  As long as any function in the chain has to return a Spy, you can't really keep any context in your chain.

A maybe better approach is to use a function and make the "chain" context explicit.  Something like this (and we'd add this to documentation):

```
spy.withArgs(1).and.sequence(s => s.returnValue(1).returnValue(2).throwError());

// Now everything "just works":
var spy = spyOn(something).and.returnValue(3)
    .withArgs('a').and.sequence(s => s.rejectValue().resolveValue());
```

No rush here, so I'll keep noodling on it.  But I'm starting to lean a lot towards a `.and.sequence(s => s.plan().plan().plan())` approach, rather than attempting to force context into the SpyStrategyDispatcher.

slackersoft:
I think it also makes sense to use something like your `and.sequence` since that makes it explicit where all the branches are. I was thinking of that as another option anyways, although I think I said [`and.do`](https://github.com/jasmine/jasmine/pull/1688#issuecomment-488880026) which actually has issues because of `do{}while`.

I'm not a super big fan of the `chain()` type context and I'm not totally sure how it would work with needing to return the _actual_ spy if you forgot to call `end()` or something. I think I would rather see a single strategy that encompasses the chaining concept by itself, like the `and.sequence` proposal. (I expect that to also be a bit easier to implement)

elliot-nelson:
Closing this PR and abandoning the `andThen` chain approach.  I should have a series/sequence version up in a few days.

Issue #1695, opened 2019-05-08
toBeCloseTo fails in edge case (+ fix)


paulvanbrenk:
the `toBeCloseTo` matcher started to fail recently on Chrome 74, i.e. the following cases are failing:

```javascript
describe("sample", function() {
    it("tobecloseto fails", function() {
        expect(4.030904708957288).toBeCloseTo(4.0309, 5);
        expect(4.82665525779431).toBeCloseTo(4.82666, 5);
        expect(-2.82665525779431).toBeCloseTo(-2.82666, 5);
    });
});
```

A possible fix is the following:
[tobecloseto.js#29](https://github.com/jasmine/jasmine/blob/eb30e8179a9073482d07cb7a066183a9fc786a27/src/core/matchers/toBeCloseTo.js#L29) change to:
`pass: Math.round(delta * pow) <= maxDelta * pow`

* Version used: 3.4.0 & 2.99
* Environment name and version: Chrome 74 & NodeJs 12.2.0
* Operating System and version (desktop or mobile): WIndows 10

slackersoft:
Thanks for finding this. I would be happy to review a pull request to fix this that includes some of those examples that fail correctly in Chrome and force the change in implementation.

paulvanbrenk:
I should have some time next week to setup a Jasmine development environment and get a PR ready.

PR #1697, opened 2019-05-10
Add core matcher 'toBeInstanceOf'

elliot-nelson:
## Description
Provide a matcher that gives developers a single way to express "expect `value` to be of type `type`".  The intent is that it can be used for both `typeof` and `instanceof` checks, like so:

Existing Check | With `toBeInstanceOf`
--- | ---
`expect(typeof result).toBe('number')` | `expect(result).toBeInstanceOf(Number)`
`expect(typeof result).toBe('string')` | `expect(result).toBeInstanceOf(String)`
`expect(result instanceof Error).toBe(true)` | `expect(result).toBeInstanceOf(Error)`
`expect(result instanceof MyClass).toBe(true)` | `expect(result).toBeInstanceOf(MyClass)`

## Motivation and Context
Primarily the goal is better failure messages - a slightly more accurate message for `typeof` (primitive) checks, and a dramatically improved message for `instanceof` checks.

```
// Old Messages
Failed: Expected 'number' to be 'string'
Failed: Expected false to be true

// New Messages
Failed: Expected instance of Number to be an instance of String
Failed: Expected instance of Object to be an instance of MyClass
```

## How Has This Been Tested?
- New unit tests (mostly to document the expected behavior)
- Local passes:
  - node v6, v8, v10
  - `JASMINE_BROWSER=firefox node ci.js`
  - `JASMINE_BROWSER=chrome node ci.js` (except for unit test mentioned in https://github.com/jasmine/jasmine/issues/1695)
  - `JASMINE_BROWSER=safari node ci.js`

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/matchers/toBeInstanceOfSpec.js
+++ b/spec/core/matchers/toBeInstanceOfSpec.js
@@ -1,0 +1,235 @@
+describe('toBeInstanceOf', function() {
+  function setPrototypeOf(object, proto) {
+    // Support older environments
+    Object.setPrototypeOf ? Object.setPrototypeOf(object, proto) : object.__proto__ = proto;
+  }
+
+  describe('when expecting Number', function() {
+    it('passes for literal number', function() {
+      var matcher = jasmineUnderTest.matchers.toBeInstanceOf();
+      var result = matcher.compare(3, Number);
+      expect(result).toEqual({
+        pass: true,
+        message: 'Expected instance of Number not to be an instance of Number'
+      });
+    });
+
+    it('passes for NaN', function() {
+      var matcher = jasmineUnderTest.matchers.toBeInstanceOf();
+      var result = matcher.compare(NaN, Number);
+      expect(result).toEqual({
+        pass: true,
+        message: 'Expected instance of NaN not to be an instance of Number'
+      });
+    });
+
+    it('passes for Infinity', function() {
+      var matcher = jasmineUnderTest.matchers.toBeInstanceOf();
+      var result = matcher.compare(Infinity, Number);
+      expect(result).toEqual({
+        pass: true,
+        message: 'Expected instance of Number not to be an instance of Number'
+      });
+    });
+
+    it('fails for a non-number', function() {
+      var matcher = jasmineUnderTest.matchers.toBeInstanceOf();
+      var result = matcher.compare('foo', Number);
+      expect(result).toEqual({
+        pass: false,
+        message: 'Expected instance of String to be an instance of Number'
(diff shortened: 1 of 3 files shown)


slackersoft:
I think you should also be able to use `jasmine.any` to accomplish what you're looking for with Jasmine as it stands now. So:

Existing Check | With jasmine.any
-- | --
expect(typeof result).toBe('number') | expect(result).toEqual(jasmine.any(Number))
expect(typeof result).toBe('string') | expect(result).toEqual(jasmine.any(String))
expect(result instanceof Error).toBe(true) | expect(result).toEqual(jasmine.any(Error))
expect(result instanceof MyClass).toBe(true) | expect(result).toEqual(jasmine.any(MyClass))

The assertions certainly don't read quite as nicely as an actual `toBeInstanceOf` check, but they should provide you with a similar level of error message when they fail.

I still think a `toBeInstanceOf` matcher could be useful in giving _better_ messages, but it might make sense to take a quick look at `jasmine.any` and see what kind of shared detection logic we can pull out to ensure they match the same things.

elliot-nelson:
:+1: Great point!

- Modified matcher source to make use of existing jasmine functionality, rather than doing it myself.
- Updated a few minor specs to reflect how this changes the behavior.  (For the better, mostly, since it works exactly like it would if you used `jasmine.any()` matchers.)

Possible additional changes:

- Test bed might be overkill if we're relying on `jasmine.any()`.  I could strip it down to a few basic examples showing we're passing through to `Any`, plus the corner cases for null/undefined/etc.  On the flip side, keeping the test bed would notify anyone who makes substantial future changes to `jasmine.Any` that they are making potential breaking changes to the behavior of this matcher...

- Naming... `toBeInstanceOf(Number)` feels a little better and even more conversational as `.toBeA(String)`, `.toBeA(Number)`.  But, you'd probably need to support the _alias_ `.toBeAn(Object)`, `.toBeAn(Error)` as while (not sure if we have any precedent for that yet).  Distant third would be `.toBeOfType(Number)` -- clear, but pretty formal sounding.

slackersoft:
I think this looks good. I like the fully specified tests here to clearly state how it works. I think the current name is the clearest.

I'm not entirely sure what's going on with the PhantomJS build for this, but you might be able to rebase onto master to get the CI fixes and see if that gets us some more logging.

elliot-nelson:
The PhantomJS build correctly caught a raw `async` function.

I've fixed this PR, and I put up a separate PR to improve the error reporting in that situation.

PR #1698, opened 2019-05-10
Improve error handling in CI test launcher

elliot-nelson:
## Description
Small tweaks to the `ci.js` test launcher:

- Ensure that if an error occurs in the launcher (mismatched Chrome browser + ChromeDriver version, or missing GeckoDriver, etc.), the details of the error will be logged to console.
- Ensure that if the driver fails to start, the server still gets closed down, allowing the test launcher to exit immediately instead of hanging forever.
- As a side benefit, the `ci.js` launcher now works in latest node v8.x _and_ v10.x.  (Although I'm fine bringing the `.finally` back if this isn't important -- we only have ~6 months to live on node 8.)

## Motivation and Context
Minor improvements that make it easier for new contributors to run CI tests in the browser locally.  Probably has no effect on CI (although for corner cases, may improve debugging).

## How Has This Been Tested?
- So far, tested locally only, with latest Safari, Chrome, and Firefox.

## Types of changes

## Checklist:

### DIFF
--- a/ci.js
+++ b/ci.js
@@ -121,12 +121,10 @@ async function getResults(driver) {
 }
 
 function cleanup() {
-  return Promise.resolve().then(() => {
-    return Promise.all([
-      driver ? driver.close() : true,
-      new Promise(resolve => server ? server.close(resolve) : resolve())
-    ]);
-  }).catch(error => {
+  return Promise.all([
+    driver ? driver.close() : true,
+    new Promise(resolve => server ? server.close(resolve) : resolve())
+  ]).catch(error => {
     console.error(error);
   });
 }
--- a/ci.js
+++ b/ci.js
@@ -120,6 +120,17 @@ async function getResults(driver) {
   return {specResults: flatten(specResults), failedSuiteResults: flatten(failedSuiteResults)};
 }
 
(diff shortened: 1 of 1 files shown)


ci.js
```diff
@@ -120,6 +120,17 @@ async function getResults(driver) {
   return {specResults: flatten(specResults), failedSuiteResults: flatten(failedSuiteResults)};
 }
 
+function cleanup() {
+  return Promise.resolve().then(() => {
```
slackersoft: What's the reason for this wrapping `Promise.resolve().then`? I would expect just `Promise.all().catch` to work correctly here.

elliot-nelson: We could probably scrap that, actually...

Originally I couldn't tell what was going wrong, because the error message was getting lost in the ether, so I wrapped the whole thing.  Then afterwards I realized the issue was driver wasn't defined yet, and added the inline checks.

TLDR: both checks are too much and could be simplified now.

slackersoft: Sweet. Please update to include the simplified checks. Thanks!

elliot-nelson:
Simplified check (removed extra promise wrapper).

Issue #1575, opened 2018-07-06
ERROR: TypeError: s.substring is not a function


bushuyev:
## Expected Behavior
jasmin nicely print error
## Current Behavior
fails with murky message

## Suite that reproduces the behavior (for bugs)

```javascript
describe("sample", function() {
  it('d', async(()=>{
//TypeError, like  (stack is from when it's wropped in try /catch)
TypeError: Cannot read property '$injector' of undefined
    at http://localhost:9876/_karma_webpack_/webpack:/src/tests/newPayment.spec.ts:214:41
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone.js:388:1)
    at AsyncTestZoneSpec.push../node_modules/zone.js/dist/zone-testing.js.AsyncTestZoneSpec.onInvoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-testing.js:713:1)
    at ProxyZoneSpec.push../node_modules/zone.js/dist/zone-testing.js.ProxyZoneSpec.onInvoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-testing.js:285:1)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone.js:387:1)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.run (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone.js:138:1)
    at http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone.js:872:1
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone.js:421:1)
    at AsyncTestZoneSpec.push../node_modules/zone.js/dist/zone-testing.js.AsyncTestZoneSpec.onInvokeTask (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/dist/zone-testing.js:697:1)
[...]

PR #1700, opened 2019-05-14
Prevent error in case of non string value in 's' parameter

masepe:
ERROR: TypeError: undefined is not a constructor (evaluating 's.substring(0, maxlen - 4)') truncate

PhantomJS 2.1.1 (Windows 8.0.0). error shown i console/karma report

## Description

Make method more safe to input

## Motivation and Context

ERROR: TypeError: undefined is not a constructor (evaluating 's.substring(0, maxlen - 4)') truncate

## How Has This Been Tested?

Tested with modification in my project node_modules folder

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -5487,7 +5487,7 @@ getJasmineRequireObj().pp = function(j$) {
       return { value: s, truncated: false };
     }
 
-    s = s.substring(0, maxlen - 4) + ' ...';
+    s = ('' + s).substring(0, maxlen - 4) + ' ...';
     return { value: s, truncated: true };
   }
 


elliot-nelson:
Looks like line in question is in `PrettyPrinter` (https://github.com/jasmine/jasmine/blob/master/src/core/PrettyPrinter.js#L284).

I'm curious if this error is occurring organically (for example, it's a normal `expect` or `spyOn` call), or if you're using a jasmine util directly -- for example, are you calling `jasmine.pp()` directly on an object (and if so, what is it), or maybe even calling `.append()` directly?

(The reason I'm interested is that while your fix does make the internal method safer, it probably doesn't produce ideal output -- if `s` is not a string, but an array, we should have detected that a a higher level and called `emitArray`; if it's an object, we should have called `emitObject`, etc., to get the desired formatting from PrettyPrinter.  So maybe there's a check missing in the higher-level `format()` function that needs to be addressed.)

slackersoft:
By the time this portion of the PrettyPrinter has been invoked, Jasmine really thinks it should have a string because we've most likely just called `.toString()` on the object in question. As I said in #1575 though, it might make sense to have a bit more protection around bad implementations of `.toString()` on objects out of our control.

Unfortunately this pull request can't be merged as it stands right now for a few reasons:
* You haven't included any tests for this change. As a testing library, Jasmine expects to have tests for all of its code.
* You have included the generated file `jasmine.js` in your Pull Request which can lead to various merge conflicts between unrelated work.
* You haven't made your changes to the relevant files in `src` which means that a future regeneration of the `jasmine.js` file will lose your changes.

Please read the [CONTRIBUTING](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md) guide.

masepe:
Hey guys, i am not allowed to spent my working time on ext. open source projects (that's bad). With my pull request i wanted more to point on error in library code, which causes user main problem hiding. In my opinion it is better to see something  like [object] in output than exception with useless stack-trace bringing to the library code. 
If it is possible, could you please fix it in correct place/way.

elliot-nelson:
I think it's a reasonable thing to fix, I have opened https://github.com/jasmine/jasmine/pull/1711 with a unit test.

PR #1716, opened 2019-06-02
Allow users to set a default spy strategy

elliot-nelson:
## Description

Allow a user to set a "default spy strategy", which assigns the default behavior of any spies created.  You can do this inline in a particular spec, or in a suite's `beforeEach` or a global `beforeEach`, depending on your needs.

## Motivation and Context

If you're creating many spies and don't want to specify their strategies, or want a particular global spy handler, this allows you to do so.

In particular, a strategy like this may be useful, depending on your application:

```
beforeEach(function() {
    jasmine.setDefaultSpyStrategy(and =>
        and.throwError('Unexpected call on spy `' + and.identity + '`')
    );
});
```

## Next Steps

This is a first stab at the concept.  I originally wanted an interface where you would interact directly with a "default spy strategy" property, like this:

    jasmine.defaultSpyStrategy.returnValue(42);

The problem with this is you don't have the flexibility you need to, for example, use properties of _this spy_ (like `and.identity` or `and.originalFn`) in your default function.  So, you really need to have the user provide a callback function, rather than interacting directly with an object.

It might make more sense (a _lot_ more sense), rather than this `and => and.blah()` syntax, to instead have the function be `setSpyDefaults`, which takes a callback function on `spy`.  In this case the user would do:

    jasmine.setSpyDefaults(spy => spy.and.callFake(() => 'You called ' + spy.and.identity));

(Next steps would be to decide what the most intuitive user interface is, then clean up the source + specs.)

## How Has This Been Tested?
Locally on Chrome, node 8, node 10.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/integration/DefaultSpyStrategySpec.js
+++ b/spec/core/integration/DefaultSpyStrategySpec.js
@@ -1,0 +1,68 @@
+describe('Default Spy Strategy (Integration)', function() {
+  var env;
+
+  beforeEach(function() {
+    env = new jasmineUnderTest.Env();
+    env.configure({random: false});
+  });
+
+  it('allows defining a default spy strategy', function(done) {
+    env.describe('suite with default strategy', function() {
+      env.beforeEach(function() {
+        env.setDefaultSpyStrategy(function (and) {
+          and.returnValue(42);
+        });
+      });
+
+      env.it('spec in suite', function() {
+        var spy = env.createSpy('something');
+        expect(spy()).toBe(42);
+      });
+    });
+
+    env.it('spec not in suite', function() {
+      var spy = env.createSpy('something');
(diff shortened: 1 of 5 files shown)


slackersoft:
This looks reasonable. The only real reason I can think of to pass the `spy` itself back instead of the `and` is to allow a default spy strategy to _more easily_ get access to the `spy` to check for things like call count. Given that you can use `and.getSpy()` for this the basic use case looks like it reads better this way.

It looks like you've got some styling stuff breaking your build, please clean that up and we can get this merged.

elliot-nelson:
:+1: I saw you got some eslint and prettier going in master!

(Ran npm cleanup and submitted another commit.)

PR #1719, opened 2019-06-07
Allow individual spyOn/spyOnProperty calls to allow respy

elliot-nelson:
## Description

Add an optional final parameter to `spyOn` and `spyOnProperty` that allows the user to override `allowRespy` to `true`, even if the global default is `false`.

(Or, allows you to override `allowRespy` to `false` if the global default is `true` -- not sure if there's a use case here, but the behavior is consistent.)

## Motivation and Context

This change is designed to fill a particular API hole today: if you have a sprawling spec and a test needs to override the behavior of a property, but that property is already a spy, it's not possible to do this unless you go and _save_ the original result of the `spyOnProperty` call (which may be difficult, if you're dealing with a tapestry of test helpers and beforeEach's etc.).  This probably means the spec has some code smell, but at least you won't be blocked anymore:

```
spyOnProperty(myModule, 'previouslySpiedProp', 'set', true).and.throwError('you cannot set that');
```

This isn't an API hole for function spies, but, I've added it there as well just for consistency.

```
spyOn(myModule, 'foo');
spyOn(myModule, 'foo').and.stub();  // <-- error!

// Just replace with:
myModule.foo.and.stub();

// Or, now, offered for consistency:
spyOn(myModule, 'foo', true).and.stub();
```

## How Has This Been Tested?
- local: Node 8, Node 10, Chrome, Firefox

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/SpyRegistrySpec.js
+++ b/spec/core/SpyRegistrySpec.js
@@ -64,6 +64,21 @@ describe('SpyRegistry', function() {
       }).toThrowError(/has already been spied upon/);
     });
 
+    it('allows overriding respy by passing respy option as true', function() {
+      var spies = [],
+        spyRegistry = new jasmineUnderTest.SpyRegistry({
+          currentSpies: function() {
+            return spies;
+          },
+          createSpy: createSpy
+        }),
+        subject = { spiedFunc: function() {} };
+
+      var originalSpy = spyRegistry.spyOn(subject, 'spiedFunc');
+
+      expect(spyRegistry.spyOn(subject, 'spiedFunc', true)).toBe(originalSpy);
+    });
+
     it('checks if it can be spied upon', function() {
       var scope = {};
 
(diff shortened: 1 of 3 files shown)


elliot-nelson:
The more I think about this, the less I like it.  Adding booleans all over the place seems like a bad time in the long run.

I wonder if it's simply "OK" that it's annoying to modify property spies, and instead, to add an article to jasmine docs dedicated to working with property spies?  I think the issue right now is we don't have an _example_ of how to do it, e.g.,

```
var obj = createBigFancyObject();

// Spy on property
spyOnProperty(obj, 'field', 'get').and.returnValue('foo');

// Later, change the behavior of property
Object.getOwnPropertyDescriptor(obj, 'field').get.and.returnValue('bar');
```

sgravrock:
> The more I think about this, the less I like it. Adding booleans all over the place seems like a bad time in the long run.
> 
> I wonder if it's simply "OK" that it's annoying to modify property spies, and instead, to add an article to jasmine docs dedicated to working with property spies?

I think so. I tend to read `spyOn(myModule, 'foo').and.stub()` and `myModule.foo.and.stub()` as conveying different intent. When I'm reading a spec, the latter tells me that the author intended to override an existing spy at that point. The former doesn't give me a clear indication of whether the respy is intentional or represents a bug in the test suite. So at first glance I think it's probably better to steer people toward the `myModule.foo.and.stub()` approach rather than making respy support more robust than it already is.

The other thing that might be nice, in addition to the docs, is to improve the error message so it tells the user what to do.

elliot-nelson:
Closing as not necessary.

PR #1720, opened 2019-06-07
Improved reporting to provide details of passed assertions

GCHQDeveloper911:
As per the title - rather than just report "Passed." for passed assertions, include details of values involved etc.

## Description
As above. Also as part of this, I did a bit of rationalisation of some of the comparators, and renamed the likes of "BuildFailureMessage" to become "BuildOutputMessage" etc (since they no longer applied to failures only.

## Motivation and Context
For tests with numerous assertions, an endless stream of "Passed." messages can be uninformative. These changes allow the user to see what values was compared with what other value.

## How Has This Been Tested?
Tests have been updated as appropriate, but also used "in anger" as part of my testing (day to day) work here.

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2008-2018 Pivotal Labs
+Copyright (c) 2008-2019 Pivotal Labs
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
@@ -1706,9 +1706,8 @@ getJasmineRequireObj().Env = function(j$) {
         throw new Error('\'fail\' was used when there was no current spec, this could be because an asynchronous test timed out');
       }
 
-      var message = 'Failed';
+      var message = '';
       if (error) {
-        message += ': ';
         if (error.message) {
           message += error.message;
         } else if (j$.isString_(error)) {
@@ -2908,12 +2907,13 @@ getJasmineRequireObj().Expectation = function(j$) {
     if (result.message) {
       if (j$.isFunction_(result.message)) {
         return result.message();
-      } else {
-        return result.message;
       }
     }
 
     args = args.slice();
+    args.unshift(result.messageSuffix);
+    args.unshift(result.expectedQualifier);
+    args.unshift(result.actualQualifier);
     args.unshift(true);
     args.unshift(matcherName);
     return util.buildFailureMessage.apply(null, args);
@@ -3046,11 +3046,15 @@ getJasmineRequireObj().buildExpectationResult = function() {
 
     return result;
 
+    
     function message() {
       if (options.passed) {
-        return 'Passed.';
+        return 'Passed: ' + options.message;
       } else if (options.message) {
-        return options.message;
+        // when formatting caught error messages, the 'Failed' prefix can be erroneously duplicated. Needs to be tidied.
+        var localMessage = typeof options.message === 'function' ? 'Failed: ' + options.message() : 'Failed: ' + options.message;
+        localMessage = localMessage.replace ('Failed: Failed:', 'Failed:');
+        return localMessage;
       } else if (options.error) {
         return messageFormatter(options.error);
       }
@@ -3059,7 +3063,7 @@ getJasmineRequireObj().buildExpectationResult = function() {
 
     function stack() {
       if (options.passed) {
-        return '';
+        return options.message;
       }
 
       var error = options.error;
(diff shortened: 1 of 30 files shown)


slackersoft:
As we discussed in #1634, I don't think this is a thing that Jasmine is likely to add. The generated message for an expectation that has passed will indicate a failure state as if the expectation had a `.not` when it didn't, or did not when it did.

It also looks like you included the generated `jasmine.js` file in your Pull Request and there are a number of other conflicts which make this PR hard to merge in as well.

elliot-nelson:
I'm also curious where you're seeing all these "Passed" messages?  (What `Reporter` that is, I mean.)

@slackersoft mentioned the technical difficulties in jasmine with making the change you've suggested, but I think there's also a consistency problem -- it is standard behavior in all testing frameworks to hide output from _passing_ tests, because if a test is passing it's intended to be uninteresting.

As an example, try editing any old JUnit XML test results file and add a bunch of detailed output into the body of a passing test, and then feed it into Xcode or Jenkins -- you'll notice that all of the output is deleted, because only _failing_ tests will retain the debugging/failure output.  For Jasmine to behave differently would potentially be swimming upstream.

GCHQDeveloper911:
Hi @elliot-nelson 

I agree there is the issue with seeing the changes in reporters. However it's a bit of a chicken an egg situation! I've been working with the devs of Webdriverio, and they have merged changes to some of the packages there - both webdriverIO core packages, and one of the reporters (wdio-spec-reporter) that support this change. 

It would indeed be something that individidual reporters would need to change to support. However, I genuinely believe this has benefit. As I believe I suggested in previous discussion on the subject, realistically Jasmine users will often have tests with multiple assertions. If everything passes, then great! However, if there is say, a single failure, then I believe there is a strong argument for being able to see exactly what values were used in all assertions, to potentially help understand and establish the true point of failure.

GCHQDeveloper911:
@slackersoft in the discussion in #1634 You quoted an example of the issue described in 'toBe' - but we agreed it didn't apply there (unlucky example comparator!)... Could you give me one or two comparators that DO demonstrate the behaviour you describe, so I can investigate and ensure my changes don't break this?

slackersoft:
As [I mentioned](https://github.com/jasmine/jasmine/issues/1634#issuecomment-463754394) in #1634, the particular example (`toBe` at that time) didn't necessarily exhibit the particular behavior (I believe it actually does now if the `actual` is an `Object`), I picked it because it was easy to reason about what _should_ happen. Just because that example was bad at the time doesn't mean the issue doesn't exist. The issue I'm concerned about with reporting a message for passed expectations is _not_ that something will necessarily break, but that the messages provided won't make any sense. _Any_ matcher, custom or built-in, that specifies its own message instead of relying on the wrapping `Expectation` object to generate one for it, is expected to provide a message that Jasmine should show in case the current state is not the actual desired state. You should also be able to see this with `toHaveBeenCalled`.

Additionally:
> However, if there is say, a single failure, then I believe there is a strong argument for being able to see exactly what values were used in all assertions, to potentially help understand and establish the true point of failure.

This seems to indicate to me that your assertions are too vague in what the expect, or you are testing too big of a portion of your codebase. I would expect the fact that the expectation has passed to be enough information about what the `actual` value is.

GCHQDeveloper911:
I wouldn't say our assertions are vague.... but it is not possible for us to work with static test data - our data is constantly changing. Therefore, our assertions tend to be less "check that some_var=7" and more "check that some_var=the same variable as captured elsewhere in the application". This in turn means that our tests often *have* to work their way through a somewhat lengthy process through the app, just to get to the point we want to check. 

In turn, if we get to this final check and get a fail, it would be nice to know where in that lengthy process things went wrong - having checks along the way, reporting pass / fail can help narrow down the fault. 

I appreciate that many users of Jasmine are lucky enough to have static data, and can just pass in a URL to jump straight to a specific point in the app. We're not that lucky, and can't just do that - but to be honest, I very much doubt we're alone there.

sgravrock:
Closing, since this has clearly failed to reach consensus.

I tend to agree with the philosophical objections to this feature. More importantly, the problem of misleading messages from custom matchers (e.g. a passing `expect(foo).toBeBar()` being reported as "expected foo not to be bar") is a show-stopper. It would need to be solved before a feature like this could move forward. And at this stage in Jasmine's development, it would need to be solved without breaking changes to custom matchers. I don't think that's possible.

The [debug logging feature](https://jasmine.github.io/api/4.5/jasmine.html#.debugLog) introduced in 4.0 was designed to address use cases that are similar to the one that motivated this PR.

PR #1732, opened 2019-07-09
add custom async matchers

### LINKED ISSUE #1703: Ability to add custom AsyncMatchers
While testing using the great new `expectAsync` matchers, I ran into a deficiency where `toBeResolvedTo` does not print what it was rejected with on rejections.

```javascript
await expectAsync(
    promiseReturningThing()
).toBeResolvedTo(jasmine.any(Buffer));
```

becomes

```
Expected a promise to be resolved to <jasmine.any(Buffer)> but it was rejected.
```

I tried to overcome that with a custom async matcher that is a direct copy of `toBeResolvedTo` but that prints out the rejection cause for easier debugging.

```javascript
import {toBeDescriptivelyResolvedTo} from "../../matchers/async/toBeDescriptivelyResolvedTo";
jasmine.addMatchers({
    toBeDescriptivelyResolvedTo: toBeDescriptivelyResolvedTo
})
```

However, I found that custom matchers **do not** count as async matchers.

## Expected Behavior

I expect to be able to define a custom async matcher.

```javascript
await expectAsync(
    promiseReturningThing()
).toBeDescriptivelyResolvedTo(jasmine.any(Buffer));
```

and have it print out the rejected value.

```
Expected a promise to be resolved to <jasmine.any(Buffer)> but it was rejected with "fake rejection".
```

## Current Behavior

I get
`TypeError: expectAsync(...).toBeDescriptivelyResolvedTo is not a function`

I can test it with a normal expect and it works, as far as being called, but then I get a failure from expect itself not working with the promise.

```javascript
await (expect(Promise.resolve('b')) as any).toBeDescriptivelyResolvedTo(2);
```

```
Expected [object Promise] to be descriptively resolved to 2.
```

I can verify it is being called by adding console.log statements and seeing them output in addition to the failure.

## Possible Solution

A possible workaround might exist by manipulating some of the internal globals, but it might be tricky as they are not all exposed. The correct code is probably going to look like something around `src/core/Expectation.js`, specifically `addAsyncCoreMatchers`. The current code is able to add the core async matchers.

`j$.Expectation` or `j$.asyncMatchers` seem useful if one could get a reference.

Also note [this todo](https://github.com/jasmine/jasmine/blob/master/src/core/matchers/matchersUtil.js#L2) about `j$.pp`, because [`toBeResolvedTo`](https://github.com/jasmine/jasmine/blob/master/src/core/matchers/async/toBeResolvedTo.js) uses it.

If my problem is a typing issue that would be great and I can open an issue elsewhere.

```javascript
export const toBeDescriptivelyResolvedTo  =
    /**
     * Expect a promise to be resolved to a value equal to the expected, using deep equality comparison.
     * @function
     * @async
     * @name async-matchers#toBeResolvedTo
     * @param {Object} expected - Value that the promise is expected to resolve to
     * @example
     * await expectAsync(aPromise).toBeResolvedTo({prop: 'value'});
     * @example
     * return expectAsync(aPromise).toBeResolvedTo({prop: 'value'});
     */
    function toBeResolvedTo(util: any, customEqualityTesters: any) {
        console.log("eggo");
        return {
            compare: function(actualPromise: any, expectedValue: any) {
                console.log("waffle");
                function prefix(passed: any) {
                    return 'Expected a promise ' +
                        (passed ? 'not ' : '') +
                        'to be resolved to ' + JSON.stringify(expectedValue);
                }

                return actualPromise.then(
                    function(actualValue: any) {
                        if (util.equals(actualValue, expectedValue, customEqualityTesters)) {
                            return {
                                pass: true,
                                message: prefix(true) + '.'
                            };
                        } else {
                            return {
                                pass: false,
                                message: prefix(false) + ' but it was resolved to ' + JSON.stringify(actualValue) + '.'
                            };
                        }
                    },
                    function(err: any) {
                        return {
                            pass: false,
                            message: prefix(false) + ' but it was rejected with ' + JSON.stringify(err) + '.'
                        };
                    }
                );
            }
        };
    };

---

      await (expectAsync(Promise.reject('b')) as any).toBeDescriptivelyResolvedTo('a');

```

## Context

Not printing the value makes debugging failed tests harder.

Not being able to extend custom async matchers reduces Jasmine's power.

## Your Environment

* Version used: 3.4.0
* Environment name and version (e.g. Chrome 39, node.js 5.4): Node 10
* Operating System and version (desktop or mobile): Alpine Linux Docker container
* Link to your project: N/A

UziTech:
Add custom async matchers with `jasmine.addAsyncMatchers()`

## Description

add custom async matchers with `jasmine.addAsyncMatchers()`

I removed the requirement for expectAsync to be passed a promise so it would be closer to [Async Matchers](https://jestjs.io/docs/en/expect.html#async-matchers) in Jest.

```js
jasmine.addAsyncMatchers({
  toMatchAsync: () => {
    return {
      compare: async (actual, expected) => {
        const value = await doSomethingAsync(actual);
        const result = {
          pass: value === expected
        };
        result.message = await getAsyncMessage(expected, result.pass);
        return result;
      }
    };
  }
});

it('matches an async value', async () => {
  await expectAsync(1).toMatchAsync(1);
});
```

## Motivation and Context

Now custom matchers and can do something asynchronous with the `actual` or `expected` values.

fixes #1703

## How Has This Been Tested?

Added tests in AsyncExpectationSpec.js

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/AsyncExpectationSpec.js
+++ b/spec/core/AsyncExpectationSpec.js
@@ -225,6 +225,587 @@ describe('AsyncExpectation', function() {
     });
   });
 
+  describe('async matchers', function() {
+    it('makes custom matchers available to this expectation', function() {
+      var asyncMatchers = {
+          toFoo: function() {},
+          toBar: function() {}
+        },
+        expectation;
+
+      expectation = jasmineUnderTest.Expectation.asyncFactory({
+        customAsyncMatchers: asyncMatchers
+      });
+
+      expect(expectation.toFoo).toBeDefined();
+      expect(expectation.toBar).toBeDefined();
+    });
+
+    it("wraps matchers's compare functions, passing in matcher dependencies", function() {
+      var fakeCompare = function() {
+          return Promise.resolve({ pass: true });
+        },
+        matcherFactory = jasmine
+          .createSpy('matcher')
+          .and.returnValue({ compare: fakeCompare }),
+        matchers = {
+          toFoo: matcherFactory
+        },
+        util = {
+          buildFailureMessage: jasmine.createSpy('buildFailureMessage')
+        },
+        customEqualityTesters = ['a'],
+        addExpectationResult = jasmine.createSpy('addExpectationResult'),
+        expectation;
+
+      expectation = jasmineUnderTest.Expectation.asyncFactory({
+        util: util,
+        customAsyncMatchers: matchers,
+        customEqualityTesters: customEqualityTesters,
+        actual: 'an actual',
+        addExpectationResult: addExpectationResult
+      });
+
+      return expectation.toFoo('hello').then(function() {
+        expect(matcherFactory).toHaveBeenCalledWith(
+          util,
+          customEqualityTesters
+        );
+      });
+    });
+
+    it("wraps matchers's compare functions, passing the actual and expected", function() {
+      var fakeCompare = jasmine
+          .createSpy('fake-compare')
+          .and.returnValue(Promise.resolve({ pass: true })),
+        matchers = {
+          toFoo: function() {
+            return {
+              compare: fakeCompare
+            };
+          }
+        },
+        util = {
+          buildFailureMessage: jasmine.createSpy('buildFailureMessage')
+        },
+        addExpectationResult = jasmine.createSpy('addExpectationResult'),
+        expectation;
+
+      expectation = jasmineUnderTest.Expectation.asyncFactory({
+        util: util,
+        customAsyncMatchers: matchers,
+        actual: 'an actual',
+        addExpectationResult: addExpectationResult
+      });
+
+      return expectation.toFoo('hello').then(function() {
+        expect(fakeCompare).toHaveBeenCalledWith('an actual', 'hello');
+      });
+    });
+
+    it('reports a passing result to the spec when the comparison passes', function() {
+      var matchers = {
+          toFoo: function() {
+            return {
+              compare: function() {
+                return Promise.resolve({ pass: true });
+              }
+            };
+          }
+        },
+        util = {
+          buildFailureMessage: jasmine.createSpy('buildFailureMessage')
+        },
+        addExpectationResult = jasmine.createSpy('addExpectationResult'),
+        errorWithStack = new Error('errorWithStack'),
+        expectation;
+
+      spyOn(jasmineUnderTest.util, 'errorWithStack').and.returnValue(
+        errorWithStack
+      );
+
+      expectation = jasmineUnderTest.Expectation.asyncFactory({
+        customAsyncMatchers: matchers,
+        util: util,
+        actual: 'an actual',
+        addExpectationResult: addExpectationResult
+      });
+
+      return expectation.toFoo('hello').then(function() {
+        expect(addExpectationResult).toHaveBeenCalledWith(true, {
+          matcherName: 'toFoo',
+          passed: true,
+          message: '',
+          error: undefined,
+          expected: 'hello',
+          actual: 'an actual',
+          errorForStack: errorWithStack
+        });
+      });
+    });
+
+    it('reports a failing result to the spec when the comparison fails', function() {
+      var matchers = {
+          toFoo: function() {
+            return {
+              compare: function() {
+                return Promise.resolve({ pass: false });
+              }
+            };
+          }
+        },
+        util = {
+          buildFailureMessage: function() {
+            return '';
+          }
+        },
+        addExpectationResult = jasmine.createSpy('addExpectationResult'),
+        errorWithStack = new Error('errorWithStack'),
+        expectation;
+
+      spyOn(jasmineUnderTest.util, 'errorWithStack').and.returnValue(
+        errorWithStack
+      );
+
+      expectation = jasmineUnderTest.Expectation.asyncFactory({
+        customAsyncMatchers: matchers,
+        util: util,
+        actual: 'an actual',
+        addExpectationResult: addExpectationResult
+      });
+
+      return expectation.toFoo('hello').then(function() {
+        expect(addExpectationResult).toHaveBeenCalledWith(false, {
+          matcherName: 'toFoo',
+          passed: false,
+          expected: 'hello',
+          actual: 'an actual',
+          message: '',
+          error: undefined,
+          errorForStack: errorWithStack
+        });
+      });
+    });
+
+    it('reports a failing result and a custom fail message to the spec when the comparison fails', function() {
+      var matchers = {
+          toFoo: function() {
+            return {
+              compare: function() {
+                return Promise.resolve({
+                  pass: false,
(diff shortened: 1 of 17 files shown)


UziTech:
I'm wondering if it would be better to add async support for regular matchers?

```js
jasmine.addMatchers({
  toMatchAsync: () => {
    return {
      compare: async (actual, expected) => {
        const value = await doSomethingAsync(actual);
        const result = {
          pass: value === expected
        };
        result.message = await getAsyncMessage(expected, result.pass);
        return result;
      }
    };
  }
});

it('matches an async value', async () => {
  await expect(1).toMatchAsync(1);
});
```

It doesn't seem like it would be hard to check if the compare function returns a promise:

```diff
  function wrapSyncCompare(name, matcherFactory) {
    return function() {
      var result = this.expector.compare(name, matcherFactory, arguments);
+     if (j$.isPromiseLike(result)) {
+       return result.then(function (returnedResult) {
+         this.expector.processResult(returnedResult);
+       });
+     }
      this.expector.processResult(result);
    };
  }
```

sgravrock:
As I mentioned in [my comment on your other PR](https://github.com/jasmine/jasmine/pull/1733#issuecomment-513324966), I think `addAsyncMatchers` is the way to go.

Could you please pull out the removal of the requirement for expectAsync to be passed a promise into a separate PR? It seems orthogonal to `addAsyncMatchers` and I think it wants more discussion.

UziTech:
> Could you please pull out the removal of the requirement for expectAsync to be passed a promise into a separate PR?

The purpose of this PR is to include custom async matchers, not to include matchers that can include async arguments.

UziTech:
One example I can think of off the top of my head is to find out a database value given an `id`

```js
await expectAsync(id).toHaveValue(1);
```

the database call needs to be async not the id

UziTech:
otherwise we will have a lot of people just doing

```js
await expectAsync(Promise.resolve(id)).toHaveValue(1);
```

it seems pointless

src/core/Expector.js
```diff
@@ -84,7 +84,7 @@ getJasmineRequireObj().Expector = function(j$) {
       matcherName: this.matcherName,
       passed: result.pass,
       message: message,
-      error: errorForStack ? undefined : result.error,
+      error: result.error,
```
UziTech: I'm not sure if this should change but I thought it would be better if async matchers were as close to regular matchers as possible

slackersoft: This difference is intentional so the stack trace that is displayed to the user includes the lines of code to the test and expectation. The `error` here _should_ be passed in as undefined in order for Jasmine's `ExpectationResult` to generate the correct stack trace.

slackersoft:
I think the basic `addAsyncMatchers` functionality looks pretty good, but we should revert the error reporting change in `Expector`.

I think the disconnect we're looking at for the two features in the PR (custom matchers for a given Promise vs. custom matchers that need to use a Promise) might be more around intended use case and whether the interface for the two should be the same. I'm comfortable pulling in the custom async matchers stuff now, but would also like to have a bit more discussion around the interface for the other use case. If you don't want to separate it out we can just have that discussion here.

src/core/Expector.js
UziTech: What about when the matcher resolves with an `error` property? Should `ExpectationResult` generate the stack for the returned `error` or the `errorForStack`?

UziTech: If I leave it the way it was the `error` returned by the matcher is just ignored.

UziTech:
> custom matchers for a given Promise vs. custom matchers that need to use a Promise

That is why i debated letting `expect` work with async matchers in #1733, but I agree with @sgravrock that it would be easier to forget to `await` the call to `expect`.

src/core/Expector.js
UziTech: Should the `error` returned by the matcher just be ignored in async matchers?

src/core/Expectation.js
```diff
@@ -62,8 +62,12 @@ getJasmineRequireObj().Expectation = function(j$) {
       );
     }
 
-    if (!j$.isPromiseLike(this.expector.actual)) {
```
sgravrock: I hear that you want to be able to support matchers that do async work themselves and not just matchers that work on actuals that are promises. I'm not necessarily opposed to that, and I think there's a case to be made for doing that at the same time as adding custom async matchers if we're going to do it at all. But as it stands right now, this PR makes the error message less clear if any of the existing async matchers are called with an actual that isn't a promise.

Before:
```
Error: Expected expectAsync to be called with a promise.
    at <Jasmine>
    at UserContext.<anonymous> (http://localhost:8888/__spec__/core/integration/EnvSpec.js:2499:12)
    at <Jasmine>
```

After:
```
TypeError: actual.then is not a function
    at <Jasmine>
    at UserContext.<anonymous> (http://localhost:8888/__spec__/core/integration/EnvSpec.js:2614:28)
    at <Jasmine>
```

So I think that what we should do for now is either put the type guard back, or add a similar check to each of the existing async matchers. Either way, when somebody does something like `expectAsync('not a promise').toBeResolved()`, the error message should clearly indicate that the actual needs to be a promise.

spec/core/AsyncExpectationSpec.js
```diff
@@ -235,6 +225,583 @@ describe('AsyncExpectation', function() {
     });
   });
 
+  describe('async matchers', function() {
```
sgravrock: These specs look good except that they fail in environments that don't support promises, notably IE. You can fix that by adding a call to `jasmine.getEnv().requirePromises()` either to each spec or to a `beforeEach`.

src/core/matchers/async/toBeRejected.js
```diff
@@ -10,9 +10,12 @@ getJasmineRequireObj().toBeRejected = function(j$) {
    * @example
    * return expectAsync(aPromise).toBeRejected();
    */
-  return function toBeResolved(util) {
+  return function toBeRejected(util) {
     return {
       compare: function(actual) {
+        if (!j$.isPromiseLike(actual)) {
+          throw new Error('Expected toBeRejected to be called on a promise.');
```
UziTech: I changed the core async matchers to check for a promise and throw an error if `actual` isn't a promise

src/core/Expectation.js
```diff
@@ -1,10 +1,4 @@
 getJasmineRequireObj().Expectation = function(j$) {
-  var promiseForMessage = {
-    jasmineToString: function() {
-      return 'a promise';
```
UziTech: Since the `actual` isn't always going to be a Promise there is no need for this to override `actual` in the default message.

UziTech:
I think I got all of the changes done.

sgravrock:
Looks good. Thank you.

sgravrock:
I merged this into master, but Github can't tell because of some rebasing I did. Closing.

PR #1733, opened 2019-07-09
allow custom matchers to be async

### LINKED ISSUE #1703: Ability to add custom AsyncMatchers

UziTech:
Allow custom matchers to be asynchronous.

## Description

If a custom matcher returns a `Promise` then you can await the call to `expect`

```js
jasmine.addMatchers({
  toMatchAsync: () => {
    return {
      compare: async (actual, expected) => {
        const value = await doSomethingAsync(actual);
        const result = {
          pass: value === expected
        };
        result.message = await getAsyncMessage(expected, result.pass);
        return result;
      }
    };
  }
});

it('matches an async value', async () => {
  await expect(1).toMatchAsync(1);
});
```

## Motivation and Context

This PR allows adding asynchronous matchers like #1732 but allows `expectAsync` to keep the requirement that it is passed a `Promise` object.

Fixes #1703 
Fixes https://github.com/jasmine/jasmine/pull/1732#issuecomment-509472042

## How Has This Been Tested?

I added two tests

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/ExpectationSpec.js
+++ b/spec/core/ExpectationSpec.js
@@ -195,6 +195,79 @@ describe('Expectation', function() {
     });
   });
 
+  it('reports a passing result to the spec when the promise comparison passes', function() {
+    var matchers = {
+        toFoo: function() {
+          return {
+            compare: function() {
+              return Promise.resolve({ pass: true });
+            }
+          };
+        }
+      },
+      util = {
+        buildFailureMessage: jasmine.createSpy('buildFailureMessage')
+      },
+      addExpectationResult = jasmine.createSpy('addExpectationResult'),
+      expectation;
+
+    expectation = jasmineUnderTest.Expectation.factory({
+      customMatchers: matchers,
+      util: util,
+      actual: 'an actual',
+      addExpectationResult: addExpectationResult
+    });
+
+    return expectation.toFoo('hello')
+      .then(function () {
+        expect(addExpectationResult).toHaveBeenCalledWith(true, {
+          matcherName: 'toFoo',
+          passed: true,
+          message: '',
+          error: undefined,
+          expected: 'hello',
+          actual: 'an actual',
+          errorForStack: undefined
+        });
+      });
+  });
+
+  it('reports a failing result and a custom fail message to the spec when the promise comparison fails', function() {
+    var matchers = {
+        toFoo: function() {
+          return {
+            compare: function() {
+              return Promise.resolve({
+                pass: false,
+                message: 'I am a custom message'
+              });
+            }
+          };
+        }
+      },
+      addExpectationResult = jasmine.createSpy('addExpectationResult'),
+      expectation;
+
+    expectation = jasmineUnderTest.Expectation.factory({
+      actual: 'an actual',
+      customMatchers: matchers,
+      addExpectationResult: addExpectationResult
+    });
+
+    return expectation.toFoo('hello')
+      .then(function () {
+        expect(addExpectationResult).toHaveBeenCalledWith(false, {
+          matcherName: 'toFoo',
+          passed: false,
+          expected: 'hello',
+          actual: 'an actual',
+          message: 'I am a custom message',
+          error: undefined,
+          errorForStack: undefined
+        });
+      });
+  });
+
   it('reports a failing result with a custom fail message function to the spec when the comparison fails', function() {
     var matchers = {
         toFoo: function() {
--- a/spec/core/ExpectationSpec.js
+++ b/spec/core/ExpectationSpec.js
@@ -218,18 +218,17 @@ describe('Expectation', function() {
       addExpectationResult: addExpectationResult
     });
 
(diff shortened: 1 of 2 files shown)


sgravrock:
I think that the separate `addAsyncMatchers` in your #1732 is likely to provide a better user experience.

We considered a design like the one in this PR when we added the first built-in async matchers. We ended up deciding on the separate `expectAsync` mechanism because we anticipated that it would be easy for users to forget to await the return value of an async matcher invocation. (The vast majority of async testing problems that people report boil down to not waiting for the right things.) Having an obvious clue like the word "async" in the code that the user writes helps with that. We considered putting it in each of the async matcher names, e.g. `toBeRejectedWithAsync`, but figured that putting it "up front" would make the hint more obvious. It also helps with custom matchers that might not have the word "async" in their names.

PR #1743, opened 2019-08-22
Resolves #1740: Adds configuration option to fail specs without expectations

### LINKED ISSUE #1740: Feature Request: Ability to configure 'spec has no expectations' as error
## Feature Request
Some of our repos have hundreds upon hundreds of tests, and seeing which ones are incorrectly implemented as they have no expectations can be difficult. I understand that by default most people would want to keep the current functionality with tests with no expectations being warnings, but I would like the option to flag these as errors.

## Possible Solution
As far as I can find, and documentation on configuring jasmine is hard to find, it would be as simple as having another configuration variable in the jasmine.json, then failing tests which come across that error if the variable is true.

eg.
```
specHasNoExpectationsIsError: true
```

### Notes
I can see that this has been referenced before, but all of the relevant issues are closed, and none of them went in the direction of making this configurable. The closest I could find was #1467, which simply suggests we just fail tests with no expectations. Someone in the thread suggested making it configurable but that went nowhere.

dtychshenko:
## Description
Resolves #1740. Adds a new Jsamine configuration option `failSpecWithNoExpectations`. By default, this option is set to false and Jasmine reports specs that have no expectations as **passed**. Setting this option to `true` tells Jasmine to report specs without as expectations as **failed**.

## Motivation and Context
Currently, when a spec has no expectations, Jasmine logs a warning but still passes the test. These warnings are easy to miss, especially on a large team, and give an impression that the test is successful, when it might be not running as intended.

Having this configuration allows the developers to optionally enforce the tests to include at least one expectation in their specs providing higher visibility to the ones that do not have expectations.

## How Has This Been Tested?
- Included unit tests for all files affected and ran in node and CI
- Included an integration test for the new configuration option
- Created a sample Angular 8 application that uses Jasmine, added a spec without expectations and tried running test with (and without) failSpecWithNoExpectations set to true in a sample project

## Types of changes

## Checklist:
- [X] I have updated the documentation accordingly.


spec/core/integration/EnvSpec.js
```diff
@@ -2138,6 +2138,25 @@ describe("Env integration", function() {
       });
     });
 
+    describe('When failSpecWithNoExpectations is enabled and spec has no expectations', function() {
```
sgravrock: Could you please add a similar integration test for the case where failSpecWithNoExpectations is not enabled?

src/html/HtmlReporter.js
```diff
@@ -112,12 +112,16 @@ jasmineRequire.HtmlReporter = function(j$) {
     this.specDone = function(result) {
       stateBuilder.specDone(result);
 
-      if (
-        noExpectations(result) &&
-        typeof console !== 'undefined' &&
-        typeof console.error !== 'undefined'
-      ) {
-        console.warn("Spec '" + result.fullName + "' has no expectations.");
+      if (noExpectations(result) && typeof console !== 'undefined') {
```
sgravrock: I think we can probably do away with the checks for `console` and its methods. Jasmine no longer supports any environments that don't provide them.

dtychshenko:
@sgravrock Thanks for reviewing! I've addressed all your comments and squashed the feature into a single commit https://github.com/jasmine/jasmine/pull/1743/commits/f92d1efe827498752ebabff0fe34e9b1a27d0db9 Cheers 🍺

sgravrock:
Thanks for this. It looks like something went wrong with the squash, and the resulting history contained both the single commit and the original ones. I did a bit more rebasing to fix that and merged 7263a38c into master. Github doesn't detect that the PR is merged because of that rebasing, so I'll close it manually.

PR #1750, opened 2019-09-05
Add trusted types support

rictic:
## Description
Makes the jasmine html reporter compatible with an enforced Trusted Types policy.

This should only be necessary for a limited time, as the next version of the proposed specification will not require any special code for URLS (except javascript: URLS).

## Motivation and Context

Trusted Types is a proposed extension to CSP that would allow sites to opt into stricter checks for certain DOM APIs, where the server can specify a set of policies that are allowed to create trusted values that may be used with potentially dangerous DOM APIs.

While the it is still only a proposed standard, it is a good time for authors of major libraries and frameworks to test the Trusted Types system to determine if it matches their security needs.

This change declares a policy named "jasmine" which is then used to bless the URLs that Jasmine constructs internally. A testing environment running jasmine in a recent version of Chrome that is running with the flag `--enable-blink-features=TrustedDOMTypes` may send CSP  headers that indicate trust in the "jasmine" policy.

This code should only be necessary for a limited time, because the Trusted Types spec has recently been updated to handle anchor URLs more transparently. However I think this change will still be useful to land now, as it will enable libraries and frameworks to test with Jasmine and the current implementation that's available in Chrome today behind a flag.

More info about the proposed Trusted Types standard at https://github.com/WICG/trusted-type

## How Has This Been Tested?

I've tested this end to end in Chrome 76.0.3809.87 and 77.0.3865.19 with `--enable-blink-features=TrustedDOMTypes`, as well as with other browsers in which this change is a no-op.

## Types of changes

## Checklist:

To test this change on CI we'd need a way to send down CSP headers, and to add a Chrome with `--enable-blink-features=TrustedDOMTypes` to the browser matrix. I'm not familiar with configuring sauce or the rest of the jasmine CI infrastructure, but I'm happy to take direction!

Related change in Karma: https://github.com/karma-runner/karma/pull/3360

### DIFF
--- a/src/html/HtmlReporter.js
+++ b/src/html/HtmlReporter.js
@@ -1,4 +1,29 @@
 jasmineRequire.HtmlReporter = function(j$) {
+  // This is a no-op if not running with a Trusted Types CSP policy, and
+  // lets tests declare that they trust the way that jasmine creates URLs.
+  //
+  // More info about the proposed Trusted Types standard at
+  // https://github.com/WICG/trusted-types
+  var policy = {
+    createURL: function(s) {
+      return s;
+    }
+  };
+  var trustedTypes = window.trustedTypes || window.TrustedTypes;
+  if (trustedTypes) {
+    policy = trustedTypes.createPolicy('jasmine', policy);
+    if (!policy.createURL) {
+      // Install createURL for newer browsers. Only browsers that
+      //     implement an old version of the spec require createURL.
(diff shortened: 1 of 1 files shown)


engelsdamien:
Any updates on this change ?

rictic:
Hm, revisiting, I don't think this is necessary anymore, as the shipping version of TrustedTypes doesn't have or need the createURL function.

PR #1762, opened 2019-10-30
Implement jasmine.clock().asyncTick()

### LINKED ISSUE #1725: Proposal: await jasmine.clock().asyncTick()
Several previous issues have discussed the way mock clock doesn't work very well with Promises #1659 #1282 #710 . Here I have a proposal for a new method on the mock clock that will inter-operate with Promises nicely.

It's quite common to write code like this:
```
async function() {
  await new Promise(resolve => setTimeout(resolve, 1000));
  console.log('A');
  await new Promise(resolve => setTimeout(resolve, 1000));
  console.log('B');
}
```
You might expect `jasmine.clock().tick(2000)` to log both A and B, but it will log neither A nor B! The issue is that `.tick()` runs any scheduled timeouts one after the other, synchronously, and doesn't allow microtasks to run.

Instead, we need to run a timeout, wait for microtasks to run, and then run the remaining timeouts, repeat. This is pretty close to the way the JS engine actually works.

I propose adding an `.asyncTick()` method that does what I've just described. The test would call it like this:
```
await jasmine.clock().asyncTick(2000);
```

One difficulty is knowing how to wait for microtasks to run. In previous issues, people pointed out that various Promise implementations work in different ways. My solution to this (in other settings) has been to schedule a (real) `setTimeout`. This does slow the test down, but it's a lot better than using a real clock.

Another alternative could be to call jasmine's `clearStack()`. If a user has a Promise implementation that is slower than whatever `clearStack()` does, we could just not care. (They can always get a better Promise implementation!) This would be my preferred approach.

If this seems like a reasonable proposal, I will try to code it up and send a PR. Seems like I need to add a `runScheduledFunctionsAsync` method to `DelayedFunctionScheduler`.

stephenfarrar:
## Description

Adds an `.asyncTick()` method to `jasmine.Clock`, and to the internal `DelayedFunctionScheduler`. These are just a slight variation on the existing `.tick()` methods, but they allow microtasks to run.

There is an existing loop in `DelayedFunctionScheduler` for running the scheduled timeouts/intervals.
I just want to make that an async loop, and call clearStack() to run any pending microtasks.
Simple.

Unfortunately, it requires transforming the code.
I changed the existing loop into an async loop, using a state machine with a switch statement. (This is how babel transpiles async-await, although I did this manually.)

## Motivation and Context
Closes #1725 

This makes it possible to test code that uses timeouts in combination with Promises or async-await.

## How Has This Been Tested?

I'm fairly confident that the existing tests verify the transformed loop with the switch statement.
Thus, I've only added a handful of tests that verify microtasks get run.
* one for the initial queue pumping.
* one for the subsequent queue pumping.
* one to verify that setTimeout and asyncTick interact as expected.

Problems:
* At the moment the tests use async-await, but once the review is close to the end, I'll manually convert them from async-await to ES5.
* I have only run this on node.js. I will have to do something different in the browser, and I'm not sure how to handle IE which doesn't even have Promises.

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -933,7 +933,10 @@ getJasmineRequireObj().Env = function(j$) {
     this.clock = new j$.Clock(
       global,
       function() {
-        return new j$.DelayedFunctionScheduler();
+        return new j$.DelayedFunctionScheduler(
+          customPromise || global.Promise,
+          clearStack
+        );
       },
       new j$.MockDate(global)
     );
@@ -2956,6 +2959,27 @@ getJasmineRequireObj().Clock = function() {
       }
     };
 
+    /**
+     * Tick the Clock forward, running any pending microtasks and enqueued
+     * timeouts.
+     * @name Clock#asyncTick
+     * @since 3.5.1
+     * @function
+     * @param {int} millis The number of milliseconds to tick.
+     * @return {Promise}
+     */
+    self.asyncTick = function(millis) {
+      if (installed) {
+        return delayedFunctionScheduler.asyncTick(millis, function(millis) {
+          mockDate.tick(millis);
+        });
+      } else {
+        throw new Error(
+          'Mock clock is not installed, use jasmine.clock().install()'
+        );
+      }
+    };
+
     return self;
 
     function originalTimingFunctionsIntact() {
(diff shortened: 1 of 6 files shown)


SLaks:
Is there any update on this?
I need this functionality too.

sgravrock:
Thank you for the PR. I'm sorry you didn't get a response much sooner.

This is impressive work, but I'm not planning to merge it. The async loop transformation adds considerable complexity to what was already one of the harder parts of Jasmine to work with. That's not always a deal-breaker, but I have to weigh the maintainability cost against the value of the feature. In this case I think the cost outweighs the benefit.

`Clock` and `DelayedFunctionScheduler` are mostly decoupled from the rest of Jasmine, so those of you who want this functionality could probably pull the versions of those classes from this PR into your own codebases. I'm also going to keep this on a branch so that we can refer to it if anything changes in the future that affects the cost/benefit analysis.

Issue #1280, opened 2017-02-20
  Error: (SystemJS) Can't find variable: jasmineRequire


ahelord:
### Are you creating an issue in the correct repository?

- When in doubt, create an issue here.
- If you have an issue with the Jasmine docs, file an issue in the docs repo
  here: https://github.com/jasmine/jasmine.github.io
- This repository is for the core Jasmine framework
- If you are using a test runner that wraps Jasmine (Jasmine npm, karma, etc),
  consider filing an issue with that library if appropriate

### When submitting an issue, please answer the following:

 - What version are you using? 3.10.3
 - What environment are you running Jasmine in (node, browser, etc)? node
 - How are you running Jasmine (standalone, npm, karma, etc)? karma
 - If possible, include an example spec that demonstrates your issue.
`describe('Work Karma Jasmine FrameWork', function () {
    it('true is true', function () {
        expect('Hello World for Karma and Jasmine FrameWork').toBe('Hello World for Karma and Jasmine FrameWork');
    });
});
//# sourceMappingURL=work-karma-jasmine.spec.js.map`

When I try to run the tests with karma get 

PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  Error: (SystemJS) Can't find variable: jasmineRequire

  	eval code
  	eval@[native code]
  	F@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:12217
  	H@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:11846
  	when@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:15520
  	run@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:14559
  	_drain@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:3250
  	drain@http://localhost:9878node_modules/systemjs/dist/system-polyfills.js:4:1667
  	Evaluating http://localhost:9878node_modules/karma-jasmine/lib/boot.js
  	Error loading http://localhost:9878node_modules/karma-jasmine/lib/boot.js

i try include directorios of lib in file config karma => `// Karma configuration
// Generated on Fri Feb 17 2017 14:46:21 GMT-0500 (COT)

module.exports = function (config) {
    config.set({

        // base path that will be used to resolve all patterns (eg. files, exclude)
        basePath: '',

        // frameworks to use
        // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
[...]

PR #1766, opened 2019-11-04
Properly import jasmineRequire object before using

amilligan:
I've been working through the process of getting Jasmine up and running with a Rails 6 project using webpack.  I'm no webpack expert, but I quickly learned that webpack is very aggressive at isolating JS code in separate files.  That means spec files have to actually import/require the files containing the code under test, which then means that Jasmine really has to be compiled by webpack.  So, I create a Jasmine pack, and I ran into the dreaded 'jasmineRequire is not defined' error.

I found this referenced in #1377 and #1280 without resolution.  I haven't looked at those issues in great detail, but I believe it's clear that this problem originates in Jasmine, not the calling libraries.  The beginning of the problem lies in the definition of `jasmineRequire` in **jasmine.js**:

```javascript
  if (typeof module !== 'undefined' && module.exports && typeof exports !== 'undefined') {
    ...
    jasmineRequire = exports;
  } else {
    ...
    jasmineRequire = jasmineGlobal.jasmineRequire = {};
  }
```
In old, asset pipeline world this just tacks `jasmineRequire` onto the global object (window), and other files access it all higgledy piggledy from there; no worries.  But, in new, webpack world, this assigns `jasmineRequire` as the export from the file, available to other files that _require_ this file.  At pack compile time the global objects assigned to `window` don't appear to be available to other files.

The files **jasmine-html.js** and **boot.js** both reference `jasmineRequire` without requiring the **jasmine.js** file which defines it, assuming it exists in the global scope.  Hence the missing reference, the failure to load, and general worldwide unhappiness.  The solution, as presented here, is simply to have those two files require **jasmine.js** in the case that `window.jasmineRequire` is not available.

I've made this change locally and Jasmine loads happily.  Peace and prosperity return to all the land.

@ragaskar 
@infews 
@slackersoft


ragaskar:
Hi Adam: I don't really do any Jasmine maintaining anymore (you def. want @slackersoft ) -- but thought I'd say hi! 

Hope you're well, and thanks for the PR. :)

amilligan:
@ragaskar Hey Rajan, I figured as much.  But, your face is still indelibly imprinting on my memory when I think of Jasmine.  And it has been a long time, so I figured I'd take the opportunity to say hi as well.  If you find yourself back in New York at any point you should let me buy you a drink.

amilligan:
I added one more small commit, with related changes.  The **boot.js** file was extending all of the Jasmine methods into `window`, which, as mentioned, doesn't work with Webpack.  This simply uses the `getGlobal` method that the Jasmine core object provides to get the appropriate global object for the given environment and extend the methods into that.  In non-Webpack scenarios the `getGlobal` method should return `window`, so nothing should change.

I think there's something to be said for not polluting the global namespace, but providing a top-level object to kick off each spec block.  RSpec made this change a while ago, and other than the necessity of adding `RSpec.describe` at the top level it didn't cause much disruption.  But, that's beyond the scope of this change; something to consider in the future.

Incidentally, these changes are in service of working through https://github.com/jasmine/jasmine-gem/issues/311.

@slackersoft

slackersoft:
Thanks for taking a crack at this. I've looked into mechanisms for not polluting the global namespace, but I haven't had time to really sit down and figure out how to properly `export` various Jasmine top level concepts without breaking the existing non-webpack way to write your tests. We don't get anything as easy as `Rspec.describe` with just `it` inside because JavaScript doesn't have allow calling member functions without `this`. I want to try running this against some browsers here too before I merge it in.

amilligan:
@slackersoft Sure, of course, the idea about not polluting the global namespace was just an aside. I struggled with the same thing in Objective C.  I don't recall finding a particularly satisfying solution for Cedar, so I can't imagine it will be trivial.  Perhaps it's not even possible. 
 Nothing's ever as easy as it is in Ruby, sadly.

stoivo:
I have tested and this seem good

sgravrock:
Thanks, @amilligan . I merged this in d04b813 but Github isn't showing the PR as merged.

PR #1776, opened 2020-01-07
Fix #1771 : Method messagePrefixLength now considers other error messages like ReferenceError or TypeError, too

### LINKED ISSUE #1771: Extracting error message in extractMessage() only works if stack trace starts with "Error .."
In extractMessage() the error message cannot be extracted if the stack trace starts with something else then "Error". For example the message for "TypeError" won't be extracted because function messagePrefixLength() in extractMessage() returns 0, because there is no match for "match(/^Error/)".
 
https://github.com/jasmine/jasmine/blob/ef3f127d270ccbbd6dde3fe161e0e85474f19369/lib/jasmine-core/jasmine.js#L7789 
In the end that leads to an issue in function filterJasmine(). 

## Expected Behavior
messagePrefixLength() should consider error messages like "TypeError" or "ReferenceError", too

## Current Behavior
messagePrefixLength() just considers error messages which start with "Error" 

## Possible Solution
In https://github.com/jasmine/jasmine/blob/ef3f127d270ccbbd6dde3fe161e0e85474f19369/lib/jasmine-core/jasmine.js#L7789 change messagePrefixLength()
`if (!stackLines[0].match(/^Error/))` 
to 
`if (!stackLines[0].match(/^\w*Error/))`

vhermannitk:
## Description

In extractMessage() the error message cannot be extracted if the stack trace starts with something else then "Error". For example the message for "TypeError" won't be extracted because function messagePrefixLength() in extractMessage() returns 0, because there is no match for "match(/^Error/)".

## Motivation and Context
messagePrefixLength() just considers error messages which start with "Error". It should also consider error messages like "TypeError" or "ReferenceError".

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2008-2019 Pivotal Labs
+Copyright (c) 2008-2020 Pivotal Labs
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
@@ -2236,12 +2236,11 @@ getJasmineRequireObj().JsApiReporter = function(j$) {
 };
 
 getJasmineRequireObj().Any = function(j$) {
-
   function Any(expectedObject) {
     if (typeof expectedObject === 'undefined') {
       throw new TypeError(
         'jasmine.any() expects to be passed a constructor function. ' +
-        'Please pass one or use jasmine.anything() to match any object.'
+          'Please pass one or use jasmine.anything() to match any object.'
       );
     }
     this.expectedObject = expectedObject;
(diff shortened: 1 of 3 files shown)


sgravrock:
Thanks for the PR. The approach you're taking seems reasonable. However, there are a couple of problems that prevent this from being merged as-is. You've modified the build product rather than the source code (`src/core/StackTrace.js` in this case), and there are no tests for the functionality that you've added. Please review the [contributor's guide](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md) and update accordingly.

vhermannitk:
@sgravrock I modified the source code instead of the build product and added some tests with my last commit. Please tell me if it is not sufficient. Sorry for the inconvenience !

Issue #1760, opened 2019-10-25
spyOnAllFunctions spies on constructors


elliot-nelson:
Currently, using `spyOnAllFunctions` on a module that has a mix of functions and classes breaks the class constructors.

## Example Behavior

```
class RetryableError extends Error {
    retry() { }
}

const API = {
    RetryableError,
    makeRequest: () => {}
};

describe('API', function () {
    it('works', function () {
        spyOnAllFunctions(API);

        new API.RetryableError().retry();
        // -> Error: "retry" is not a function (the new returns {})

        API.RetryableError.and.callThrough();

        new API.RetryableError().retry();
        // -> Error: Class constructor RetryableError cannot be invoked without 'new'
    });
});
```

## Possible Solution
It seems like the best solution is to ignore constructors, because once you spy on one, you can't "fix it" with `.and.callThrough()` (and in general, spying on constructors is an anti-pattern).

## Your Environment
* node 8.x, 10.x
* jasmine 3.4.0/3.5.0

sgravrock:
The problem is, how does Jasmine determine whether a given function is a constructor or not? A constructor is simply a regular (non-arrow, non-shorthand-member) function that gets called with the `new` keyword. As far as I know there's not any difference identifiable at runtime between a function that we think of as a constructor and one that we don't. 

I'd be happy to be proven wrong on this. But I can't help noticing that the various `isConstructor` functions that I've been able to find all define "constructor" to mean one of the following:

* A function that isn't an arrow function or a shorthand member
* A function that doesn't throw when you call it with `new` and no arguments

I doubt that either of those definitions is what you, or anyone else who isn't in the middle of writing an `isConstructor` function, have in mind.

slackersoft:
I think I would rather work to fix this type of issue by allowing `spyOn(thing, 'SomeConstructor').and.callThrough()` to do the right thing and invoke the original constructor with `new`. This would mean detecting use of the `new` keyword when the spy is invoked and passing that down to the `SpyStrategy` to do the actual invocation correctly.

I would be happy to review a pull request to add `new` detection to spy invocations.

PR #1782, opened 2020-01-29
Allow callThrough to call constructor functions without errors

elliot-nelson:
## Description
This is an initial stab at fixing https://github.com/jasmine/jasmine/issues/1760.

The idea is that if you spy on a constructor function and `callThrough`, and then construct that object, it should work as expected without errors.  Unfortunately, this ends up requiring quite a bit of plumbing changes, as there's not really any way to pass the information that this is a `new` call all the way from the end caller down to the eventual spy strategy using the current implementation.

I'm not 100% certain this functionality is worth this much churn, there may be a safer way to do this I'm not seeing yet.

## Motivation and Context
See https://github.com/jasmine/jasmine/issues/1760 -- this is hopefully to fix overreach by spyOnAllFunctions.

## How Has This Been Tested?
Ran basic tests, added an integration test confirming it works on common patterns (used by node packages like `aws-sdk`, etc.)

## Types of changes

## Checklist:

### DIFF
--- a/src/core/Spy.js
+++ b/src/core/Spy.js
@@ -20,8 +20,8 @@ getJasmineRequireObj().Spy = function(j$) {
     getPromise
   ) {
     var numArgs = typeof originalFn === 'function' ? originalFn.length : 0,
-      wrapper = makeFunc(numArgs, function(context, args, isConstructor) {
-        return spy(context, args, isConstructor);
+      wrapper = makeFunc(numArgs, function(context, args, invokeNew) {
+        return spy(context, args, invokeNew);
       }),
       strategyDispatcher = new SpyStrategyDispatcher({
         name: name,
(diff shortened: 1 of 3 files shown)


elliot-nelson:
Also, one thought: if this change is accepted, we should probably rename all of my `isConstructor` booleans to `isNew` to make it clear that we aren't tracking whether or not the target function "is a constructor", we're just tracking whether or not the caller used the `new` keyword on this particular call to the function.

src/core/Spy.js
```diff
@@ -33,21 +33,21 @@ getJasmineRequireObj().Spy = function(j$) {
...
-          object: this,
+          object: context,
           invocationOrder: nextOrder(),
-          args: Array.prototype.slice.apply(arguments)
+          args: Array.prototype.slice.apply(args)
         };
 
         callTracker.track(callData);
-        var returnValue = strategyDispatcher.exec(this, arguments);
+        var returnValue = strategyDispatcher.exec(this, args, isConstructor);
```
slackersoft: Even if it is technically correct, the `this` should probably now be `context` to be consistent with which object is supposed to be passed around.

elliot-nelson:
Changes:
 - Use context instead of this.
 - Settled on `invokeNew` as the parameter name for the new plumbing (shorter, active voice).
 - Minor other naming changes, re-run prettier, etc.

PR #1796, opened 2020-02-28
toHaveSize

wokier:
toHaveSize
 
## Description
toHaveSize, new matcher for array, collections, object

## Motivation and Context
simplify the assertions:
assert(array.length).toEqual(2)
becomes
assert(array).toHaveSize(2)

## How Has This Been Tested?
unit tested

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -144,7 +144,6 @@ getJasmineRequireObj().requireMatchers = function(jRequire, j$) {
       'toBeUndefined',
       'toContain',
       'toEqual',
-      'toHaveSize',
       'toHaveBeenCalled',
       'toHaveBeenCalledBefore',
       'toHaveBeenCalledTimes',
@@ -291,9 +290,9 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return typeof jasmineGlobal.Node !== 'undefined'
       ? obj instanceof jasmineGlobal.Node
       : obj !== null &&
-          typeof obj === 'object' &&
-          typeof obj.nodeType === 'number' &&
-          typeof obj.nodeName === 'string';
+      typeof obj === 'object' &&
+      typeof obj.nodeType === 'number' &&
+      typeof obj.nodeName === 'string';
     // return obj.nodeType > 0;
   };
 
@@ -764,7 +763,7 @@ getJasmineRequireObj().Spec = function(j$) {
       onComplete: function() {
         onComplete(
           self.result.status === 'failed' &&
-            new j$.StopExecutionError('spec failed')
+          new j$.StopExecutionError('spec failed')
         );
       },
       userContext: this.userContext()
@@ -829,8 +828,8 @@ getJasmineRequireObj().Spec = function(j$) {
       this.result.failedExpectations.length > 0 ||
       (failSpecWithNoExpectations &&
         this.result.failedExpectations.length +
-          this.result.passedExpectations.length ===
-          0)
+        this.result.passedExpectations.length ===
+        0)
     ) {
       return 'failed';
     }
@@ -1171,7 +1170,7 @@ getJasmineRequireObj().Env = function(j$) {
       }
       runnableResources[
         currentRunnable().id
-      ].defaultStrategyFn = defaultStrategyFn;
+        ].defaultStrategyFn = defaultStrategyFn;
     };
 
     this.addSpyStrategy = function(name, fn) {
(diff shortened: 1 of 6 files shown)


wokier:
Does the documentation update have to be done in
https://github.com/jasmine/jasmine.github.io ?

spec/core/matchers/toHaveSizeSpec.js
```diff
@@ -0,0 +1,21 @@
+describe('toHaveSize', function() {
+  'use strict';
+
+  it('delegates to equals function', function() {
```
slackersoft: Given that we expect the size/length to be an actual number all of the time, what is the reason to use the deep equals functionality in matchersUtil?

src/core/matchers/toHaveSize.js
```diff
@@ -0,0 +1,57 @@
...
+          diffBuilder = j$.DiffBuilder({prettyPrinter: matchersUtil.pp});
+
+        // Avoid misleading collections size matching
+        if (actual instanceof WeakSet
+          || actual instanceof WeakMap
+          || actual instanceof DataView)
+          throw new Error('Cannot get size of ' + actual + '.');
+
+        // Ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects
+        if (Array.isArray(actual))
```
slackersoft: If we're going to call out this many special cases, I'd like to see specific checks for each of them in the test. If that seems excessive, I'd be happy to see a more lenient check (e.g. does the `actual` have a `length` or `size`) and then just add some tests for those features.

src/core/matchers/toHaveSize.js
```diff
@@ -0,0 +1,57 @@
...
+   */
+  function toHaveSize(matchersUtil) {
+    return {
+      compare: function(actual, expected) {
+        var result = {
+            pass: false
+          },
+          diffBuilder = j$.DiffBuilder({prettyPrinter: matchersUtil.pp});
+
+        // Avoid misleading collections size matching
```
slackersoft: I appreciate including the better error messages, but I'd like to see tests for this in the `toHaveSizeSpec`

lib/jasmine-core/jasmine.js
```diff
@@ -144,6 +144,7 @@ getJasmineRequireObj().requireMatchers = function(jRequire, j$) {
       'toBeUndefined',
       'toContain',
       'toEqual',
+      'toHaveSize',
```
slackersoft: Please remove the generated `jasmine.js` file [as mentioned in the contributing guide](https://github.com/jasmine/jasmine/blob/master/.github/CONTRIBUTING.md)  from the PR so we don't have to deal manage a number of conflicts in this file.

spec/core/matchers/toHaveSizeSpec.js
wokier: I just took exemple from toEqual.
What should i use for simple comparison ?

src/core/matchers/toHaveSize.js
wokier: Some js framework uses length property to detects arrays, but this is quite dangerous.
I could use the lodash way, which provides some additionnal checks
https://lodash.com/docs/4.17.15#isArrayLike
I agree that would simplify.

lib/jasmine-core/jasmine.js
wokier: ok

spec/core/matchers/toHaveSizeSpec.js
slackersoft: For a simple comparison of two numbers, it's probably fine to just use `===`

src/core/matchers/toHaveSize.js
slackersoft: Since we're also having this work for strings, it might actually be fine just to check for a `length` property that is a `number` and then some specific checks for `Map` and `Set`

wokier: see abe38429f5ee4e40727b8a2c22e36ebc9bf5b10e

src/core/matchers/toHaveSize.js
wokier: see ae340da2419ac960f16fd2ec1034f3dfe03e487a

spec/core/matchers/toHaveSizeSpec.js
wokier: see 442f2cb6d4447ab5b022caf4e5bece9f9be9aec3

wokier: and 0d9e36080b25721fe36a134064a38aa66538b939

lib/jasmine-core/jasmine.js
wokier: see 0dfdd2d8a896929b690a2eaa47a7a8369b56d6f6

wokier:
Ready for a second review

slackersoft:
Merged with 7e940542f89930817771f8b2497b2446c507c895

PR #1801, opened 2020-03-13
Add a toHaveBeenCalledOnceWith matcher

### LINKED ISSUE #1717: Feature suggestion: Add toHaveBeenCalledOnceWith matcher
It would be extremely useful to have a `toHaveBeenCalledOnceWith()` matcher combining both `toHaveBeenCalledTimes(1)` and `toHaveBeenCalledWith()` matchers.

About 70% of my unit test expectations about functions look like this:
```typescript
expect(backendService.get).toHaveBeenCalledTimes(1);
expect(backendService.get).toHaveBeenCalledWith('SomeArgument', someSecondArgument);
```

It would be really much, much more readable if we could have a single matcher that combines both matchers:
```typescript
expect(backendService.get).toHaveBeenCalledOnceWith('SomeArgument', someSecondArgument);
```

Maximaximum:
This PR adds an toHaveBeenCalledOnceWith, which passes only if the spy has been called once, and the parameters of the call equal to the expected ones. Effectively, it combines `toHaveBeenCalledTimes(1)` and `toHaveBeenCalledWith()` matchers.

Fixes #1717

## Types of changes

## Checklist:


elliot-nelson:
I too would use this version over `toHaveBeenCalledWith` for a lot of tests, but I think incorporating the `DiffBuilder` (the way the above matcher does) would improve it a lot.

I use the enhanced diff output of `toHaveBeenCalledWith` constantly -- it's amazing for calls between functions with many options, for example.  I'd like the extra "and only once" clause, but I don't think I could give up the enhanced diff output for it.

Maximaximum:
@elliot-nelson do you mean that you suggest that this PR should be enhanced somehow? Could you please provide an example?

elliot-nelson:
@Maximaximum yes, if you look at the `toHaveBeenCalledWith` matcher, it includes this section:

```js
    var prettyPrintedCalls = actual.calls.allArgs().map(function(argsForCall) {
      return '  ' + matchersUtil.pp(argsForCall);
    });

    var diffs = actual.calls.allArgs().map(function(argsForCall, callIx) {
    var diffBuilder = new j$.DiffBuilder();
      matchersUtil.equals(argsForCall, expectedArgs, diffBuilder);
      return 'Call ' + callIx + ':\n' +
        diffBuilder.getMessage().replace(/^/mg, '  ');
    });

    return 'Expected spy ' + actual.and.identity + ' to have been called with:\n' +
      '  ' + matchersUtil.pp(expectedArgs) + '\n' + '' +
      'but actual calls were:\n' +
      prettyPrintedCalls.join(',\n') + '.\n\n' +
      diffs.join('\n');
  };
```

In this case when it says `but actual calls were:`, it not only prints out the calls that were made, but a detailed diff between the expected call and the actual call.  Your new matcher is based on `toHaveBeenCalledWith` but doesn't include that detailed diff part, and I think it would be even better if it included it.

Maximaximum:
@elliot-nelson thanks for your feedback!

I've updated the matcher to print the details diff between the actual call and the expected call. 

Please note however that the diff is only printed if there has been only 1 call, and the arguments were different from the expected. If there have been 0 calls or multiple calls, the diff does not make much sense.

elliot-nelson:
Nice! The new code and test looks good to me FWIW.  🙇 

(Looks like there might be a minor fix for unit tests required.)

Maximaximum:
Hm, I have no idea why one test fails :( On my machine it passes successfully

elliot-nelson:
It looks to me like you have one test that is constructing a fake "utils" object - the one for the custom equality testers - that would fail.  Rather than create an empty object from scratch, you probably need to do what the toBeCalledWith spec does and create an instance of MatchersUtil and override the appropriate function.

Maximaximum:
I wonder why is this PR still not merged in? Would love to hear any feedback from the team

Maximaximum:
@slackersoft you said that you'd be happy to review this PR, could you please do that?

sgravrock:
This looks good. I've merged it. Thanks.

Note that the matcher interface changed recently. I made some changes in b3ab9fad to update this matcher for the new interface.

Maximaximum:
@sgravrock Thanks a lot!

Do you plan to make a release soon?

I guess that documentation also has to be updated to add info about the new matcher?

sgravrock:
You're welcome. I've just regenerated the docs. You can see your new matcher at <https://jasmine.github.io/api/edge/matchers.html#toHaveBeenCalledOnceWith>.

I hope to get 3.6 out soon. I wish I could give a better estimate than that but these days it's hard to predict when we'll have the solid block of time that it takes to put out a release.

sgravrock:
We just released 3.6. Enjoy.

PR #1822, opened 2020-06-02
allow spy throwError to throw an Object

terencehonles:
## Description

When using the following code to simulate a node error:

```javascript
spyOn(process, 'kill').and.throwError({code: 'ESRCH'})
```

The object passed in will be converted to a string by the Error constructor and result in `'[object Object]'` which is not very useful. This PR changes the ``throwError`` spy strategy to only convert
strings into an Error object, but any other objects which are passed in will be thrown as is. This means the spy strategy can never emulate throwing a bare string ``throw 'error'``, but this would be a backward incompatible change.

## Motivation and Context

To more easily throw a node like object / allow `throwError` to match `throw expression` better.

## How Has This Been Tested?

This is change just reverses an if/else check and checks for a string instead of an Error object so it's a constrained change.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/SpyStrategy.js
+++ b/src/core/SpyStrategy.js
@@ -150,10 +150,10 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
    * @name SpyStrategy#throwError
    * @since 2.0.0
    * @function
-   * @param {Error|String} something Thing to throw
+   * @param {Error|Object|String} something Thing to throw
    */
   SpyStrategy.prototype.throwError = function(something) {
-    var error = something instanceof Error ? something : new Error(something);
+    var error = j$.isString_(something) ? new Error(something) : something;
     this.plan = function() {
       throw error;
     };
--- a/spec/core/SpyStrategySpec.js
+++ b/spec/core/SpyStrategySpec.js
@@ -70,7 +70,7 @@ describe('SpyStrategy', function() {
     expect(originalFn).not.toHaveBeenCalled();
   });
 
-  it('allows a non-Error to be thrown, wrapping it into an exception when executed', function() {
+  it('allows a string to be thrown, wrapping it into an exception when executed', function() {
     var originalFn = jasmine.createSpy('original'),
       spyStrategy = new jasmineUnderTest.SpyStrategy({ fn: originalFn });
 
@@ -79,6 +79,18 @@ describe('SpyStrategy', function() {
     expect(function() {
       spyStrategy.exec();
     }).toThrowError(Error, 'bar');
+    expect(originalFn).not.toHaveBeenCalled();
+  });
+
+  it('allows a non-Error to be thrown when executed', function() {
+    var originalFn = jasmine.createSpy('original'),
+      spyStrategy = new jasmineUnderTest.SpyStrategy({ fn: originalFn });
+
+    spyStrategy.throwError({ code: 'ESRCH' });
+
+    expect(function() {
+      spyStrategy.exec();
+    }).toThrow({ code: 'ESRCH' });
     expect(originalFn).not.toHaveBeenCalled();
   });
 


elliot-nelson:
I'm torn on this change because on the one hand, I don't think jasmine should be super prescriptive about what you do with it... On the other hand, I wouldn't want to encourage throwing raw objects like this around, because this kind of code rarely lives in a vacuum.  If the code under test is going to bother looking at `error.code`, it's more likely than likely also going to attempt to print `error.message` and/or `error.stack`.

I'd recommend this:

```js
spyOn(process, 'kill').and.throwError(Object.assign(new Error('Process Not Found'), { code: 'ESRCH' }));
```

Or if you have spec that throws a whole bunch of these errors, a tiny helper:

```js
class ErrorCode extends Error {
    constructor(code) {
        super(code);
        this.code = code;
    }
}

spyOn(process, 'kill').and.throwError(new ErrorCode('ESRCH'));
```

src/core/SpyStrategy.js
```diff
@@ -150,10 +150,13 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
    * @name SpyStrategy#throwError
    * @since 2.0.0
    * @function
-   * @param {Error|String} something Thing to throw
+   * @param {Error|Object|String} something Thing to throw
    */
   SpyStrategy.prototype.throwError = function(something) {
-    var error = something instanceof Error ? something : new Error(something);
+    var error =
+      something instanceof String || typeof something === 'string'
```
sgravrock, line 157: Please use j$.isString_ to do the type check.

src/core/SpyStrategy.js
```diff
@@ -150,10 +150,13 @@ getJasmineRequireObj().SpyStrategy = function(j$) {
...
+        ? new Error(something)
+        : something;
```
sgravrock, line 159: Please add a new test for this and update/rename the test for the old behavior as appropriate.

sgravrock:
It's a bit of a judgment call, but I think this falls on the side of *allowing* users to test code that might not be ideal rather than *encouraging* them to write such code. Ideally we'd never have added the conversion to `Error`, but we did, and on balance I think it's better to have this change than to leave things as they are.

terencehonles:
> I'm torn on this change because on the one hand, I don't think jasmine should be super prescriptive about what you do with it... On the other hand, I wouldn't want to encourage throwing raw objects like this around, because this kind of code rarely lives in a vacuum. If the code under test is going to bother looking at `error.code`, it's more than likely also going to attempt to print `error.message` and/or `error.stack`.
> 
> I'd recommend this:
> 
> ```js
> spyOn(process, 'kill').and.throwError(Object.assign(new Error('Process Not Found'), { code: 'ESRCH' }));
> ```
> 
> Or if you have spec that throws a whole bunch of these errors, a tiny helper:
> 
> ```js
> class ErrorCode extends Error {
>     constructor(code) {
>         super(code);
>         this.code = code;
>     }
> }
> 
> spyOn(process, 'kill').and.throwError(new ErrorCode('ESRCH'));
> ```

I understand your concern, and it's something we discussed on https://github.com/angular/angular/pull/37250 but I was just trying to help contribute and I was copying/simplifying the case where they were throwing a simple object and *not* looking at the error message and I was confused why something that I expected to just work didn't. If this was not a desired change an error complaining that the argument was not a string or error type would be better than coercing the object into a string `"[object Object]"`.

> It's a bit of a judgment call, but I think this falls on the side of _allowing_ users to test code that might not be ideal rather than _encouraging_ them to write such code. Ideally we'd never have added the conversion to `Error`, but we did, and on balance I think it's better to have this change than to leave things as they are.

Agreed, and why I submitted this PR.

src/core/SpyStrategy.js
terencehonles: Done, thanks for the suggestion

src/core/SpyStrategy.js
terencehonles: Done, was just waiting to see if this was a desired change before adding a test.

PR #1834, opened 2020-07-19
fixes problems that occur when jasmine.clock.install() is in action

iliakan:
I found out that Jasmine is terribly slow for me.

After hours of debugging, the reason seems to be the typo.

The PR is quite obvious, as this really looks like a typo: `setTimeout` instead of `global.setTimeout`.

So when the "default" `setTimeout" is mocked, then things go wrong. Do I get it right?

The PR patches it. 

P.S. You don't want to use `queueMicrotask` here, right? Using `setTimeout/setImmediate` on purpose, to let things render etc?


iliakan:
Wrong patch. Ignore plz.

PR #1851, opened 2020-08-31
Don't try to handle number as DOM node

paulheising:
## Description

If a `number` comes in, handle it like a `string`, so jasmine won't try to handle it like a `Node` and fail.

## Motivation and Context

I was running into problems when using [karma-jasmine-order-reporter](https://www.npmjs.com/package/karma-jasmine-order-reporter). I was getting errors like this:

```
Chrome Headless 85.0.4183.83 (Linux x86_64) ERROR
  An error was thrown in afterAll
  TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
      at <Jasmine>
      at createDom (/home/paul.heising/dev/pwa/node_modules/karma-jasmine-html-reporter/src/lib/html.jasmine.reporter.js:660:16)
      at HtmlReporter.jasmineDone (/home/paul.heising/dev/pwa/node_modules/karma-jasmine-html-reporter/src/lib/html.jasmine.reporter.js:257:11)
```

Even though this problem only occurs with karma-jasmine-order-reporter, I think the change is a quite obvious fix for the case that a `number` comes in here, because of course it causes an error if you try to handle a `number` like a `Node`.

## How Has This Been Tested?

I did some local testing, by modifying the corresponding file inside the `node_modules` of my project that depends on jasmine. Bit since the change is small and obvious in my eyes, I did not test quite extensively.

Environment:

- Ubuntu 20.04
- Chrome 85
- Node 12.18.3

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine-html.js
+++ b/lib/jasmine-core/jasmine-html.js
@@ -655,7 +655,7 @@ jasmineRequire.HtmlReporter = function(j$) {
       for (var i = 2; i < arguments.length; i++) {
         var child = arguments[i];
 
-        if (typeof child === 'string') {
+        if (typeof child === 'string' || typeof child === 'number') {
           el.appendChild(createTextNode(child));
         } else {
           if (child) {
--- a/lib/jasmine-core/jasmine-html.js
+++ b/lib/jasmine-core/jasmine-html.js
@@ -656,7 +656,7 @@ jasmineRequire.HtmlReporter = function(j$) {
         var child = arguments[i];
 
         if (typeof child === 'string' || typeof child === 'number') {
-          el.appendChild(createTextNode(child));
+          el.appendChild(createTextNode(child.toString()));
         } else {
           if (child) {
             el.appendChild(child);


slackersoft:
Please take a look again at the [CONTRIBUTING guide](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md), specifically what to do [before submitting a pull request](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md#before-committing-or-submitting-a-pull-request). Specifically:

* You have added a generated file to your pull request (this generally leads to merge conflicts that we then have to resolve).
* You didn't make changes to the source files, which means your changes will vanish the next time we generate the compiled files
* You have also not added any tests for this functionality. Since Jasmine is _itself_ a testing framework, we believe new code should have tests.


On top of all of this, I'm not sure what is going on with some of those libraries with Karma, but that `createDom` method is not part of Jasmine's external interface, and should only be called by inside Jasmine with arguments that it expects, so I don't think this is something we really need to accommodate in Jasmine. It sounds like the karma-jasmine-order-reporter is causing the issue.

paulheising:
> It sounds like the karma-jasmine-order-reporter is causing the issue.

Possibly. For now, we have been working around the issue by activating the order-reporter only if chrome runs in headless mode - no more issues since then. I will keep an eye on that, but for now, I think I can close the PR.

PR #1860, opened 2020-10-19
GlobalErrors: use global.onError in browserRejectionHandler so can overwrite

greghuc:
## Description
A recent [change](https://github.com/jasmine/jasmine/pull/1778) started forwarding unhandled promise errors to the jasmine onerror handler, via an "unhandledrejection" listener. This PR tweaks this code to call the jasmine onerror function via `global.onerror`, not the `onerror` local variable.
 
## Motivation and Context
When testing global error handling in our own code, it's sometimes useful to disable jasmine's global error handling temporarily, to prevent clashes. This is doable by replacing (jasmine's) `global.onerror` with our own function: `global.onerror = ownOnError`. However, this approach fails with the recent unhandledrejection [change](https://github.com/jasmine/jasmine/pull/1778), as this new logic calls jasmine's error function directly via the `onerror` local variable, rather than `global.onerror`.

This PR  tweaks this unhandledrejection code to call the jasmine onerror function via `global.onerror`, so enabling the onerror overriding approach to work again. It also ensures that this overriding approach behaves consistently for errors triggered from both promise and non-promise situations.

## How Has This Been Tested?
Ran the jasmine test suite.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/GlobalErrorsSpec.js
+++ b/spec/core/GlobalErrorsSpec.js
@@ -285,5 +285,64 @@ describe('GlobalErrors', function() {
         })
       );
     });
+
+    describe('Enabling external interception of reported rejections by overriding global.onerror', function() {
+      it('overriding global.onerror intercepts rejections whose reason is a string', function() {
+        var fakeGlobal = jasmine.createSpyObj('globalErrors', [
+            'addEventListener'
+          ]),
+          handler = jasmine.createSpy('errorHandler'),
+          hijackHandler = jasmine.createSpy('hijackErrorHandler'),
+          errors = new jasmineUnderTest.GlobalErrors(fakeGlobal);
+
+        errors.install();
+        errors.pushListener(handler);
+
+        fakeGlobal.onerror = hijackHandler;
+
+        var addedListener = fakeGlobal.addEventListener.calls.argsFor(0)[1];
+        addedListener({ reason: 'nope' });
+
+        expect(hijackHandler).toHaveBeenCalledWith(
+          'Unhandled promise rejection: nope'
+        );
+        expect(handler).not.toHaveBeenCalled();
+      });
+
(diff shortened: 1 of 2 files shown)


sgravrock:
In the long term I want to come up with something a bit more full-featured, but this seems like a good stopgap approach to handle situations like yours. I'd like to see some test coverage, though. Preferably something that demonstrates the expected usage and verifies that it works. As it is, I'm concerned that this change could easily be refactored away because it's not obvious that using `global.onerror` instead of `onerrro is load-bearing.

greghuc:
@sgravrock ok I added some tests to support the changes to global.error for promises. First, I added two tests for the new behaviour, which show that overriding global.onerror lets you do custom non-jasmine global error handling of promise errors. The tests go from red to green with the new code changes. Second, I added a single test showing that overriding global.onerror lets you do custom non-jasmine global error handling for __non-promise__ errors. This just exercises existing jasmine behaviour - no code changes required - but ensures that we're covered if this behaviour changes in future.

Let me know if this is ok, or you would prefer extra changes.

greghuc:
Hi @sgravrock any progress on this GlobalErrors pr? My original code changes now have supporting tests as requested, so tell me if there's anything else you need.

sgravrock:
Thank you. I merged this in 1320b06 but Github doesn't show the PR as merged because I rebased the commits.

greghuc:
Great - thanks for merging :-)

tanov:
Correct me if I am wrong, but in case of receiving an object, the string I get is 'Unhandled promise rejection: [object Object]'
global.onerror('Unhandled promise rejection: ' + event.reason);
Isn't it better to change this line to:
global.onerror('Unhandled promise rejection: ' + event.reason.error.message);

Issue #491, opened 2014-01-02
Feature Request: Sourcemap support


nicosommi:
It would be nice if the original line number of the error is displayed when developing in coffeescript.

For example for this error:
TypeError: entityCollection.get(...) is undefined in http://somehost:[someport]/somepath/SomeSpec.js (line 37)

It can be something like this (reading the sourcemap):
TypeError: entityCollection.get(...) is undefined in http://somehost:[someport]/somepath/Spec.coffee (line 15) (and line 37 on /somepath/SomeSpec.js)

I think this can be done reading the //@sourceMappingURL or the //#sourceMappingURL attribute of the compiled js and later the sourcemap itself (using the value of that attribute), then there is the sourcemap interpretation using https://github.com/mozilla/source-map

alexeagle:
+1 this would be useful for running unit tests written in Typescript

caseyhoward:
:+1: I compile all my tests beforehand using typescript. Source map support would be really nice. I hate digging through the generated file.

geddski:
+1 as someone using webpack, I have no idea where the actual errors are without source map support.

geddski:
For anyone else needing this, I've found that adding `?catch=false` to the URL at least lets the errors make it to the console so the browser can use source maps to show you where the errors are.

cmdelatorre:
+1 to support tests in coffeescript
In the meanwhile, @novocaine , do you have some tips on how to integrate your module in my project without patching Jasmine?

novocaine:
sorry, you definitely have to patch Jasmine..

inf3rno:
@geddski I tried out, but it does not work, because the error was rethrown by jasmine, so it points to the code of jasmine instead of my source code. (Btw. did anybody manage to use sourcemaps in firefox and chrome. I got proper support only in opera and msie.)

If somebody needs this feature, it works with karma + browserify + jasmine (tested with phantomjs). Here is a project template: https://github.com/inf3rno/javascript-project-template

guncha:
If anyone else is looking for a quick hack to get the source mapped stack traces working with Jasmine's HTML reporter while this is being figured out, here's a gist for you: https://gist.github.com/guncha/f45ceef6d483c384290a

DanielSchuech:
thx @guncha 

I am using it with karma jasmine html reporter and its working well! I haven't found another solution so I just can recommand that atm.

guncha:
[...]

PR #1862, opened 2020-11-02
Support source maps

JannesMeyer:
## Description
Node.js >=12.12.0 has support for source maps when run with the flag [`--enable-source-maps`](https://nodejs.org/dist/latest/docs/api/cli.html#cli_enable_source_maps) but the additional stack trace information is not being shown by Jasmine's stack traces. This pull request stops stack trace information from being discarded. Among other things, this causes the source map information to show up.

## Motivation and Context
Currently Jasmine discards source map information from stack traces when the `file` property is not set on those stack frames.
This would proably fix https://github.com/jasmine/jasmine/issues/491.

It might be relevant that the "competing" framework Mocha has specific support for `--enable-source-maps`: https://mochajs.org/#-enable-source-maps

Personally, I would prefer a completely unmodified stack trace from Node/V8, because I think that most users are educated enough to be able to read those. The extra information is just too valuable.

## How Has This Been Tested?

I have manually tested this with a test suite that throws an error. Before running Jasmine I had to set the following environment variable:

```powershell
# Setting environment variables might look different in your shell. This is what I ran in PowerShell:
$env:NODE_OPTIONS = '--enable-source-maps'
```

 The stack trace looks much more complete now and it finally includes the source map information. I am not sure if it's feasible to add an automated test, because the feature requires at least Node.js v12.12.0, so the test would fail on Node.js 10. Additionally, we would have to `--enable-source-maps` for the test process.

Output before:

```
  Stack:
        at <Jasmine>
        at UserContext.<anonymous> (C:\test\spec\spec.js:3:15)
        at <Jasmine>
        at processImmediate (node:internal/timers:462:21)
```

Output after:

```
  Stack:
    C:\test\spec\spec.ts:9
            throw new Error('This is a test');
                  ^
    Error: This is a test
        at UserContext.<anonymous> (C:\test\spec\spec.js:3:15)
            -> C:\test\spec\spec.ts:9:15
        at <Jasmine>
        at processImmediate (node:internal/timers:462:21)
```
Note that this includes the correct TypeScript filename as well as correct line/column numbers. Many editors (such as VS Code) can turn this into a clickable link that directly opens the source file.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/ExceptionFormatterSpec.js
+++ b/spec/core/ExceptionFormatterSpec.js
@@ -86,6 +86,37 @@ describe('ExceptionFormatter', function() {
       );
     });
 
+    it('filters Jasmine stack frames from V8-style traces but leaves unmatched lines intact', function() {
+      var error = {
+        message: 'nope',
+        stack:
+          'C:\\__spec__\\core\\UtilSpec.ts:120\n' +
+          "                new Error('nope');\n" +
+          '                ^\n' +
+          '\n' +
+          'Error: nope\n' +
+          '    at fn1 (C:\\__spec__\\core\\UtilSpec.js:115:19)\n' +
+          '        -> C:\\__spec__\\core\\UtilSpec.ts:120:15\n' +
+          '    at fn2 (C:\\__jasmine__\\lib\\jasmine-core\\jasmine.js:7533:40)\n' +
+          '    at fn3 (C:\\__jasmine__\\lib\\jasmine-core\\jasmine.js:7575:25)\n' +
+          '    at fn4 (node:internal/timers:462:21)\n'
+      };
+      var subject = new jasmineUnderTest.ExceptionFormatter({
+        jasmineFile: 'C:\\__jasmine__\\lib\\jasmine-core\\jasmine.js'
+      });
+      var result = subject.stack(error);
+      expect(result).toEqual(
+        'C:\\__spec__\\core\\UtilSpec.ts:120\n' +
+          "                new Error('nope');\n" +
+          '                ^\n' +
+          'Error: nope\n' +
+          '    at fn1 (C:\\__spec__\\core\\UtilSpec.js:115:19)\n' +
+          '        -> C:\\__spec__\\core\\UtilSpec.ts:120:15\n' +
+          '    at <Jasmine>\n' +
+          '    at fn4 (node:internal/timers:462:21)'
+      );
+    });
+
     it('filters Jasmine stack frames from V8 style traces', function() {
       var error = {
         message: 'nope',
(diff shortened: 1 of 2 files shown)


JannesMeyer:
Not sure why the CI on Node v14 failed to be honest. Everything passed on Node v10 and v12.

```
1) Env integration specDone reporting reports the duration of the spec

  Message:

    Expected 9 to be greater than or equal 10.
```

Seems like a rounding error to me?

sgravrock:
Can you tell me how to reproduce the problem that this solves? I regularly use Jasmine in Node to test TypeScript code and I haven't seen a non-source-mapped stack trace in literally years, so I think there's probably an important difference between your setup and mine.

JannesMeyer:
Hmm, that's interesting. I just use the normal TypeScript compiler. No other tools involved (no webpack, etc.)

I have created a minimal example that shows the problematic behaviour: [minimal-example.zip](https://github.com/jasmine/jasmine/files/5478249/minimal-example.zip)

Just run these three commands in the folder:

```
npm install
npm start
npm test
```

You should see output that contains the non-sourcemapped stack trace:

```
Randomized with seed 89343
Started
F

Failures:
1) test suite fails
  Message:
    Error: This is a test
  Stack:
    Error: This is a test
        at UserContext.<anonymous> (C:\minimal-example\spec\spec.js:5:15)
        at <Jasmine>
        at processImmediate (node:internal/timers:462:21)

1 spec, 1 failure
Finished in 0.005 seconds
Randomized with seed 89343 (jasmine --random=true --seed=89343)
```

I am curious about your setup as well. It would be great if there was a way to get it working without modifying Jasmine.

Node.js version: 15.0.1
Jasmine version: 3.6.3

sgravrock:
Thanks, that helps. I'm able to reproduce what you're seeing. I usually use `@babel/register`, which rewrites the existing stack trace lines with correct filenames and line numbers rather than adding an extra line to each frame. 

The approach you've taken here looks fine to me. If you'll add a test for this (probably next to, and in the same style as, the existing [stack trace filtering tests](https://github.com/jasmine/jasmine/blob/b1bcd6e82537ff1849e2d95afa5480101bc85282/spec/core/ExceptionFormatterSpec.js#L89-L128)) and modify the source files in `src/` rather than the build output in `lib/`, I'll merge it.

sgravrock:
The failure you saw on Node 14 is a flaky test that I haven't found the time to deal with. Don't worry about it.

JannesMeyer:
Thanks, that sounds great! I have followed your advice. Could you please review again?

In the test I used Windows-style paths with backslashes and starting with `C:\` instead of `http://`. I thought that adding a little bit variety to the stack traces couldn't hurt. These are the kinds of stack traces that I see in Node.js. Hopefully that is okay with you.

The rest should be in the style as the existing tests.

JannesMeyer:
Sorry, I originally forgot to revert the changes to `jasmine.js` as your CONTRIBUTING.md suggests.

JannesMeyer:
I just squashed my commits into a single commit to make the history a bit cleaner

PR #1867, opened 2020-11-13
HtmlReporter: Handle child of number type

pmccready:
## Description

Really small change to handle the rendering of numerical child elements in the HtmlReporter `createDom` method.

> It is my first time contributing here so apologies if I've not done it correctly.

## Motivation and Context

Currently when the child element is a number it hits the `else` block and tries to use `appendChild` with a number. Unfortunately this breaks as shown in the screenshot below:

image

## How Has This Been Tested?

I discovered this having run tests in work code (sorry, it is confidential so cannot share). I changed the code within node_modules and re-ran my tests - results shown in screenshot below:

image

Also ran the existing tests `npm test` and `node spec/support/ci.js` when following the contributor guidelines.

## Types of changes

## Checklist:

### DIFF
--- a/src/html/HtmlReporter.js
+++ b/src/html/HtmlReporter.js
@@ -641,8 +641,8 @@ jasmineRequire.HtmlReporter = function(j$) {
       for (var i = 2; i < arguments.length; i++) {
         var child = arguments[i];
 
-        if (typeof child === 'string') {
-          el.appendChild(createTextNode(child));
+        if (typeof child === 'string' || typeof child === 'number') {
+          el.appendChild(createTextNode(child.toString()));
         } else {
           if (child) {
             el.appendChild(child);


pmccready:
I believe one way to replicate the issue here is to install the [karma-jasmine-seed-reporter](https://www.npmjs.com/package/karma-jasmine-seed-reporter) and provide a seed value as an integer instead of a string. E.g.

```javascript
var jasmineSeedReporter = require('karma-jasmine-seed-reporter')
 
module.exports = function (config) {
  config.set({
    ...
    plugins: ["karma-*", jasmineSeedReporter],
    reporters: ['jasmine-seed'],
    client: {
      jasmine: {
        random: true,
        seed: 16288 // change to '16288'
      }
    },
  });
};
```

Example error output:

image

slackersoft:
Duplicate of #1851 (which the original submitter has closed).

Please take a look again at the [CONTRIBUTING guide](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md), specifically what to do [before submitting a pull request](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md#before-committing-or-submitting-a-pull-request). Specifically:

* You have added a generated file to your pull request (this generally leads to merge conflicts that we then have to resolve).
* You didn't make changes to the source files, which means your changes will vanish the next time we generate the compiled files
* You have also not added any tests for this functionality. Since Jasmine is _itself_ a testing framework, we believe new code should have tests.


On top of all of this, I'm not sure what is going on with some of those libraries with Karma, but that `createDom` method is not part of Jasmine's external interface, and should only be called by inside Jasmine with arguments that it expects, so I don't think this is something we really need to accommodate in Jasmine. It sounds like the karma-jasmine-order-reporter is causing the issue.

sgravrock:
The random seed type wasn't correctly documented as of the last activity on this PR, but as of 3.8.0 strings and numbers are both valid, so Jasmine ought to handle number seeds correctly 59848ca1 should fix this. If not, feel free to open a new issue with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example).

PR #1871, opened 2020-12-07
Add string shortening in `DiffBuilder`

### LINKED ISSUE #1870: Long String Truncation Omits Failure Cause in Message
## Long String Truncation Omits Failure Cause in Message

Long failure messages get truncated and omit the culprit.

## Current Behavior

When using:

```
expect(a).toBe(b);
```

And `a` and `b` are long strings, the cause of failure can be elided in the reported message, due to the truncation mechanism. For example, if `a='abcdefgh...'` and `b='abcdefgh...(5000 equal characters)x'`, the reported message omits the different `'x'`, present in `b` and not in `a`, as follows: 

```
'Expected abcdefgh...(truncated at character 1000) to equal abcdefgh...(truncated at character 1000)'
```

If you use the reproduction suite below, you will find that the message that gets printed to the console with [jasmine-spec-reporter](https://github.com/bcaudan/jasmine-spec-reporter) will show the **same** string for the expected and the actual values!

## Expected Behavior

Failure message should highlight the culprit. Following the above example, it should report something like: 

<pre>
'Expected [4020 omitted characters ...]bcdefghijklmnopqrst<b>x</b> to equal [4020 omitted characters ...]bcdefghijklmnopqrst'
</pre>

```javascript

describe('repro test', () => {
  it('should give a decent message', () => {
    let filled = 'abcdefghijklmnopqrstuvwxyz';
    filled = filled + filled + filled;
    let q = 1;
    let r = 1;
    const arr = new Array(10000).fill(0).map((x, i) => {
      q = (q * 31) % 67;
      r = (r * 17) % 31;
      return filled.substring(q, q + r);
    });
    const s1 = arr.join('\n');
    arr[9876] += 'x';
    const s2 = arr.join('\n');
    expect(s1).toEqual(s2);
  });
});
```
## Context

Not showing the relevant piece of content that caused an assertion to fail slows down programming: one needs to devise ways to be able to know what is the failure culprit.

jginsburgn:
* Fixes #1870

## Description

Added a function to modify the `actual` and `expected` used to `getMessage` in `DiffBuilder`. This function will be a no-op if `actual` or `expected` is not a string. When both are, the function shortens the used value of each so that at least the first different parts of the strings compared are shown. 
## How Has This Been Tested?

Added a spec.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/matchers/DiffBuilderSpec.js
+++ b/spec/core/matchers/DiffBuilderSpec.js
@@ -210,4 +210,32 @@ describe('DiffBuilder', function() {
 
     expect(diffBuilder.getMessage()).toEqual(expectedMsg);
   });
+
+  it('shortens strings', function() {
+    var abc = 'abcdefghijklmnopqrstuvwxyz',
+      repeatedABC = '';
+    for (var i = 0; i < 100; i++) {
+      repeatedABC += abc;
+    }
+    var expected = repeatedABC;
+    var actual =
+      repeatedABC.substring(0, 1800) + 'XXXXX' + repeatedABC.substring(1805);
+
+    var prettyPrinter = jasmineUnderTest.makePrettyPrinter([]),
+      diffBuilder = new jasmineUnderTest.DiffBuilder({
+        prettyPrinter: prettyPrinter
+      });
+
+    diffBuilder.setRoots(actual, expected);
+    diffBuilder.recordMismatch();
+
+    var expectedMsg =
+      "Expected '[1761 omitted " +
+      'characters...]tuvwxyzabcdefghijklmnopqrstuvwxyzabcdefXXXXXlmnopqrstuvwxyzabcdefghijklmnopqrstu[...759 ' +
+      "omitted characters]' to equal '[1761 omitted " +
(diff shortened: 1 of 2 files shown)


sgravrock:
Thanks for your contribution, and I'm sorry it's taken so long to get back to you. I like the idea of truncating long strings in a smarter way than we do now.

Some thoughts based on a quick read of the code in this PR:

How does this interact with the existing truncation mechanisms in the pretty printer, especially the one that uses MAX_PRETTY_PRINT_CHARS? What happens when both of them truncate the same string?

80 characters seems pretty short. Such a low limit could make test failure output less readable in the case of medium-length strings, especially when there's more than one difference. If we go forward with this approach I think it would make sense to make that limit both larger and configurable.

This also seems light on test coverage. It looks like there are a number of possible code paths but only one test.

sgravrock:
Closing due to inactivity. While I appreciate the effort you put into this PR, I'm not prepared to merge it in its current state. Feel free to resubmit if you decide to come back to this and address the issues mentioned above.

PR #1892, opened 2021-03-26
fix jsdoc config.seed type

UziTech:
## Description

I believe the `seed` configuration option should be a number but the docs say it should be a function.

Also according to the tests it can be a string? so I'm not sure if it should be a string or number, but I know it is not supposed to be a function.

## Motivation and Context

Docs are wrong.

## How Has This Been Tested?

There are already tests for the seed function passing a number as a string.

https://github.com/jasmine/jasmine/blob/cd1131354bea551ef11478e313fe66fb11721cf4/spec/core/integration/SpecRunningSpec.js#L697

## Types of changes

- [x] Docs Change (only documentation changes)

## Checklist:

### DIFF
--- a/src/core/Env.js
+++ b/src/core/Env.js
@@ -53,7 +53,7 @@ getJasmineRequireObj().Env = function(j$) {
        * Null causes the seed to be determined randomly at the start of execution.
        * @name Configuration#seed
        * @since 3.3.0
-       * @type function
+       * @type number
        * @default null
        */
       seed: null,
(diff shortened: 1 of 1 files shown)


sgravrock:
Nice catch. The existing type definitely looks wrong. Thank you.

I'm not sure `number` is the right type though. It looks like boot.js passes a string, and [order.js assumes a string, but probably works fine if the seed is a number](https://github.com/jasmine/jasmine/blob/main/src/core/Order.js#L16). So maybe `(number|string)` would be a more accurate type than just `number`.

sgravrock:
Merged, thanks. I've got a couple of other jsdocs to tackle soon and then I'll publish the updates to the docs site.

Issue #1897, opened 2021-04-07
spyOnAllFunctions does not create any spies for class instance methods


Dari-K:
The specific case is when using classes in Typescript, and passing a class instance to `spyOnAllFunctions`.
Not sure what the expected behaviour is for private vs public methods of class instances, but in this case neither are getting spies created for them.

## Expected Behavior
spyOnAllFunctions creates a spy for each function in the object given.

## Current Behavior
No spies are created.

## Possible Solution
This seems to happen because when compiling to ES5 the generated JS would use functions and prototypes, but when compiling to a later target it just uses javascript classes. With the old prototype style the class methods would end up being enumerable properties, but with the new class style they are not enumerable, which means the `for (prop in obj)` https://github.com/jasmine/jasmine/blob/2fc5182ddcdd67456b89c92c2eedbc2443c928c6/src/core/SpyRegistry.js#L179
does not find anything because it loops over enumerable properties only. Using `Object.getOwnPropertyNames` would get all the properties, though this would include the constructor, callee, caller, arguments, etc, which would not have been included before as they are not enumerable.

slackersoft:
We would be happy to review a pull request to add the ability spy on things with this configuration as long as it doesn't break the current functionality. (It probably also shouldn't accidentally spy on the constructor itself).

Dari-K:
What would be an acceptable way to skip the constructor, is it good enough just to skip functions that are called constructor (is there anything else we can check)? Should this be a separate function something like a toggle `includeUnenumerable` which defaults to false?

PR #1909, opened 2021-04-22
Allow option for spyOnAllFunctions to include non-enumerable props

Dari-K:
## Description

- Added a new optional parameter to `spyOnAllFunctions` which, when true, will also check non-enumerable properties of the passed object.
- Could instead be split into 2 functions.
- I set it up to ignore the Object prototype, but only when you are including non-enumerable so as not to change the old behaviour, so this while condition looks a bit strange.
- To make sure we don't spy on the constructor I'm just ignoring any non-enumerable properties called constructor (again, no change for when non-enumerable is falsey, just uses all the enumerable props).
-  try-catch added because now we are accessing non-enumerable properties we might get an error when trying to get a property (TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them). Not sure if there's a nicer way to make sure we can access a property before we try.
- Added a test to make sure non-enumerable properties are spied on when true is passed, and one to make sure Object.prototype is not spied on. Do all the existing `spyOnAllFunctions` tests need to run again for true being passed though?

## Motivation and Context
In newer JS and compiled TS using `class`, methods end up non-enumerable, so `spyOnAllFunctions` does not spy on anything.
See #1897 

## How Has This Been Tested?
Tested on:
node 12.13.0
chrome 90.0.4430.85
firefox 86.0.1

## Types of changes

## Checklist:
- [x] My code follows the code style of this project. (Ran prettier)
- [x] My change requires a change to the documentation. (New param for spyOnAllFunctions)

### DIFF
--- a/src/core/SpyRegistry.js
+++ b/src/core/SpyRegistry.js
@@ -163,7 +163,7 @@ getJasmineRequireObj().SpyRegistry = function(j$) {
       return spy;
     };
 
-    this.spyOnAllFunctions = function(obj) {
+    this.spyOnAllFunctions = function(obj, includeNonEnumerable) {
       if (j$.util.isUndefined(obj)) {
         throw new Error(
           'spyOnAllFunctions could not find an object to spy upon'
@@ -171,30 +171,27 @@ getJasmineRequireObj().SpyRegistry = function(j$) {
       }
 
       var pointer = obj,
-        props = [],
-        prop,
-        descriptor;
+        propsToSpyOn = [],
+        properties,
+        propertiesToSkip = [];
 
-      while (pointer) {
-        for (prop in pointer) {
-          if (
-            Object.prototype.hasOwnProperty.call(pointer, prop) &&
-            pointer[prop] instanceof Function
-          ) {
-            descriptor = Object.getOwnPropertyDescriptor(pointer, prop);
-            if (
-              (descriptor.writable || descriptor.set) &&
-              descriptor.configurable
-            ) {
-              props.push(prop);
-            }
-          }
-        }
+      while (
+        pointer &&
+        (!includeNonEnumerable || pointer !== Object.prototype)
+      ) {
+        properties = getProps(pointer, includeNonEnumerable);
+        properties = properties.filter(function(prop) {
+          return propertiesToSkip.indexOf(prop) === -1;
+        });
+        propertiesToSkip = propertiesToSkip.concat(properties);
+        propsToSpyOn = propsToSpyOn.concat(
+          getSpyableFunctionProps(pointer, properties)
+        );
         pointer = Object.getPrototypeOf(pointer);
       }
 
-      for (var i = 0; i < props.length; i++) {
-        this.spyOn(obj, props[i]);
+      for (var i = 0; i < propsToSpyOn.length; i++) {
+        this.spyOn(obj, propsToSpyOn[i]);
       }
 
       return obj;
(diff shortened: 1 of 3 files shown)


src/core/SpyRegistry.js
```diff
@@ -175,18 +175,34 @@ getJasmineRequireObj().SpyRegistry = function(j$) {
...
-              (descriptor.writable || descriptor.set) &&
-              descriptor.configurable
-            ) {
-              props.push(prop);
+      while (
+        pointer &&
+        (!includeNonEnumerable || pointer !== Object.prototype)
+      ) {
+        var enumerableProperties = Object.keys(pointer);
+        var properties = includeNonEnumerable
```
sgravrock, line 183: I find long multiline ternaries hard to read, especially when they have contain large expressions like the filter call here. I'd rather see this written as an if-else. There's some missing test coverage as well: if I make the filter function always return true, the tests still pass. I think there should be a test that verifies that non-enumerable constructors are filtered out and one that verifies that enumerable constructors are included.

src/core/SpyRegistry.js
```diff
@@ -175,18 +175,34 @@ getJasmineRequireObj().SpyRegistry = function(j$) {
...
+              return (
+                prop !== 'constructor' ||
+                enumerableProperties.includes('contructor')
+              );
+            })
+          : enumerableProperties;
+        for (var i = 0; i < properties.length; i++) {
+          prop = properties[i];
+          if (Object.prototype.hasOwnProperty.call(pointer, prop)) {
+            try {
```
sgravrock, line 194: Big try-catch-ignore blocks are scary because it's easy to end up ignoring more exceptions than intended. I'd rather see the try block scoped to the single expression that's expected to throw the exception. It would also be good to have a test for the scenario that the try-catch supports.

sgravrock (changes requested):
Thanks for your contribution. I think this looks like a good approach, but it could use some more test coverage and a bit of stylistic cleanup. Please see my review comments.

Dari-K:
Thanks for reviewing, will clean up when I've got some time. 👍

src/core/SpyRegistry.js
Dari-K: Try-catch is over only the line where we access the property now.

src/core/SpyRegistry.js
Dari-K: Removed the ternary, and added tests to make sure we are actually spying on enumberable constructors, but not non-enumerable constructors.

src/core/SpyRegistry.js
```diff
@@ -175,18 +175,41 @@ getJasmineRequireObj().SpyRegistry = function(j$) {
...
+          properties = enumerableProperties;
+        }
+        for (var i = 0; i < properties.length; i++) {
+          prop = properties[i];
+          if (Object.prototype.hasOwnProperty.call(pointer, prop)) {
+            var value;
+            try {
+              value = pointer[prop];
+            } catch (e) {
+              /* tried to access a property that should not be accessed */
```
sgravrock, line 203: I missed it the first time through, but there's a subtle bug here. Or possibly several subtle bugs interacting with each other. When I run the tests on Safari 13 and 14:

```
Failures:
1) SpyRegistry #spyOnAllFunctions when includeNonEnumerable is true should not throw an exception if we try and access strict mode restricted properties
  Message:
    Expected function not to throw, but it threw Error: <spyOn> : caller is not declared writable or has no setter
    Usage: spyOn(<object>, <methodName>).
  Stack:
    <Jasmine>
    http://localhost:53530/__spec__/core/SpyRegistrySpec.js:498:31
    <Jasmine>
```

I think there's at least two things going on. The main issue is that certain properties of functions on Safari are defined as completely inaccessible on the function itself but as read-only higher up the prototype chain. The right thing to do is to treat the first descriptor found for a given property as definitive and ignore any others that are further up the prototype chain, but this code gives a read-write property descriptor on an object's prototype precedence over a read-write property descriptor on the object itself. If you don't have Safari handy you should be able to reproduce the problem by setting up a similar object graph using `Object.create` and `Object.defineProperty`, and that could be the foundation for a unit test that exercises this case in all browsers.

The other issue is that if we catch an exception on line 202, we continue on with the rest of the loop reusing the value of the `value` variable from the previous iteration. I don't know if that's currently causing any visible breakage but it's definitely not right. There's three things I'd like to see here:

1. Break out of the loop from the catch block.
2. Consider extracting parts of this into smaller functions, if that will make the flow control easier to understand.
3. Move all `var` statements to the beginning of the function, to match their actual scope and avoid the suggestion that they have block scope like variables in other languages.

sgravrock (changes requested):

sgravrock:
Looks good. Would you mind squashing your commits? Or I can do it for you, but in that case Github will show the PR as closed rather than merged.

PR #1912, opened 2021-05-11
Remove negative margin

### LINKED ISSUE #1788: Negative margin shifts content
The element with the CSS class `jasmine_html-reporter` has a margin of `-8px`. This results in the compiled components being pushed up into the gray box and the gray box is not 100% wide and leaves a small gap on the right side:

Karma

## Expected Behavior
Page layout is correct and no elements overlap each other.

## Current Behavior
Tested components overlap with gray box + white gap on the right side.

## Possible Solution
Add following rules:
```css
body {
  margin: 0;
}

.jasmine_html-reporter {
  box-sizing: border-box;
  // margin: -8px;        // removing this line from the current CSS
}
```

## Your Environment
* Version used: 4.4.1
* Environment name and version: Chrome 79
* Operating System and version (desktop or mobile): macOS 10.15.3

Spielboerg:
## Description

The styles contain a negative margin that causes a shift up and to the left.

## Motivation and Context

Fixes #1788 

## Types of changes

This change could break existing tests if, for example, the position of an HTML element is checked relative to the page.

## Checklist:

### DIFF
--- a/src/html/_HTMLReporter.scss
+++ b/src/html/_HTMLReporter.scss
@@ -30,7 +30,6 @@ body {
   width: 100%;
   background-color: $page-background-color;
   padding: 5px;
-  margin: -8px;
 
   font-size: $font-size;
   font-family: Monaco, "Lucida Console", monospace;


sgravrock:
I'm afraid it's not quite as simple as that. This change shifts the Jasmine HTML reporter down and to the right when it's the only thing on the page: 

jasmine shifted

(You can see that for yourself by running `npm run serve` on your PR branch and looking at <http://localhost:8888/>.)

In theory that could be fixed by setting `margin: 0` on the body, as you suggested in the linked issue. But we  avoid setting styles on DOM elements that we don't own, because that can affect the behavior of people's code under test. See [the PR that introduced the negative margin](https://github.com/jasmine/jasmine/pull/600).

PR #1919, opened 2021-06-13
Update .npmignore

styfle:
## Motivation and Context
Prevent unnecessary files from being published to npm.

See https://unpkg.com/browse/jasmine@3.7.0/

## How Has This Been Tested?
`npm pack`

## Types of changes

## Checklist:


sgravrock:
Nice catch. Thank you. I decided to solve this by specifying the files to include rather than the ones to exclude. I think that's going to be a safer approach in the long run than what we've been doing.

PR #1937, opened 2021-10-07
added additional error msg for delayedExpectation

AndreWillomitzer:
## Description

The current error message as described above may leave the user confused as to why the error occurred. As suggested by the issue raiser and others I added some extra messages with numbers as a "checklist" of things that might be a cause such as 
**Was done() invoked before an async operation completed?**
**Did an expectation follow a call to done()** 
The second of which was mentioned by @slackersoft 
## Motivation and Context

This change was required because the current error message is describing that the issue can only be with expectAsync, when really done() being called before any async or expectation is done is the issue I believe.
## How Has This Been Tested?

I ran npm test which runs all the jasmine suite tests, and in there I saw that my change was reflected: 
image

If there is a better way to test this I really want to learn!!! Please let me know and I will do so. I am new to pull requesting and Open Source.

Since the change is an error message and simply a change in the string I did not attempt to reproduce the issue mentioned.

- Windows 10
- Version 3.9.0
- 

This will affect other areas of the code because when an error message is printed regarding late expectations, this new message will be pushed to the top along with delayedExpectationResult.
## Types of changes

## Checklist:

### DIFF
--- a/spec/core/integration/EnvSpec.js
+++ b/spec/core/integration/EnvSpec.js
@@ -2932,7 +2932,9 @@ describe('Env integration', function() {
           message:
             'Spec "a suite does not wait" ran a "toBeResolved" expectation ' +
             'after it finished.\n' +
-            'Did you forget to return or await the result of expectAsync?',
+            '1. Did you forget to return or await the result of expectAsync?\n' +
+            '2. Was done() invoked before an async operation completed?\n' +
+            '3. Did an expectation follow a call to done()?',
           matcherName: 'toBeResolved'
         }),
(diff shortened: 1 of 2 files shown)


AndreWillomitzer:
Yahoo all checks passed!!!!!!!!! Let me know if this error message works for you and I can change it if need be :) 👍 👍 👍 👍 👍 👍 👍

AndreWillomitzer:
Okay thanks will get to that first thing this week probably even tomorrow :) Much appreciated for looking it over.

PR #1953, opened 2022-01-18
Avoid NPE exception when  is null under IE9

aleen42:
## Description

`Error.stack` Is not supported under IE9, and it means that an NPE exception will be thrown when trying to access `error.stack.split` under the method `StackTrace()`.

## Motivation and Context

To avoid this blocking exception.

## How Has This Been Tested?

Integrated with Krama, and TrifleJS.

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -679,12 +679,18 @@ getJasmineRequireObj().util = function(j$) {
     // Don't throw and catch. That makes it harder for users to debug their
     // code with exception breakpoints, and it's unnecessary since all
     // supported environments populate new Error().stack
-    return new Error();
+    const err = new Error();
+    if (err.stack) {
+      return err;
+    }
   };
 
   function callerFile() {
-    var trace = new j$.StackTrace(util.errorWithStack());
-    return trace.frames[2].file;
+    var err = util.errorWithStack();
+    if (err) {
+      var trace = new j$.StackTrace(err);
+      return trace.frames[2].file;
+    }
   }
 
   util.jasmineFile = (function() {


sgravrock:
Thanks for the contribution, but Jasmine dropped support for IE 9 almost four years ago. This issue is just the tip of a very large iceberg of things that won't work in that browser. If you need to test against IE, I suggest using Jasmine 2.99 or earlier.

aleen42:
I have tested that it is the only problem, and it seems not a large iceberg actually, @sgravrock 

Please think it twice. Otherwise, I have to patch `Error.stack`.

sgravrock:
I'm not sure how you're able to get Jasmine running in IE 9. Jasmine currently uses a number of JavaScript features that IE (any version) never supported such as async functions, generators, the spread operator, arrow functions, etc. as well as library features like promises and iterable maps and sets. When I try to run the current version in IE 11 (the only version of IE that I have access to any more), the browser fails to even parse the JS file:

ie 11

We recently dropped IE 10 and 11 support because it was an obstacle to contributors and it was preventing us from moving Jasmine forward. Continuing to support IE 9 would be even harder. Aside from the JavaScript compatibility issues, we don't even have any computers that can run IE 9. The service that we use to run tests on browsers in CI doesn't provide IE 9 any more. As far as I know none of the Windows versions that can run it still receive security updates, which means it's not safe to connect them to the internet.

Sorry, but the answer is no. IE 9 is long dead and we're not going to bring support for it back.

aleen42:
@sgravrock Thanks a lot. I think it is the duty of `karma-jasmine` to patch for it.

PR #1959, opened 2022-02-16
Add toHaveSpyInteractions matcher to check spy interactions on spy objects

### LINKED ISSUE #1568: Feature request: Matcher that verifies that none of the spies in a SpyObj were called
## Desired Behavior
A matcher that verifies that none of the spies of the specified object have been called - i.e. there have been zero interactions with an entire mock service.
```
httpClient = jasmine.createSpyObj('HttpClient',
  ['get', 'post', 'put', 'delete', 'head', 'options', 'patch']);

// Checks that there have been no calls to httpClient.get, httpClient.post, etc.
expect(httpClient).not.toHaveBeenCalled();
```

## Current Behavior
```
Error: <toHaveBeenCalled> : Expected a spy, but got
  Object({ get: spy on HttpClient.get, post: spy on HttpClient.post, ... })
```

Thus requiring:
```
expect(httpClient.get).not.toHaveBeenCalled();
expect(httpClient.post).not.toHaveBeenCalled();
expect(httpClient.put).not.toHaveBeenCalled();
expect(httpClient.delete).not.toHaveBeenCalled();
expect(httpClient.options).not.toHaveBeenCalled();
expect(httpClient.head).not.toHaveBeenCalled();
expect(httpClient.patch).not.toHaveBeenCalled();
```

## Context
If I want to test there have been no HTTP calls for a specific condition, I don't want to have to check every possible method - someone might add a PUT in there but the test doesn't checking for that so still passes.

Other frameworks such as Java's Mockito have this capability - `Mockito.verifyZeroInteractions(mock)` - to allow us to verify that a particular service was not interacted with at all.

nitobuendia:
## Description
toHaveSpyInteractions matcher checks whether the spies contained within a spy object have been called. It returns true if one or more of the spies of the spy object has been called and false otherwise.

## Motivation and Context
Fixes #1568 - full context and motivation is provided there.

## How Has This Been Tested?
Run `npm test`. Run `npm run build` and run `npm test` again.

## Types of changes

## Checklist:


nitobuendia:
From the checklist I understand that I need to / should update the documentation. Any guidance on where the documentation lives? Thanks in advance.

nitobuendia:
Closed by accidental rename of branch. Reopening for review.

sgravrock:
Thanks for your contribution. The implementation looks good to me, but there are some issues with the tests. Please see my comments on spec/core/matchers/toHaveSpyInteractions.js.

The jsdoc comment that you've already written is enough documentation (we'll take care of getting it over to the docs site after the PR is merged). But please remove the @since annotation. The version it references already shipped without this feature.

spec/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,57 @@
+describe('toHaveSpyInteractions', () => {
```
sgravrock, line 1: This spec file doesn't run because its name doesn't end in Spec.js. If I change the name, all the specs fail.

spec/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,57 @@
...
+  let spyObj;
+  beforeEach(() => {
```
sgravrock, line 3: I'd rather see each spec contain just the setup it needs, inline. De-duplicating test setup like this introduces coupling between tests and reduces locality. Sometimes that's worthwhile, but IMO not in this case where we're looking at one line of setup in most specs.

spec/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,57 @@
...
+    spyObj = jasmineUnderTest.createSpyObj('NewClass', ['spyA', 'spyB']);
+    spyObj.otherMethod = function() {};
+  });
+
+  it('detects spy interactions', () => {
+    spyObj.spyA();
+    expect(spyObj).toHaveSpyInteractions();
```
sgravrock, line 10: This (and other lines like it) tests the functionality of the Jasmine that's hosting the test run, not the Jasmine under test. See <https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md#self-testing>, <https://jasmine.github.io/pages/faq.html#self-test>, and neighboring files like toHaveBeenCalledSpec.js.

You might not be able to create a spy object via the obvious approach of calling `jasmineUnderTest.createSpyObj` -- I don't think we currently expose that method. In that case you can use `jasmineUnderTest.getEnv().createSpyObj` instead.

spec/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,57 @@
...
+
+  [true, 123, 'string'].forEach(testValue => {
+    it(`throws error if a non-object (${testValue}) is passed`, () => {
+      expect(() => {
+        expect(true).toHaveSpyInteractions();
+      }).toThrowError(Error, /Expected a spy object, but got/);
+    });
+  });
+
+  [['argument'], [false, 0]].forEach(testValue => {
```
sgravrock, line 37: What's the significance of `[false, 0]`? It seems to me that this could probably be reduced down to a single spec, but if we need both cases I'd rather see something that makes it clear to future maintainers (e.g. me in 6 months) why those specific values were chosen.

src/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,73 @@
...
+
+        if (!hasSpy) {
+          throw new Error(
+            getErrorMsg(
+              'Expected a spy object with spies, but object has no spies.'
+            )
+          );
+        }
+
+        let resultMessage = 'Expected spy object spies to have been called';
```
sgravrock, line 54: This doesn't produce the right message when the matcher is negated. GIven this spec:

```
  it('handles negation correctly', function() {
    const foo = jasmine.createSpyObj('foo', ['bar']);
    foo.bar();
    expect(foo).not.toHaveSpyInteractions();
  });
```

I get this message:

```
Expected spy object spies to have been called, the following spies were called: foo.bar called 1 time(s)
```

Something like this would be more appropriate:

```
Expected spy object spies not to have been called, but the following spies were called: foo.bar called 1 time(s)
```

Or better yet:

```
Expected spy object spies not to have been called, but foo.bar was called
```

The most important thing is that the failure message reads correctly in both the normal and negated cases. I'm less opinionated about whether to include the call count, although I think it's going to be excess noise for most users.

sgravrock (changes requested):

spec/core/matchers/toHaveSpyInteractions.js
slackersoft: Please don't use this form of test in Jasmine's suite, especially for this small of a set of possible options. As @sgravrock points out, it is generally hard to understand what is really being tested, and is not in the style of the rest of the Jasmine repository.

Additionally, the `${testValue}` interpolation will only work inside of Node.js and not in the browsers that Jasmine also tests against, leaving test failures very hard to track down.

spec/core/matchers/toHaveSpyInteractions.js
nitobuendia: Renamed file to add Spec

spec/core/matchers/toHaveSpyInteractions.js
```diff
@@ -0,0 +1,57 @@
+describe('toHaveSpyInteractions', () => {
+  let spyObj;
+  beforeEach(() => {
+    spyObj = jasmineUnderTest.createSpyObj('NewClass', ['spyA', 'spyB']);
+    spyObj.otherMethod = function() {};
+  });
+
+  it('detects spy interactions', () => {
```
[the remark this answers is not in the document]
nitobuendia: Replaced arrow functions with anonymous functions.

spec/core/matchers/toHaveSpyInteractions.js
nitobuendia: I was following DRY principle since the same setup is required for all tests except for one, but I do not have a strong opinion here. Hence, moved the setup to each test call.

spec/core/matchers/toHaveSpyInteractions.js
nitobuendia: Done. Thanks.

spec/core/matchers/toHaveSpyInteractions.js
nitobuendia: I will simplify `  [['argument'], [false, 0]].forEach(function(testValue) {` and just use one test. The relevance here is simply to have one or multiple arguments, but I do not think the value is large.


@slackersoft 
However, what's the suggestion on this test?

```
  [true, 123, 'string'].forEach(function(testValue) {
    it(`throws error if a non-object (${testValue}) is passed`, function() {
      let spyObj = jasmineUnderTest.getEnv().createSpyObj('NewClass', ['spyA', 'spyB']);

      expect(function() {
        expect(testValue).toHaveSpyInteractions();
      }).toThrowError(Error, /Expected a spy object, but got/);
    });
  });
```

This is trying to make sure that the logic is detecting as many as non-objects as possible. In order to avoid this for loop, I could think of two options:


A) I would need to create one new test for each of the possible types. This would go against DRY principle as we will have 3 tests (potentially more in the future) with exactly the same code, testing exactly the same principle and just changing the input. 

B) I would need to do the loop inside the function. However, if the tests fails, it would be hard to debug which exact variant is failing, which partially defeats the purpose of testing.

What's your suggested approach here?

src/core/matchers/toHaveSpyInteractions.js
nitobuendia: I would need some guidance here on the inners of Jasmine. How can I detect within toHaveSpyInteractions matcher whether this was called as `expect().toHaveSpyInteractions()` or `expect().not.toHaveSpyInteractions()`? Could you point me to an example or documentation? Thanks!

spec/core/matchers/toHaveSpyInteractions.js
nitobuendia: I've also refactored the tests to rely on jasmineUnderTest with `let matcher = jasmineUnderTest.matchers.toHaveSpyInteractions();` - currently one test is still failing, but the rest are resolved.

nitobuendia: The remaining test has been solved and all tests are passing now.

nitobuendia:
Thanks for the review. I've addressed all the comments, except 2 where I would appreciate guidance:

1. On the multivariant test that is using the for loop - I do not see a better way to achieve the results without overly repeating code pointelessly. Open to suggestions.

2. On the negation message: I agree with the feedback, but I am unsure how to detect this on the matcher itself.

spec/core/matchers/toHaveSpyInteractions.js
slackersoft: I think aiming for DRY in your tests is usually harmful to the readability and debuggability of your test suite, and as a rule, Jasmine does _NOT_ try to achieve DRY-ness in its own test suite. An array of 3 items, probably doesn't even deserve a loop. If you look at tests for other matchers, you'll see that we tend to have a single spec labeled something like "throws an error if a non-object is passed" and then inside that single test multiple explicit calls to `expect` with the different values.

src/core/matchers/toHaveSpyInteractions.js
slackersoft: https://jasmine.github.io/tutorials/custom_matcher

sgravrock:
> 2\. On the negation message: I agree with the feedback, but I am unsure how to detect this on the matcher itself.

Take a look at how other built-in matchers like [toHaveBeenCalledWith](https://github.com/jasmine/jasmine/blob/dfa5b6a53d2b12301b0fbd93d496ab14bacfbf5c/src/core/matchers/toHaveBeenCalledWith.js#L47-L57) do it. They don't detect negation. They just assume that if the matcher passes, the failure message will only be displayed if it was negated.

src/core/matchers/toHaveSpyInteractions.js
nitobuendia: I had seen the customer_matcher. I think this explanation is what clarified the behaviour:

> Take a look at how other built-in matchers like [toHaveBeenCalledWith](https://github.com/jasmine/jasmine/blob/dfa5b6a53d2b12301b0fbd93d496ab14bacfbf5c/src/core/matchers/toHaveBeenCalledWith.js#L47-L57) do it. They don't detect negation. They just assume that if the matcher passes, the failure message will only be displayed if it was negated.


I've updated the logic and tests now to follow that pattern.

nitobuendia:
This should be ready for re-review. All requested changes have been applied.

PR #1990, opened 2022-11-27
fix: use instanceof to check if obj is Node

bsafwen:
Use instanceof Node when jasmine runs in a browser.

## Description

I work on DOM serialization into javascript Objects, and because isDomNode incorrectly identify the serialized objects as DOM nodes, my test passes even though the serialization was incorrect.

## Motivation and Context

```typescript
  it("should fail but it does not", () => {
    expect({
      nodeType: 1,
      nodeName: "a"
    }).toEqual({
      nodeType: 2,
      nodeName: "b"
    });
  });
```

## How Has This Been Tested?
npm run test
npm run serve and manually checked in Firefox, Chrome, Safari.
npm test
npm run build and retested

I did not understand this step though:
`4. Revert your changes to jasmine.js and jasmine-html.js.`

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project. (Not sure but happy to change)

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -309,8 +309,14 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return Object.prototype.toString.apply(value);
   };
 
+  const isBrowser =
+    typeof window !== 'undefined' && typeof window.document !== 'undefined';
+
   j$.isDomNode = function(obj) {
     // Node is a function, because constructors
+    if (isBrowser) {
+      return obj instanceof Node;
+    }
     return typeof jasmineGlobal.Node !== 'undefined'
       ? obj instanceof jasmineGlobal.Node
       : obj !== null &&


lib/jasmine-core/jasmine.js
```diff
@@ -309,8 +309,14 @@ getJasmineRequireObj().base = function(j$, jasmineGlobal) {
     return Object.prototype.toString.apply(value);
   };
 
+  const isBrowser =
+    typeof window !== 'undefined' && typeof window.document !== 'undefined';
+
```
bsafwen, line 314: is this the right place?

sgravrock:
I can't reproduce the problem. When I run the example spec that you provided, it fails as it should:

Screen Shot 2022-11-27 at 12 31 03 PM

I also wouldn't expect your code change to make any difference. `jasmineGlobal` should be `window` in a browser context, and in that case your check is equivalent to the one that's already there. Additionally, if Jasmine was misidentifying regular objects as DOM nodes, I'd expect the effect to be an exception when `MatchersUtil` tries to call `isEqualNode` on them, not a wrong comparision result. 

I wonder if there's something unusual about your setup that causes `jasmineGlobal` to get the wrong value. Could you provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) that demonstrates the problem?

Additionally, there are a couple of problems with the code changes in this PR:

* You've edited the build product instead of the source files. (That's probably why you didn't understand the instructions to revert your changes to jasmine.js and jasmine-html.js.)
* You haven't written any tests for the new functionality.

But let's try to come to a shared understanding of what the bug is before you make any revisions.

bsafwen:
I am sorry for disturbing you, by checking the source in node_modules I see that my jasmine-core version does not have the fix made on the main branch

image

Thanks for maintaining the project, I will be closing this.

Issue #1991, opened 2022-12-08
Verify no other calls


jerone:
### Feature Proposal

I want to verify that no other calls have been made after verifying the known call.

### Context

Some mocked spies can have many methods. Instead of checking every method to not have been called, I'm looking for a way to do this automatically.

I know there is `toHaveSpyInteractions`, but that only checks if any interactions have been made.

### Example

I **want** to write something like this:

```ts
describe("Multiple spies", function () {
  let person;
  beforeEach(function () {
    person = jasmine.createSpyObj('person', ['eat', 'drink', 'sleep', 'walk']);
    person.eat();
  });
  it("creates spies for each requested function", function () {
    expect(person.eat).toHaveBeenCalled();
    expect(person).verifyNoOtherCalls(); // <-- feature request.
  });
});
```

Instead of **having** to write something like this:
```ts
describe("Multiple spies", function () {
  let person;
  beforeEach(function () {
    person = jasmine.createSpyObj('person', ['eat', 'drink', 'sleep', 'walk']);
    person.eat();
  }); 
  it("creates spies for each requested function", function () {
    expect(person.eat).toHaveBeenCalled();
    expect(person.drink).not.toHaveBeenCalled();
    expect(person.sleep).not.toHaveBeenCalled();
    expect(person.walk).not.toHaveBeenCalled();
  });
});
```

### Other Information

In C# there is something similar for `Moq`, called [`VerifyNoOtherCalls`](https://stackoverflow.com/a/71011973/108448). 

Looking through the source-code, it looks like this would only be possible if the `toHaveBeenCalled` method (and similar) sets some kind of `isVerified` state to `true` in the [`CallTracker`](https://github.com/jasmine/jasmine/blob/135ff20123d91da374a4ae5ed41ebdb36535c6f6/src/core/CallTracker.js) most likely. Exactly [like `Moq` does](https://github.com/moq/moq4/blob/b7b07279c45c29a201399a7fa93764892422e687/src/Moq/IInvocation.cs#L30-L33).

sgravrock:
That seems like it would be a nice addition. I occasionally use the equivalent Mockito facility. I'd rather call the matcher `toHaveNoOtherSpyInteractions` or `toHaveNoMoreSpyInteractions` for consistency and to make the expectation read better, but other than that it makes sense to me.  Want to submit a PR?

PR #1999, opened 2023-02-24
#1991 Add toHaveNoOtherSpyInteractions matcher to verify that no not checked interactions

Ulys:
## Description
Added matcher `toHaveNoOtherSpyInteractions`.
This matcher receives an object with Spies and if there were any spy calls which wasn't checked in the same suit before with
- `toHaveBeenCalled`
- `toHaveBeenCalledWith`
- `toHaveBeenCalledBefore`
- `toHaveBeenCalledTimes`
fails. And if there are no calls or all calls were verified with mentioned matchers succeed.

## Motivation and Context
Implementation of issue #1991 - content and motivation provided there.

## How Has This Been Tested?
Run `npm run build && npm run test`
After it did `npm run serve` and checked in Chrome 109 + 110 and Safari 16.1

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/boot0.js
+++ b/lib/jasmine-core/boot0.js
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2008-2023 Pivotal Labs
+Copyright (c) 2008-2022 Pivotal Labs
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
--- a/lib/jasmine-core/boot0.js
+++ b/lib/jasmine-core/boot0.js
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2008-2022 Pivotal Labs
+Copyright (c) 2008-2023 Pivotal Labs
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
--- a/spec/core/CallTrackerSpec.js
+++ b/spec/core/CallTrackerSpec.js
@@ -147,13 +147,13 @@ describe('CallTracker', function() {
   it('default value for interactionChecked is false', function() {
     const callTracker = new jasmineUnderTest.CallTracker();
 
-    expect(callTracker.getInteractionChecked()).toBeFalsy();
+    expect(callTracker.getInteractionChecked()).toBeFalse();
   });
 
   it('allows interactionChecked to be set', function() {
     const callTracker = new jasmineUnderTest.CallTracker();
 
     callTracker.markInteractionAsChecked();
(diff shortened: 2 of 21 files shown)


lib/jasmine-core/boot0.js
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2008-2022 Pivotal Labs
+Copyright (c) 2008-2023 Pivotal Labs
```
sgravrock, line 2: Please revert your changes in lib/.

spec/core/CallTrackerSpec.js
```diff
@@ -143,4 +143,17 @@ describe('CallTracker', function() {
 
     expect(callTracker.mostRecent().args).toEqual(args);
   });
+
+  it('default value for interactionChecked is false', function() {
+    const callTracker = new jasmineUnderTest.CallTracker();
+
+    expect(callTracker.getInteractionChecked()).toBeFalsy();
```
sgravrock, line 150: Better to use `.toBe(false)` or `.toBeFalse()` here, and `toBe(true)` or `toBeTrue()` in the following spec, since we know exactly what the value should be.

src/core/CallTracker.js
```diff
@@ -125,6 +125,24 @@ getJasmineRequireObj().CallTracker = function(j$) {
     this.saveArgumentsByValue = function() {
       opts.cloneArgs = true;
     };
+
+    /**
```
sgravrock, line 129: I don't think markInteractionAsChecked and getInteractionChecked should be part of Jasmine's public API, so they shouldn't have jsdocs.

sgravrock (commented):
Thanks for the PR. This looks like solid, well tested work. Would you mind addressing the comments I added?

ReDICKulous (approved):

sgravrock:
I found a problem when I was manually testing this prior to merging:

```
  it('handles multiple interactions with a single spy', function() {
    const spyObj = jasmine.createSpyObj('spyObj', ['foo']);
    spyObj.foo('x');
    spyObj.foo('y');
    expect(spyObj.foo).toHaveBeenCalledWith('x');
    expect(spyObj).toHaveNoOtherSpyInteractions();
  });
```

I'd expect that spec to fail because no matcher matches `spyObj.foo('y')`, but it passes. The spy system probably needs to track whether individual calls have been checked.

Ulys:
I will take a look.

sgravrock:
Closing since this PR has gone almost a year with no updates and isn't in a merge-able state. Please feel free to open a new PR if you decide to pick this work back up.

PR #2009, opened 2023-05-18
Create oit (options it) function

jcarlosadm:
## Description
This Pull Request add a function `oit`, which is the same as `it` with options. These options are described below.

## Motivation and Context
(see below)

## How Has This Been Tested?
An `OptionsSpec.js` file has been created with all the necessary tests.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation (new feature only).

# Test Smells

Code smell is code that does not follow good software engineering practices. Code like this can cause several problems, like misunderstandings, bugs, or maintenance problems.

```javascript
// from https://blog.bitsrc.io/your-javascript-smells-3574121bcfbe
let a = "start";
let b;

for(let i = 0; i < 5; i++) {
  b += a[i].toUpperCase();
}
```

Like code smell, there is test smell, which is applied to test code. The following code is an example of a test smell:

```javascript
it("testing network", () => {
    if (!isOffline) {
        expect(testNetwork()).toBe(true);
    }
});
```

In the above test, if the constant `isOffline` has a true value, then the assert will not be executed. In this case, it gave the impression that it had passed.

How can we deal with the test smell? One way is to refactor the smell as soon as it is detected. Test code has unique structures that make it distinct from the code smell, like isolation of tests or repetition of code executions. Because of this, we can't use the same solutions for code smells.

## One solution

One solution for this problem is to use markers on a test. With markers, that test will execute slightly differently. For example, to simulate a network attack, you can run a specific test 500 times. Instead of a for loop, you can use a marker that makes that test repeat 500 times. So, what are the differences? Well, in that case, readability. To illustrate that, see the next section.

## JUnit 5 annotations

JUnit has an annotation called `@RepeatedTest`. That annotation can make a test repeat n times. Take the following example:

```java
@Test
void networkAttackTest() {
    for (int i = 0; i < 500; i++) {
        // code of test with asserts
    }
}
```

With the annotation of JUnit 5,

```java
@RepeatedTest(500)
void networkAttackTest() {
    // code of test with asserts
}
```

The second piece of code is more readable, only with the test code, without the for loop.

JUnt 5 has a lot of test annotations to deal with test smells, like `@enableIf` or `@ParameterizedTest`. These annotations can improve rediability and reduce test smells.

## Decorators?

Jasmine doesn't work with *decorators* to modify tests. However, *annotations* (in the context of Java) are metadata applied to the annotated target. So we can work this way, applying metadata to the test to change its execution. We show how we can apply this to Jasmine in the next section.

# Option it!

Instead of using decorators, we can use a function, something similar to `it`. And add metadata through a parameter or *option*.

First, let's create the option object:

```javascript
var option = {
    enableIf: {
        value: true
    }
};
```

This *option* is metadata that tells us that we want `enableIf` active. So, our goal with this is to only enable the test under certain circumstances.

After that, just add the metadata to the test. To do this, we need to create a function similar to `it`. We create a function called `oit`, which means *option it*, following the patterns of other jasmine functions, i.e., `xit` and `fit`. The `oit` has all parameters of `it`, adding a new one: *options*.

Following the last example, just add the option to the test:

```javascript
oit("generic test", () => {
    expect(2 == 2).toBe(true)
}, {
    enableIf: {
        value: true
    }
});
```

## Types

There are three options: `enableIf`, `repeat`, and `parametersValue`.

### enableIf

The `enableIf` option enables the test under some conditions. Example: Just enable the *networkAttackTest* if the vpn is turned on,

```javascript
oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    enableIf: {
        value: !isOffline
    }
});
```

If the condition is not satisfied, the test is not executed. Similar to the `xit` function, it's excluded from execution.

You can use a *boolean* value.

### repeat

The `repeat` option makes the test repeat *n* times, as individuals test. Example: A network attack that attacks something once isn't a real network attack, right? So, do this test 500 times,

```javascript
oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    repeat: {
        value: 500
    }
});
```

Each test is a separate test, and any failure of one will not interfere with the others. If you need the repetitions to behave like a single test, where one failure stops the entire process, then the `repeat` option has a `unifiedTest` parameter. Example:

```javascript
oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    repeat: {
        value: 500,
        args: {
            unifiedTest: true
        }
    }
});
```

With this parameter, if failure occurs, the test will be terminated.

You can use an *integer* value.

### parametersValues

Normally, the test for *Jasmine* has no parameters. What if we could add some parameters? And test various value scenarios for these same parameters? This option allows it. Example: In our scenario of a network attack, we now need to test several values of retry and delay (in ms).

```javascript
oit("networkAttackTest", (retry, delay) => {
    expect(networkAttack(retry, delay)).toBe(true);
}, {
    parametersValues: {
        value: [[1,20],[5,50],[10,200]]
    }
});
```

Note that parameter values are positioned line by line, i.e., the values of each variable are read in sequence for each test set scenario. Instead, you can transpose the values using the `transpose` parameter:

```javascript
oit("networkAttackTest", (retry, delay) => {
    expect(networkAttack(retry, delay)).toBe(true);
}, {
    parametersValues: {
        value: [[1,5,10],[20,50,200]],
        args: {
            transposed: true
        }
    }
});
```

In the example above, all possible values for a variable are given immediately before the next variable.

You can use a two-dimensional array value.

### Mixing options

You can mix several options. The next example shows using `enableIf` and `repeat` in the same test.

```javascript
oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    enableIf: {
        value: !isOffline
    },
    repeat: {
        value: 500
    }
});
```

### Using functions instead of variables values

You can use functions with no arguments or functions with arguments instead of variables in *value*.

Before each case, consider the following example:

```javascript
const isOffline = false;

oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    enableIf: {
        value: !isOffline
    }
});
```

Now, we can use a function without arguments in the value:

```javascript
function isOffline() {
    return false;
}

oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    enableIf: {
        value: isOffline
    }
});
```

Or a function with arguments:

```javascript
const systemPassword = "555";
function isOffline(password) {
    return (systemPassword == password)
}

oit("networkAttackTest", () => {
    expect(networkAttack()).toBe(true);
}, {
    enableIf: {
        value: {
            fn: isOffline,
            args: ["333"]
        }
    }
});
```

# Turning **it** to **oit**

To turn an `it` into an `oit`, just change the function name from `it` to `oit` and add an option object as one of its parameters.

Original test:

```javascript
it("test example", () => {
    // expect
});
```

Modified test:

```javascript
oit("test example", () => {
    // expect
}, {
    enableIf: {
        value: booleanVariable
    }
});
```

So, that's it! Your option object will change the test execution. In the case of parameters option, maybe you need to modify the test, but nothing too complicated.

```javascript
oit("test example", (a, b) => {
    // expect using a and b
}, {
    parametersValues: {
        value: [[2,3],[5,6]]
    }
});
```

# References

E. Soares, M. Ribeiro, R. Gheyi, G. Amaral and A. Santos, "Refactoring Test Smells With JUnit 5: Why Should Developers Keep Up-to-Date?," in IEEE Transactions on Software Engineering, vol. 49, no. 3, pp. 1152-1170, 1 March 2023, doi: 10.1109/TSE.2022.3172654.

### DIFF
--- a/spec/core/OptionsSpec.js
+++ b/spec/core/OptionsSpec.js
@@ -1,0 +1,305 @@
+function isEnabled() {
+  return true;
+}
+
+function isNotEnabled() {
+  return false;
+}
+
+function setEnable(isEnable) {
+  return isEnable;
+}
+
+function tenTimes() {
+  return 10;
+}
+
+function nTimes(n) {
+  return n;
+}
+
+function getDefaultArgs() {
+  return [[2, 3], [4, 4]];
+}
+
+function getRandomArgs(d1) {
+  let returnArray = [];
+  for (let i = 0; i < d1; i++) {
+    let a = Math.floor(Math.random() * 10) + 1;
+    let b = Math.floor(Math.random() * 10) + 1;
+
+    returnArray.push([a, b]);
+  }
+
+  return returnArray;
+}
+
+describe('option tests - ', () => {
+  oit(
+    'repeat 3',
+    () => {
+      let n = Math.random() * 10;
+      if (n < 5) {
+        expect(n).toBeLessThan(5);
+      } else {
+        expect(n).toBeGreaterThanOrEqual(5);
+      }
+    },
+    {
+      repeat: {
+        value: 3
+      }
+    }
+  );
+
+  oit(
+    'enable true',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: true
+      }
+    }
+  );
+
+  oit(
+    'enable false',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: false
+      }
+    }
+  );
+
+  oit(
+    'enable true with function without args',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: isEnabled
+      }
+    }
+  );
+
+  oit(
+    'enable true with function with args',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: {
+          fn: setEnable,
+          args: [true]
+        }
+      }
+    }
+  );
+
+  oit(
+    'enable false with function without args',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: isNotEnabled
+      }
+    }
+  );
+
+  oit(
+    'enable false with function with args',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      enableIf: {
+        value: {
+          fn: setEnable,
+          args: [false]
+        }
+      }
+    }
+  );
+
+  oit(
+    'enable true and repeat 5',
+    () => {
+      let n = Math.random() * 10;
+      if (n < 5) {
+        expect(n).toBeLessThan(5);
+      } else {
+        expect(n).toBeGreaterThanOrEqual(5);
+      }
+    },
+    {
+      enableIf: {
+        value: true
+      },
+      repeat: {
+        value: 5
+      }
+    }
+  );
+
+  oit(
+    'enable false and repeat 5',
+    () => {
+      let n = Math.random() * 10;
+      if (n < 5) {
+        expect(n).toBeLessThan(5);
+      } else {
+        expect(n).toBeGreaterThanOrEqual(5);
+      }
+    },
+    {
+      enableIf: {
+        value: false
+      },
+      repeat: {
+        value: 5
+      }
+    }
+  );
+
+  oit(
+    'function with args',
+    (a, b) => {
+      if (a == b) {
+        expect(a).toBe(b);
+      } else {
+        expect(a).not.toBe(b);
+      }
+    },
+    {
+      parametersValues: {
+        value: [[1, 4], [2, 2], [4, 7]]
+      }
+    }
+  );
+
+  oit(
+    'repeat with function without arguments',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      repeat: {
+        value: tenTimes
+      }
+    }
+  );
+
+  oit(
+    'repeat with functions with arguments',
+    () => {
+      expect(1).toBe(1);
+    },
+    {
+      repeat: {
+        value: {
+          fn: nTimes,
+          args: [10]
+        }
+      }
+    }
+  );
+
+  oit(
+    'parameters with function without args',
+    (a, b) => {
+      if (a != b) {
+        expect(a).not.toBe(b);
+      } else {
+        expect(a).toBe(b);
+      }
+    },
+    {
+      parametersValues: {
+        value: getDefaultArgs
+      }
+    }
+  );
+
+  oit(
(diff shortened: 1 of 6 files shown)


sgravrock:
I appreciate the effort that you put into this, but I think the problems it addresses are already solved at least as well by existing features and idioms.

> Code smell is code that does not follow good software engineering practices

I think [ Wikipedia's definition](https://en.wikipedia.org/wiki/Code_smell) is more in line with common usage:

> In computer programming, a code smell is any characteristic in the source code of a program that possibly indicates a deeper problem.[1][2] Determining what is and is not a code smell is subjective, and varies by language, developer, and development methodology. 

That definition differs from yours in a few important ways, but I particularly want to call out "varies by language". Good Java is not necessarily the same as good JavaScript. Testing frameworks tend to be *very* different due to differences in the two languages' capabilities. As a result, a feature that's a good fit for JUnit may be awkward in Jasmine, or vice versa.

Here's how I'd solve the same problems that `oit` addresses, using existing Jasmine and JavaScript features.

Repeated test:

```
for (let i = 0; i < 500; i++) {
    it(`does something (run ${i})`), function() {
    });
}
```

Parametrized test: https://jasmine.github.io/tutorials/sharing_behaviors

Conditional test:

```
it('returns true for an Error that originated from another frame', function() {
    if (typeof window === 'undefined') {
        pending('This test only runs in browsers.');
    }
    // ...
});
```

Subjective questions of style aside, those approaches all have the advantage of composing well with other Jasmine features. For instance, each technique can be used to create suites as well as individual specs. You can also focus them with `fit`/`fdescribe`, exclude them with `xit`/`xdescribe`, and so on.

> Like code smell, there is test smell, which is applied to test code. The following code is an example of a test smell:
> 
> <code>
> it("testing network", () => {
>    if (!isOffline) {
>        expect(testNetwork()).toBe(true);
>    }
> });
> </code>
>
> In the above test, if the constant isOffline has a true value, then the assert will not be executed. In this case, it gave the impression that it had passed.

You're right about that. But `oit` doesn't solve the problem. Your own tests in this PR show that. If I modify the code in `Options.js` to always run conditional tests regardless of the value of `enableIf.value`, the overall result of the test suite is still success. The same thing happens if I make it never run conditional tests. `oit` is no worse than `pending` in this respect, but it's also no better.


> E. Soares, M. Ribeiro, R. Gheyi, G. Amaral and A. Santos, "Refactoring Test Smells With JUnit 5: Why Should Developers Keep Up-to-Date?," in IEEE Transactions on Software Engineering, vol. 49, no. 3, pp. 1152-1170, 1 March 2023, doi: 10.1109/TSE.2022.3172654.

I didn't read that because it seems to only be available behind a paywall.

Issue #2011, opened 2023-07-17
Fails to parse error without message.


angrycat9000:
### Steps to Reproduce

1. Install `jamine` and `restify-errors`
2. Paste this into a file name `error.spec.js` and run jasmine.
```js
const { UnprocessableEntityError } = require("restify-errors")

it("fails to parse error", () => {
    throw new UnprocessableEntityError();
})
```

### Expected Behavior

Jasmine prints that an error occured and gives the line number of the spec that caused the error. 

> 1) fails to parse error
>   Message:
>     Error thrown
>   Stack:
>         at UserContext.<anonymous> (/home/mdane/repos/jasmine-split-error/spec/support/error.spec.js:4:11)

### Actual Behavior

Jasmine throws an error in its internal code and provides not information about what spec caused the error.

> Failures:
> 1) fails to parse error
>   Message:
>     Unhandled promise rejection: TypeError: Cannot read properties of undefined (reading 'split')
>   Stack:
>         at <Jasmine>
>   Message:
>     Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
>   Stack:
>         at <Jasmine>
>         at listOnTimeout (node:internal/timers:564:17)
>         at process.processTimers (node:internal/timers:507:7)
> 
> 1 spec, 1 failure
> Finished in 5.011 seconds
> Randomized with seed 59104 (jasmine --random=true --seed=59104)

### Example code that reproduces the problem

```JavaScript
https://github.com/angrycat9000/jasmine-split-error
```

### Possible Solution

https://github.com/jasmine/jasmine/pull/2012

### Context

I work on a project with a large number of specs.  Every so often something breaks and throws an error that Jasmine doesn't parse.  It makes it very difficult to track down where the error is happening.

### jasmine-core version

5.0.2

### Node.js or browser version

node 18.13.0

### Operating System

Linux CSM-9Z1WW33 5.19.0-46-generic #47~22.04.1-Ubuntu

PR #2012, opened 2023-07-17
Fix error for undefined message

angrycat9000:
## Description

Check that a message and stack properties exists on an exception before trying to use `split`. 

## Motivation and Context

Fixes https://github.com/jasmine/jasmine/issues/2011 when an error is thrown without a message.

## How Has This Been Tested?
Ran unit tests

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/StackTraceSpec.js
+++ b/spec/core/StackTraceSpec.js
@@ -282,4 +282,16 @@ describe('StackTrace', function() {
 
     expect(result_no_error.message).not.toEqual(jasmine.anything());
   });
+
+  it('handles errors with no message without throwing', function() {
+    const error = {
+      stack:
+        'HttpError [UnprocessableEntityError]\n' +
+        '    at UserContext.<anonymous> (/spec/support/error.spec.js:4:17)\n' +
+        '    at QueueRunner.attempt (/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:7745:40)'
+    };
+    expect(function() {
+      new jasmineUnderTest.StackTrace(error);
+    }).not.toThrowError();
+  });
 });
--- a/spec/core/StackTraceSpec.js
+++ b/spec/core/StackTraceSpec.js
@@ -294,4 +294,11 @@ describe('StackTrace', function() {
       new jasmineUnderTest.StackTrace(error);
     }).not.toThrowError();
   });
+
+  it('handles errors with no stack without throwing', function() {
+    const error = { code: 'UnprocessableEntity', message: '' };
+    expect(function() {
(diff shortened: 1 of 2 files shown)


sgravrock:
I think the problem is a little different than what you thought it was. With a slight modification of the code from your steps to reproduce, you can see that the message isn't undefined, and neither is the stack trace:

```
% node
Welcome to Node.js v18.0.0.
Type ".help" for more information.
> const { UnprocessableEntityError } = require("restify-errors")
undefined
> const e = new UnprocessableEntityError()
undefined
> e.message
''
> e.stack
'UnprocessableEntityError\n' +
  '    at REPL2:1:9\n' +
  '    at Script.runInThisContext (node:vm:129:12)\n' +
  '    at REPLServer.defaultEval (node:repl:571:29)\n' +
  '    at bound (node:domain:421:15)\n' +
  '    at REPLServer.runBound [as eval] (node:domain:432:12)\n' +
  '    at REPLServer.onLine (node:repl:898:10)\n' +
  '    at REPLServer.emit (node:events:539:35)\n' +
  '    at REPLServer.emit (node:domain:475:12)\n' +
  '    at [_onLine] [as _onLine] (node:internal/readline/interface:424:12)\n' +
  '    at [_line] [as _line] (node:internal/readline/interface:885:18)'
```

All of that is fine as far as Jasmine is concerned, and perfectly consistent with other errors. Here's the problem:

```
> e.cause
[Function: we_cause]
```

Jasmine assumes that the `cause` property is an [inner exception](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) and tries to format it as an error. The errors you saw related to `message` and `stack` being undefined were because Jasmine was trying to treat the `we_cause` function as an `Error`. You can also see that by running the example spec from the linked issue using a copy of Jasmine that has your fix:

```
1) fails to parse error
  Message:
    UnprocessableEntityError thrown
  Stack:
    error properties: Object
    UnprocessableEntityError
        at UserContext.<anonymous> ([...]/aSpec.js:4:11)
        at <Jasmine>
    Caused by: Error: undefined
    error properties: Object({ call$0: Function, call$1: Function, call$2: Function, call$3$1: Function, call$2$1: Function, call$1$1: Function, call$3: Function, call$4: Function, call$3$3: Function, call$2$2: Function, call$6: Function, call$5: Function, call$2$0: Function, call$1$2: Function, call$1$0: Function, call$2$3: Function })
```

Note the "Caused by: Error: undefined".

I think a better way to fix this would be to do some type checking on the `cause` property and ignore it if it's not an Error. It also might be a good idea to filter out error properties whose values are functions.

Issue #2008, opened 2023-05-10
Overall test runtime on Safari is slow


apepper:
### Steps to Reproduce

Given one `it` that is focused (`fit`) and 4000 other `it`s that are not focused:

```js
fit('works', () => {
  expect(true).toBe(true);
})

for (let i = 0; i < 4000; i++) {
  it('also works', () => {
    expect(true).toBe(true);
  })
}
```

### Expected Behavior

The overall test suite is "fast" - on my machine executing it on Chrome takes ~1.3 seconds and on Firefox ~0.6 seconds.

### Actual Behavior

The overall test suite is slow. Option 1 (see below) reports "5.889 seconds" and option 2 reports "5.713 seconds" seconds.

### Example code that reproduces the problem

1. Clone https://github.com/apepper/jasmine_slow_fit_on_safari
2. In it: `npm i`
3. * Option 1: `npm run karma` and open http://localhost:9876/ in Safari
    * Option 2: `npm run jasmine-browser` and open  http://localhost:8888 in Safari

### Context

This this both affects running `jasmine-core` with `karma` and with `jasmine-browser-runner` I suspect that the bug is within `jasmine-core` itself.

### jasmine-core version

4.6.0 (latest) - but I originally saw this with 3.9.0

### Versions of other relevant packages

jasmine-browser-runner 1.3.1
karma 6.4.2
karma-jasmine 5.1.0

### Node.js or browser version

node 16 and Safari 16.4

### Operating System

MacOs 13.3 (Ventura)

sgravrock:
This is a bug in Safari. `setTimeout` is very slow in Safari compared to other browsers. Unfortunately Jasmine has to do a `setTimeout` every so often in browsers to allow the garbage collector to run, prevent the browser from killing the tab, and prevent tools like Karma from timing out. The result is that Safari has (on my machine) an unavoidable ~1.5ms per spec overhead that other browsers don't. This is regardless of whether `fit` is used or not.

We've tried a variety alternatives to the current scheme of doing a `setTimeout` on every 10th call to `clearStack`, and they've all introduced serious stability problems.

[...]

PR #2034, opened 2024-07-25
Improve performance on Playwright Windows WebKit

m-akinc:
## Description
1. (primary change) Expanded the user agent string regex used to identify Safari browsers so that it also covers the Windows WebKit browser used by Playwright. There is an existing code comment that explains that `queueMicrotask` should be used on Safari because it is "dramatically faster" than `MessageChannel`. The same is true for the Windows WebKit browser.
2. (minor change) Update `glob` call in test setup code to work on Windows machines.

## Motivation and Context
Running tests on Windows against the WebKit browser installed by Playwright is an order of magnitude slower than running the same tests on other browsers. This is especially evident when focusing only a small number of tests, as skipped tests still take a fraction of a second each (on WebKit).

## How Has This Been Tested?
Ran all tests in this repo. Ran my repo's tests with this modification on Windows WebKit and saw huge improvement in performance.

## Types of changes

## Checklist:

### DIFF
--- a/spec/helpers/nodeDefineJasmineUnderTest.js
+++ b/spec/helpers/nodeDefineJasmineUnderTest.js
@@ -16,7 +16,9 @@
       return path.join(__dirname, '../../', 'src/', file);
     });
 
-    const files = src_files.flatMap(g => glob.sync(g));
+    const files = src_files.flatMap(g =>
+      glob.sync(g, { windowsPathsNoEscape: true })
+    );
     files.forEach(function(resolvedFile) {
       require(resolvedFile);
     });
--- a/src/core/ClearStack.js
+++ b/src/core/ClearStack.js
@@ -68,8 +68,8 @@ getJasmineRequireObj().clearStack = function(j$) {
       global.process.versions &&
       typeof global.process.versions.node === 'string';
 
-    // Windows builds of WebKit have a fairly generic user agent string when no application name is provided.
-    // See: https://github.com/WebKit/WebKit/blob/d898a3cffd9c992980016cb1fbdba272cb0c992d/Source/WebCore/platform/win/UserAgentWin.cpp#L37
+    // Windows builds of WebKit have a fairly generic user agent string when no application name is provided:
+    // e.g. "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/605.1.15 (KHTML, like Gecko)"
     const SAFARI_OR_WIN_WEBKIT =
       global.navigator &&
       /(^((?!chrome|android).)*safari)|(Win64; x64\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\)$)/i.test(
@@ -84,8 +84,9 @@ getJasmineRequireObj().clearStack = function(j$) {
       SAFARI_OR_WIN_WEBKIT ||
       j$.util.isUndefined(global.MessageChannel) /* tests */
     ) {
-      // queueMicrotask is dramatically faster than MessageChannel in Safari,
-      // at least through version 16.
+      // queueMicrotask is dramatically faster than MessageChannel in Safari
+      // and other WebKit-based browsers, such as the one distributed by Playwright
+      // to test Safari-like behavior on Windows.
       // Some of our own integration tests provide a mock queueMicrotask in all
       // environments because it's simpler to mock than MessageChannel.
       return browserQueueMicrotaskImpl(global);
--- a/src/core/ClearStack.js
+++ b/src/core/ClearStack.js
@@ -70,7 +70,9 @@ getJasmineRequireObj().clearStack = function(j$) {
 
     const SAFARI =
       global.navigator &&
-      /^((?!chrome|android).)*safari/i.test(global.navigator.userAgent);
+      /(^((?!chrome|android).)*safari)|(^((?!chrome|android|firefox).)+$)/i.test(
+        global.navigator.userAgent
+      );
 
     if (NODE_JS) {
       // Unlike browsers, Node doesn't require us to do a periodic setTimeout
--- a/src/core/ClearStack.js
+++ b/src/core/ClearStack.js
@@ -68,9 +68,11 @@ getJasmineRequireObj().clearStack = function(j$) {
       global.process.versions &&
       typeof global.process.versions.node === 'string';
 
-    const SAFARI =
+    // Windows builds of WebKit have a fairly generic user agent string when no application name is provided.
+    // See: https://github.com/WebKit/WebKit/blob/d898a3cffd9c992980016cb1fbdba272cb0c992d/Source/WebCore/platform/win/UserAgentWin.cpp#L37
+    const SAFARI_OR_WIN_WEBKIT =
       global.navigator &&
-      /(^((?!chrome|android).)*safari)|(^((?!chrome|android|firefox).)+$)/i.test(
+      /(^((?!chrome|android).)*safari)|(Win64; x64\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\)$)/i.test(
         global.navigator.userAgent
       );
 
@@ -79,7 +81,7 @@ getJasmineRequireObj().clearStack = function(j$) {
       // so we avoid the overhead.
       return nodeQueueMicrotaskImpl(global);
     } else if (
-      SAFARI ||
+      SAFARI_OR_WIN_WEBKIT ||
       j$.util.isUndefined(global.MessageChannel) /* tests */
     ) {
       // queueMicrotask is dramatically faster than MessageChannel in Safari,
--- a/spec/core/ClearStackSpec.js
+++ b/spec/core/ClearStackSpec.js
@@ -22,6 +22,19 @@ describe('ClearStack', function() {
     });
   });
 
+  describe('in WebKit (Playwright\'s build for Windows)', function() {
+    usesQueueMicrotaskWithSetTimeout(function() {
+      return {
+        navigator: {
+          userAgent:
+            'Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/605.1.15 (KHTML, like Gecko)'
+        },
+        // queueMicrotask should be used even though MessageChannel is present
+        MessageChannel: fakeMessageChannel
+      };
+    });
+  });
+
   describe('in browsers other than Safari', function() {
     usesMessageChannel(function() {
       return {
--- a/spec/core/ClearStackSpec.js
+++ b/spec/core/ClearStackSpec.js
@@ -22,7 +22,7 @@ describe('ClearStack', function() {
     });
   });
 
-  describe('in WebKit (Playwright\'s build for Windows)', function() {
+  describe("in WebKit (Playwright's build for Windows)", function() {
     usesQueueMicrotaskWithSetTimeout(function() {
       return {
         navigator: {


src/core/ClearStack.js
```diff
@@ -70,7 +70,9 @@ getJasmineRequireObj().clearStack = function(j$) {
 
     const SAFARI =
       global.navigator &&
-      /^((?!chrome|android).)*safari/i.test(global.navigator.userAgent);
+      /(^((?!chrome|android).)*safari)|(^((?!chrome|android|firefox).)+$)/i.test(
+        global.navigator.userAgent
+      );
 
     if (NODE_JS) {
```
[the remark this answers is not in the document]
sgravrock: > Is there more detail on why/if running tests on browsers really can't use the more lightweight nodeQueueMicrotaskImpl?

The short answer is that we've tried not doing a setTimeout in browsers several times and it ended badly each time. https://github.com/jasmine/jasmine/issues/2008 was the last attempt. There are a few reasons:

* Browsers don't allow JS to run forever without yielding control. When a site does that, most browsers will eventually decide that it's caught in an infinite loop and kill the tab.
* Even if they did, tools like Karma conclude that the test runner has crashed if it goes too long without producing any output.
* Some browsers will defer or minimize garbage collection if the running code doesn't go idle enough. This is probably why previous attempts to optimize this have caused major slowdowns for suites with a lot of synchronous specs. 

Those problems mainly show up in suites with a lot of synchronous specs, or even a few heavyweight synchronous specs that happen to get scheduled back to back. You won't see them if all of your specs contain async or if you're only running a few focused specs.

src/core/ClearStack.js
```diff
@@ -70,7 +70,9 @@ getJasmineRequireObj().clearStack = function(j$) {
 
     const SAFARI =
       global.navigator &&
-      /^((?!chrome|android).)*safari/i.test(global.navigator.userAgent);
+      /(^((?!chrome|android).)*safari)|(^((?!chrome|android|firefox).)+$)/i.test(
```
sgravrock, line 73: If I'm reading it correctly, the added bit of regex says that every user agent string that doesn't include "chrome", "android", or "firefox" should get the Safari treatment. I think that's too broad, especially since the Safari mode is known to be slower in other browsers. I think the regex should specifically target the Playwright Webkit user agent string.

Additionally:

* This needs a unit test, like the other tests in ClearStackSpec.js. Otherwise it's very likely that your change will be broken by the next person who touches this code.
* The name of the SAFARI constants should be updated.

spec/helpers/nodeDefineJasmineUnderTest.js
```diff
@@ -16,7 +16,9 @@
       return path.join(__dirname, '../../', 'src/', file);
     });
 
-    const files = src_files.flatMap(g => glob.sync(g));
+    const files = src_files.flatMap(g =>
+      glob.sync(g, { windowsPathsNoEscape: true })
```
sgravrock, line 20: Can you tell me more about what problem this solves? The tests work on my Windows machine without this change. And none of the globs contain backslashes. In any case, `windowsPathsNoEscape` has its share of rough edges and I'd rather find another way to solve the problem.

sgravrock (changes requested):
Before you spend any more time on this, I should point out that Playwright Webkit is not a supported browser. (For that to change, I'd need to see significant user interest and have a way to test against it in CI.) I don't mind merging fixes for unsupported browsers as long as they don't make anything worse. But be aware that Jasmine isn't tested against Playwright Webkit, so there's always a risk of things breaking in the future.

src/core/ClearStack.js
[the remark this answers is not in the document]
m-akinc: Has there been any attempt at an elapsed time based approach, rather than something like "every 10 calls"? I.e. only call setTimeout if it has been at least 500ms since the last setTimeout. I can't imagine something like that would significantly increase the chance of getting killed by the browser, and it would probably be a big win when there are lots of short or skipped tests. I'd be happy to put up a PR for something like that if there's interest.

src/core/ClearStack.js
m-akinc: I'll add a test and update the variable name. Further scoping the regex may not be possible, unfortunately, as the user agent string is almost identical to Chrome's, minus the "Chrome" part at the end (if I recall correctly). I can dig a little more and see if I can find another reliable way to distinguish it, though.

src/core/ClearStack.js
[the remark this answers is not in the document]
sgravrock: Yes, and it dramatically slowed down mostly-synchronous test suites including jasmine-core's. It could also be enough to cause some test suites to run out of memory.

More generally, getting this right is _hard_. It took years to stabilize the between-specs stack clearing process so that nobody's test suite crashed their browser, crashed Karma, ran out of memory or ran unreasonably slow. Those are the worst kind of bugs: hard to find before release, completely breaks Jasmine for some people, and the affected users often can't provide steps to reproduce the problem. 

The existing approach looks crude but it's been running since 2017 with no complaints except for this one and the issue I linked to above. I'm not completely unwilling to change it, but the bar to demonstrating that a change is safe is quite high. We'd need reason to be confident that the change wouldn't break Jasmine for anyone, regardless of what their test suite is shaped like. That's hard to come by.

dcsaszar:
FYI revisiting #2008 because of the cross link I ran a test with 9000 skipped and 2 focused tests:

| Browser  | Karma time (s) |
|---|---|
| Safari  | 53.915  |
| Webkit  |  13.853 |
| WebkitHeadless  | 1.667  |
| Chrome  | 0.194  |
| ChromeHeadless  |  0.191 |

That is, on a Mac headless Webkit is way faster than Webkit, although there's still a big difference to the fast browsers. 
@m-akinc:
Do you see similar results on Windows, too?
How many specs are we talking about in your case?

m-akinc:
> @m-akinc: Do you see similar results on Windows, too? How many specs are we talking about in your case?

Here are my timings for 5000 skipped tests and 1 focused:

| Browser | Karma time (s) |
|----------|-----------------|
| Webkit | 228 |
| WebkitHeadless | 225 |
| Chrome | 14 |
| ChromeHeadless | 14 |
| Firefox | 11 |

About 6s of each of those is startup/webpack before the browser is launched.

spec/helpers/nodeDefineJasmineUnderTest.js
m-akinc: `path.join()` returns paths using backslashes, so the result of:
```ts
  const src_files = ['core/**/*.js', 'version.js'].map(function(file) {
    return path.join(__dirname, '../../', 'src/', file);
  });
```
is 
```ts
  'C:\dev\jasmine\src\core\**\*.js',
  'C:\dev\jasmine\src\version.js'
```
`glob` [doesn't treat backslashes as separators by default](https://www.npmjs.com/package/glob#windows), so the result of globbing those paths are `[]` and `[ 'src\\version.js' ]`, respectively. The error reported is this:
```
C:\dev\jasmine> npm test

> jasmine-core@5.2.0 test
> grunt --stack execSpecsInNode

Running "execSpecsInNode" task
Error: Cannot find module 'src\version.js'
Require stack:
- C:\dev\jasmine\spec\helpers\nodeDefineJasmineUnderTest.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at C:\dev\jasmine\spec\helpers\nodeDefineJasmineUnderTest.js:24:7
    at Array.forEach (<anonymous>)
    at getSourceFiles (C:\dev\jasmine\spec\helpers\nodeDefineJasmineUnderTest.js:23:11)
    at C:\dev\jasmine\spec\helpers\nodeDefineJasmineUnderTest.js:28:3
    at Object.<anonymous> (C:\dev\jasmine\spec\helpers\nodeDefineJasmineUnderTest.js:32:3)
    at Module._compile (node:internal/modules/cjs/loader:1358:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'C:\\dev\\jasmine\\spec\\helpers\\nodeDefineJasmineUnderTest.js' ]
}
Warning: Task "execSpecsInNode" failed. Use --force to continue.
Error: Task "execSpecsInNode" failed.
    at Task.<anonymous> (C:\dev\jasmine\node_modules\grunt\lib\util\task.js:198:15)
    at Timeout._onTimeout (C:\dev\jasmine\node_modules\grunt\lib\util\task.js:234:33)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)

Aborted due to warnings.
```

src/core/ClearStack.js
m-akinc: I changed my added regex pattern so that it matches any user agent string that ends with `Win64; x64) AppleWebKit/[0-9.]+ (KHTML, like Gecko)`. Does that seem acceptably specific? As I mentioned before, the string I'm trying to match is pretty generic: `Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/605.1.15 (KHTML, like Gecko)`.

dcsaszar:
I had a look why my results differ that much between Webkit and WebkitHeadless, while yours don't. On my machine the two report different user agents, triggering different code paths.

### Webkit 
* UA: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.4 Safari/605.1.15`
* SAFARI=true
* `queueMicrotask`
* slow

### WebkitHeadless
* UA: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
* SAFARI=false
* `MessageChannel`
* fast

It looks like while
> queueMicrotask should be used on Safari because it is "dramatically faster" than MessageChannel. The same is true for the Windows WebKit browser.

the opposite appears to be the case for MacOS WebKit.

Maybe the best way out of this dilemma is to let the user configure the clear-stack implementation.

m-akinc:
> * potentially comes with a caveat of introducing [Overall test runtime on Safari is slow #2008](https://github.com/jasmine/jasmine/issues/2008) for focused tests

I've changed the regex I was using so that the only change is that it includes the Webkit build for Windows, that way there will be zero impact on Mac browsers.

spec/helpers/nodeDefineJasmineUnderTest.js
sgravrock: Neat. The existing code works when the repo is checked out under c:\Users but not when it's checked out under c:\dev. I would not have expected that to make a difference. 

[
I've pushed a commit to main that fixes this without using windowsPathNoEscape](https://github.com/jasmine/jasmine/commit/3c4b73f136d0b7b10f49c2dfbe38b8e8e119260a).

src/core/ClearStack.js
```diff
@@ -68,16 +68,20 @@ getJasmineRequireObj().clearStack = function(j$) {
       global.process.versions &&
       typeof global.process.versions.node === 'string';
 
-    const SAFARI =
+    // Windows builds of WebKit have a fairly generic user agent string when no application name is provided.
+    // See: https://github.com/WebKit/WebKit/blob/d898a3cffd9c992980016cb1fbdba272cb0c992d/Source/WebCore/platform/win/UserAgentWin.cpp#L37
```
sgravrock, line 72: Mind replacing that link with a concrete example of the user agent string?

src/core/ClearStack.js
```diff
@@ -68,16 +68,20 @@ getJasmineRequireObj().clearStack = function(j$) {
...
     if (NODE_JS) {
       // Unlike browsers, Node doesn't require us to do a periodic setTimeout
       // so we avoid the overhead.
       return nodeQueueMicrotaskImpl(global);
     } else if (
-      SAFARI ||
+      SAFARI_OR_WIN_WEBKIT ||
       j$.util.isUndefined(global.MessageChannel) /* tests */
     ) {
       // queueMicrotask is dramatically faster than MessageChannel in Safari,
```
sgravrock, line 87: This comment could use an update.

PR #2035, opened 2024-08-11
Use queueMicrotask in clearStack

### LINKED ISSUE #1835: Consider queueMicrotask in clearStack
Right now, `MessageChannel/setTimeout` is used in browsers to clear stack (see https://github.com/jasmine/jasmine/blob/main/src/core/ClearStack.js#L41).

I tried to use Jasmine for running tests in the invisible iframe, and some browsers (Chrome namely) throttle such operations for out-of-view iframes. Brought me to something like 1 sec delay between tests. So tests took too long. 

I suggest to use `queueMicrotask` as a top-priority variant if available. Supported everywhere excepts IE. No throttling. Modern stuff.  The code is exactly the same as `setTimeoutImpl`, just use `queueMicrotask`.

### LINKED ISSUE #2008: Overall test runtime on Safari is slow

dcsaszar:
As discussed in #1835, `queueMicrotask` is faster than `MessageChannel` in most situations if `setTimeout` clamping is taken care of. This can be done by interleaving `setTimeout` and `MessageChannel`.

## Description
* All browsers use the same `clearStack` implementation
* always call `queueMicrotask`
* every 10 `queueMicrotask` calls: call `setTimeout` (clears the stack)
* every 4 `setTimeout` calls: call `MessageChannel` (prevents clamping `setTimeout` to 4 ms)

## Motivation and Context
Fixes #1835
Fixes #2008

## How Has This Been Tested?
Ran on an existing ~10 K tests suite in five (mostly headless) browsers. See analysis: https://github.com/jasmine/jasmine/issues/1835#issuecomment-2254546734

## Types of changes

#### Not really breaking, but
Because of the changed run-time behavior of spec suites this may change the outcome of specs in race cases.
However, since Safari already uses `queueMicrotask`, suites that were green in Safari are very likely to be green in other browsers.

#### Performance
* Test suites may run slower in Firefox unless `dom.clamp.timeout.nesting.level` is set to `-1`. 
* The Jasmine suite `npm run serve` runs slower in headful Chrome. This is because we create fresh instances of `clearStack`, which resets the `currentSetTimeoutCallCount` to zero, missing `setTimeout` calls by previous specs, ultimately leading to browser `setTimeout` clamping. The full speed can be verified with `let currentSetTimeoutCallCount = maxSetTimeoutWithoutClampingCallCount` (though this will fail the respective test).

## Checklist:

### DIFF
--- a/spec/core/ClearStackSpec.js
+++ b/spec/core/ClearStackSpec.js
@@ -56,12 +56,7 @@ describe('ClearStack', function() {
 
     describe('when MessageChannel is unavailable', function() {
       usesQueueMicrotaskWithSetTimeout(function() {
-        return {
-          navigator: {
-            userAgent: 'CERN-LineMode/2.15 libwww/2.17b3',
-            MessageChannel: undefined
-          }
-        };
+        return { MessageChannel: undefined };
       });
     });
   });
--- a/spec/core/ClearStackSpec.js
+++ b/spec/core/ClearStackSpec.js
@@ -21,6 +21,39 @@ describe('ClearStack', function() {
       };
     });
 
+    it('uses MessageChannel every four setTimeout calls to avoid timeout clamping', function() {
+      const fakeChannel = fakeMessageChannel();
+      spyOn(fakeChannel.port2, 'postMessage');
+      const setTimeout = jasmine
+        .createSpy('setTimeout')
+        .and.callFake(function(fn) {
+          fn();
+        });
+      const global = {
+        queueMicrotask: function(fn) {
+          fn();
+        },
+        setTimeout,
+        MessageChannel: function() {
+          return fakeChannel;
+        }
+      };
+      const clearStack = jasmineUnderTest.getClearStack(global);
+
+      for (let i = 0; i < 39; i++) clearStack(function() {});
+
+      expect(setTimeout).toHaveBeenCalledTimes(3);
+      expect(fakeChannel.port2.postMessage).not.toHaveBeenCalled();
+
+      clearStack(function() {});
+      expect(setTimeout).toHaveBeenCalledTimes(4);
+      expect(fakeChannel.port2.postMessage).toHaveBeenCalledTimes(1);
+
+      for (let i = 0; i < 39; i++) clearStack(function() {});
+      expect(setTimeout).toHaveBeenCalledTimes(7);
+      expect(fakeChannel.port2.postMessage).toHaveBeenCalledTimes(1);
+    });
+
     describe('when MessageChannel is unavailable', function() {
       usesQueueMicrotaskWithSetTimeout(function() {
         return {
@@ -44,70 +77,6 @@ describe('ClearStack', function() {
       };
     });
   });
-
-  function usesMessageChannel(makeGlobal) {
-    it('uses MessageChannel', function() {
-      const global = {
-        ...makeGlobal(),
-        MessageChannel: fakeMessageChannel
-      };
-      const clearStack = jasmineUnderTest.getClearStack(global);
-      let called = false;
-
-      clearStack(function() {
-        called = true;
-      });
-
-      expect(called).toBe(true);
-    });
-
-    it('uses setTimeout instead of MessageChannel every 10 calls to make sure we release the CPU', function() {
-      const fakeChannel = fakeMessageChannel();
-      spyOn(fakeChannel.port2, 'postMessage');
-      const setTimeout = jasmine.createSpy('setTimeout');
-      const global = {
-        ...makeGlobal(),
-        setTimeout,
-        MessageChannel: function() {
-          return fakeChannel;
-        }
-      };
-      const clearStack = jasmineUnderTest.getClearStack(global);
-
-      for (let i = 0; i < 9; i++) {
-        clearStack(function() {});
-      }
-
-      expect(fakeChannel.port2.postMessage).toHaveBeenCalled();
-      expect(setTimeout).not.toHaveBeenCalled();
-
-      clearStack(function() {});
-      expect(fakeChannel.port2.postMessage).toHaveBeenCalledTimes(9);
-      expect(setTimeout).toHaveBeenCalledTimes(1);
-
-      clearStack(function() {});
-      expect(fakeChannel.port2.postMessage).toHaveBeenCalledTimes(10);
-      expect(setTimeout).toHaveBeenCalledTimes(1);
-    });
-
-    it('calls setTimeout when onmessage is called recursively', function() {
-      const setTimeout = jasmine.createSpy('setTimeout');
-      const global = {
-        ...makeGlobal(),
-        setTimeout,
-        MessageChannel: fakeMessageChannel
-      };
-      const clearStack = jasmineUnderTest.getClearStack(global);
-      const fn = jasmine.createSpy('second clearStack function');
-
-      clearStack(function() {
-        clearStack(fn);
-      });
-
-      expect(fn).not.toHaveBeenCalled();
-      expect(setTimeout).toHaveBeenCalledWith(fn, 0);
-    });
-  }
 
   function usesQueueMicrotaskWithSetTimeout(makeGlobal) {
     it('uses queueMicrotask', function() {
--- a/spec/core/ClearStackSpec.js
+++ b/spec/core/ClearStackSpec.js
@@ -9,20 +9,7 @@ describe('ClearStack', function() {
     });
   });
 
-  describe('in Safari', function() {
-    usesQueueMicrotaskWithSetTimeout(function() {
-      return {
-        navigator: {
-          userAgent:
-            'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.0.8 (KHTML, like Gecko) Version/15.1 Safari/605.0.8'
-        },
-        // queueMicrotask should be used even though MessageChannel is present
-        MessageChannel: fakeMessageChannel
-      };
-    });
-  });
-
-  describe("in WebKit (Playwright's build for Windows)", function() {
+  describe('in browsers', function() {
     usesQueueMicrotaskWithSetTimeout(function() {
       return {
         navigator: {
@@ -31,18 +18,6 @@ describe('ClearStack', function() {
         },
         // queueMicrotask should be used even though MessageChannel is present
         MessageChannel: fakeMessageChannel
-      };
-    });
-  });
-
-  describe('in browsers other than Safari', function() {
-    usesMessageChannel(function() {
-      return {
-        navigator: {
-          // Chrome's user agent string contains "Safari" so it's a good test case
-          userAgent:
-            'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36'
-        }
       };
     });
 
--- a/src/core/ClearStack.js
+++ b/src/core/ClearStack.js
@@ -68,32 +68,15 @@ getJasmineRequireObj().clearStack = function(j$) {
       global.process.versions &&
       typeof global.process.versions.node === 'string';
 
-    // Windows builds of WebKit have a fairly generic user agent string when no application name is provided:
-    // e.g. "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/605.1.15 (KHTML, like Gecko)"
-    const SAFARI_OR_WIN_WEBKIT =
-      global.navigator &&
-      /(^((?!chrome|android).)*safari)|(Win64; x64\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\)$)/i.test(
-        global.navigator.userAgent
-      );
-
     if (NODE_JS) {
       // Unlike browsers, Node doesn't require us to do a periodic setTimeout
       // so we avoid the overhead.
       return nodeQueueMicrotaskImpl(global);
-    } else if (
-      SAFARI_OR_WIN_WEBKIT ||
-      j$.util.isUndefined(global.MessageChannel) /* tests */
-    ) {
-      // queueMicrotask is dramatically faster than MessageChannel in Safari
-      // and other WebKit-based browsers, such as the one distributed by Playwright
-      // to test Safari-like behavior on Windows.
-      // Some of our own integration tests provide a mock queueMicrotask in all
-      // environments because it's simpler to mock than MessageChannel.
-      return browserQueueMicrotaskImpl(global);
     } else {
-      // MessageChannel is faster than queueMicrotask in supported browsers
-      // other than Safari.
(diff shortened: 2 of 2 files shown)


sgravrock:
I haven't been able to reproduce your results. I'm seeing gains, but nothing like what you linked to. Can you share a link to a repo that will let me easily run the same experiment? I see that you were using Karma. That might matter. The exact details of your Karma configuration might matter. With the kind of test suite I think you're running, reporter performance could be a very significant factor.

> Test suites may run slower in Firefox unless dom.clamp.timeout.nesting.level is set to -1.

It's not reasonable to expect users to reconfigure their browser. That's especially true when we're talking about a config property that's undocumented as far as I can tell. I won't knowingly ship a performance regression in the default configuration of a supported browser.

> The Jasmine suite npm run serve runs slower in headful Chrome. This is because we create fresh instances of clearStack, which resets the currentSetTimeoutCallCount to zero, missing setTimeout calls by previous specs, ultimately leading to browser setTimeout clamping. The full speed can be verified with let currentSetTimeoutCallCount = maxSetTimeoutWithoutClampingCallCount (though this will fail the respective test).

I'll need to look into that more. I'm seeing worse results -- a 2.2-2.5x slowdown in Chrome and Firefox. That's probably acceptable if it only affects Jasmine's own tests, but I'd want to be very sure of that.

This is interesting, and we might be able to use this approach in browsers where we're confident it will be faster, but it sounds like it's not quite a one size fits all solution.

sgravrock:
I'm also seeing a *much* bigger performance difference between Firefox and other browsers than your data shows, both before and after your change. That also makes me think there's an important difference between your test scenario and mine.

dcsaszar:
> Can you share a link to a repo that will let me easily run the same experiment?

Here you are: https://github.com/dcsaszar/jasmine_slow_fit_on_safari

Some results:

https://plotly.com/~dcsaszar/104/
image


> It's not reasonable to expect users to reconfigure their browser.

Makes sense. Luckily I cannot reproduce a significant difference with the new test suite (see `noclamp` items).

> I'll need to look into that more. I'm seeing worse results -- a 2.2-2.5x slowdown in Chrome and Firefox. That's probably acceptable if it only affects Jasmine's own tests, but I'd want to be very sure of that.

The new test suite shows a 4 % slowdown in one area: Jasmine browser runner Firefox and HeadlessFirefox. 

I've done some measurements with parts of the Angular test suite. There's one significant (up to 25 %) slowdown in `core/test:test_web`. A code search indicates this may be caused by tests using `setTimeout`. This is a general issue with the simple approach of this PR: Since we only count our own calls to `setTimeout` we may be too late to avoid clamping.

https://plotly.com/~dcsaszar/102/
image


> I'm also seeing a much bigger performance difference between Firefox and other browsers than your data shows, both before and after your change. That also makes me think there's an important difference between your test scenario and mine.

I think we can ignore this aspect for now. Ideas:
* In the previous measurements I've only used Karma, because that's what we use in our private repo.
* A different outcome in my previous measurements could also be caused by the fact that we're still on jasmine 4.

triskweline:
For some reason this change causes a performance regression in our test suite. We went from `[105.779s, 107.627s]` (Jasmine 5.2) to `[153.324s, 153.996s]` (this change).

- Repo [unpoly/unpoly](https://github.com/unpoly/unpoly)
- 3013 specs
- Chrome 127 (not headless)
- Linux

triskweline:
> For some reason this change causes a performance regression in our test suite. 

Unpoly's test suite uses `setTimeout()` extensively in its own specs. This may play into this.

sgravrock:
> Unpoly's test suite uses setTimeout() extensively in its own specs.

I suspect that's a lot more common than it used to be. For instance, `setTimeout`/`setInterval` based polling is a standard technique for testing UI components that have async behavior.

triskweline:
> I suspect that's a lot more common than it used to be. For instance, setTimeout/setInterval based polling is a standard technique for testing UI components that have async behavior.

I agree 100%! Even when you mock everything (network, clock, etc.), a single `await` in the tested code requires you to delay expectations to be able to observe an effect. And because you cannot reason about the maximum length of a promise chain involved, waiting for a full task with `setTimeout()` is the only way to be sure that everything async has played out.

dcsaszar:
Quick feedback:
* Thanks @triskweline for a real-life test environment. This is very useful and I can reproduce your findings.
* I agree that we must expect suites to use `setTimeout`. Thus we have to handle clamping caused by a test suite.
* For a reasonably fast execution the simple approach implemented here is obviously not good enough.
* I'd like to experiment with an adaptive approach (first mentioned here https://github.com/jasmine/jasmine/issues/1835#issuecomment-2254546734), but I don't know yet how it will look like. What I have on my mind right now is to measure the `performance.now()` between `setTimeout(cb)` and the `cb()` call to conditionally trigger some clamping cooldown. Other suggestions?

sgravrock:
Before I respond to specifics, I want to ask a big picture question: What's the problem we're trying to solve here? Is it:

* Test suite performance in browsers generally
* Performance of large suites with a few focused specs in Safari
* Performance of large mostly synchronous suites in Safari

Of course it would be good to improve all of those. But would doing major surgery on clearStack still be worth the effort and risk if running a single focused spec was reasonably fast in all browsers? Although the Safari perf problem is [not specific to focusing](https://github.com/jasmine/jasmine/issues/2008), the complaints I've heard have all been in the context of focusing one or a few specs. That makes me think that if focusing was fast, approximately nobody would notice the general perf problem in safari.

sgravrock:
>     * I'd like to experiment with an adaptive approach (first mentioned here [Consider queueMicrotask in clearStack #1835 (comment)](https://github.com/jasmine/jasmine/issues/1835#issuecomment-2254546734)), but I don't know yet how it will look like. What I have on my mind right now is to measure the `performance.now()` between `setTimeout(cb)` and the `cb()` call to conditionally trigger some clamping cooldown.

That seems like it would be worth experimenting with. How would the clamping cooldown be implemented? Are you thinking of skipping `setTimeout` or doing it with a longer delay? My knee-jerk reaction is that the latter sounds less risky.

sgravrock:
> Here you are: https://github.com/dcsaszar/jasmine_slow_fit_on_safari
This is an artificial test suite where the number of tests and the test duration resembles our suite (10 k tests in 10 s).

The realistic-ness of that is good, but it also muddies the waters a bit. I think it's worth running it both with and without the DOM manipulation in the test. Running in jasmine-browser-runner with the DOM manipulation, I see:

* Chrome: around 19 seconds
* Safari: around 19 seconds
* Firefox: wildly variable.

Without the DOM manipulation:

* Chrome: 2.7 seconds
* Safari: around 15 seconds
* Firefox: 0.5 seconds

Also, if I run that second scenario using Karma, it takes Karma about another 4 seconds to actually report the suite result after Firefox is done. (That's using the built in OS X terminal emulator, which isn't the fastest but is also by no means the slowest.) So that probably sets a limit on how much we can improve things for Karma users in non-focused scenarios.

> A different outcome in my previous measurements could also be caused by the fact that we're still on jasmine 4.

I think that's probably a red herring. `ClearStack` hasn't changed since 4.5.0 (other than https://github.com/jasmine/jasmine/pull/2034 from @m-akinc , which hasn't shipped yet) and the other relevant code hasn't changed much in recent years either.

sgravrock:
One other thing to be aware of if your real application is Angular: [Zone.js heavily monkey patches Jasmine](https://github.com/angular/angular/blob/main/packages/zone.js/lib/jasmine/jasmine.ts), including replacing some of the core "move on to the next spec" code. I don't plan to test against that. (My willingness to do extra work in my free time because Google decided to monkey patch half the world is very limited.) But if you're running Angular, you might want to.

triskweline:
> For some reason this change causes a performance regression in our test suite. We went from [105.779s, 107.627s] (Jasmine 5.2) to [153.324s, 153.996s] (this change).

I looked into this some more. Since Unpoly's tests very carefully clean up any async effects before ending a spec, I wondered if I could get away with not clearing the stack at all:

```js
function clearStack(fn) {
  fn()
}
```

To my surprise actually *increased* test runtime by 53%. How can `postMessage()` be faster than not doing any work at all?

I think this is what happens:

- Most of Unpoly's specs use 4+ `setTimeout()` calls. That means the next spec will always start with clamped timeouts, unless something resets the timeout nesting.
- Jasmine's existing `clearStack()` schedules with `postMessage()` 10 out of 11 times. I think this resets any clamping nesting, so the next spec can start without clamping.
- By removing `postMessage()` with the sync implementation above, the nesting was no longer reset and clamping was active all of the time.
- This PR keeps `postMessage()`, but only calls it 1 out of 40 (?) times. So clamping is active most of the time.

triskweline:
> That makes me think that if focusing was fast, approximately nobody would notice the general perf problem in safari.

What is the reason we're calling `clearStack()` when we're skipping a non-focused spec?

When I remove this, skipping specs becomes instantaneous for me on Safari and Chrome.

sgravrock:
> What is the reason we're calling clearStack() when we're skipping a non-focused spec?

Reporter dispatching. Jasmine makes at least two calls to each reporter for each spec. Most reporters are quick and synchronous but they can be slow and/or async, just like specs. Reporter calls are dispatched through `QueueRunner`, and therefore through `ClearStack`, just like other user-provided functions.

sgravrock:
One way around that would be to not "run" non-focused specs at all, and instead simply report a single dummy spec indicating that other specs were skipped. [That would have to be configurable because it could have some nasty side effects](https://github.com/jasmine/jasmine/issues/2008#issuecomment-1565611824), but it would probably better for most people most of the time.

triskweline:
Could we detect async reporter callbacks, by checking if they either accept a `done` argument or return a promise?
This way we could skip `clearStack()` if we know the last job was sync, like with the default `HtmlReporter`.

Sorry if I'm bringing up ideas that have already been considered elsewhere. I'm getting into this part of Jasmine for the first time.

sgravrock:
It's possible to detect async (albeit not necessarily `setTimeout` itself) after a fn returns. QueueRunner actually does that already: it looks at the fn signature and return value to determine whether it was synchronous, and [does an explicit setTimeout if it was](https://github.com/jasmine/jasmine/commit/578f63b9bd34f75074a00b9a404b48d567dbe180#diff-a7a80173e8395a81054aa4df41b588f91379e81d085e356da4d15145e4011c72). That's probably an important piece of the clearStack performance puzzle and I wish I'd remembered it earlier.

sgravrock:
Jasmine constructs a QueueRunner instance for each level of the hierarchy, and also to dispatch each reporter call. So for e.g. two specs in a single suite, with any number of reporters, it constructs:

* A QueueRunner for the suite, which runs:
    * A QueueRunner to dispatch suiteStarted to all reporters
    * Any beforeAlls
    * For each of the suite's specs, a separate QueuRunner to run:
        * A QueueRunner to dispatch specStarted to all reporters
        * Any beforeEaches
        * The spec fn itself
        * Any afterEaches
        * A QueueRunner to dispatch specDone to all reporters
    * Any afterAlls
    * A QueueRunner to dispatch suiteDone to all reporters

dcsaszar:
> I want to ask a big picture question: What's the problem we're trying to solve here? Is it:
> 
> * Test suite performance in browsers generally
> * Performance of large suites with a few focused specs in Safari
> * Performance of large mostly synchronous suites in Safari
> 
> Of course it would be good to improve all of those. But would doing major surgery on clearStack still be worth the effort and risk if running a single focused spec was reasonably fast in all browsers?

Good question. Maybe we should do one at a time and start with the single focused spec issue.

> That seems like it would be worth experimenting with. How would the clamping cooldown be implemented? Are you thinking of skipping setTimeout or doing it with a longer delay? My knee-jerk reaction is that the latter sounds less risky.

For my first test outside of the Jasmine lib I implemented an adaptive prototype which has a cold (simple implementation of this PR so far: `queueMicrotask` and minimal `postMessage` to prevent clamping) and a hot (always `postMessage`) implementation of clearStack. It gradually turns cold and gets hot if clamping is detected. I don't dare to touch the `setTimeout` rate b/c I saw in other discussions that this could break things.

In order to get some insights in clamping/declamping options and behavior I ran some tests here: https://codepen.io/dcsaszar/full/oNrdQdY

The results confirm:
* `postMessage` is slower than `queueMicrotask` but since it resets `setTimeout` clamping it wins in many situations. 
* The simple strategy implemented here is only fast in the blue (no setTimeout) scenario.
* The current Jasmine default for Chrome/Firefox is in the top four for these two browsers.
* The current Jasmine default for Chrome/Firefox may perform better in Safari with heavy clamping, but `queueMicrotask` being way faster than `postMessage` will hide slight clamping.

One outcome these tests suggest is that wrapping the `setTimeout` in the Safari clearStack implementation in a `postMessage` would solve the performance issue with focused specs, and potentially speed up suites in general. I gave this a try and my artifical suite went from 17 to 10 seconds, Unpoly from 188 to 161 s.

Should this be my next go at #2008?

sgravrock:
Sorry for the delay in responding.

> One outcome these tests suggest is that wrapping the setTimeout in the Safari clearStack implementation in a postMessage would solve the performance issue with focused specs, and potentially speed up suites in general. I gave this a try and my artifical suite went from 17 to 10 seconds, Unpoly from 188 to 161 s.
> 
> Should this be my next go at #2008?

That seems like it's worth a try. A 15% speedup in a real world test suite is certainly interesting.

I'm also planning to dig more into the possibility of skipping the queue runs for non-focused specs altogether and just reporting a single "some specs were skipped" dummy spec. But there's a good chance that that'll have to be opt-in, which would mean most people won't benefit from it.

dcsaszar:
Closing this for now. Will be back with another one soon. See above.

PR #2042, opened 2024-09-11
feat(Clock): Add ability to automatically tick the clock asynchronously

### LINKED ISSUE #1725: Proposal: await jasmine.clock().asyncTick()

### LINKED ISSUE #1932: Feature suggestion: jasmine.clock().withMockAsync()
In some of my tests, I need to use a mocked clock combined with async functions, basically the equivalent of withMock() that accepts a Promise-returning closure and returns a Promise itself. I already have a home-made wrapper that I use:

```ts
export async function withMockedClock(testBlock: () => Promise<void>): Promise<void> {
  jasmine.clock().install();

  return testBlock().finally(() => jasmine.clock().uninstall());
}
```

Example usage:
```
it('runs async tests with mocked clock', async () => {
  await jasmine.clock().withMockAsync(async () => {
      // setup code
      await expectAsync(promiseReturningFunction()).toBeResolvedTo(...);
  });
});
```

Would there be an issue to include something like this in the core library?

atscott:
## Description
Testing with mock clocks can often turn into a real struggle when dealing with situations where some work in the test is truly async and other work is captured by the mock clock. This can happen for many reasons, but as one example:

An asynchonrous change from a task in the mocked clock may change DOM where a resize observer then gets triggered. This browser API is truly asynchronous and would require the user to wait real time for it to fire. If there is follow-up work after the resize observer fires, it may be captured by the mock clock again. This would require the tester to write something like the following:

```
// flush the timer
jasmine.clock().tick();
// wait for resize observer
await new Promise(resolve => setTimeout(resolve));
// flush follow-up work from the resize observer callback
jasmine.clock().tick();
```

When using mock clocks, testers are always forced to write tests with intimate knowledge of when the mock clock needs to be ticked. Oftentimes, the purpose of using a mock clock is to speed up the execution time of the test when there are timeouts involved. It is not often a goal to test the exact timeout values. This can cause tests to be riddled with `tick`. It ideal for test code to be written in a way that is independent of whether a mock clock is installed. For example:

```
document.getElementById('submit');
// https://testing-library.com/docs/dom-testing-library/api-async/#waitfor
await waitFor(() => expect(mockAPI).toHaveBeenCalledTimes(1))
```

When mock clocks are involved, the above may not be possible if there is some delay involved between the click and the request to the API. Instead, developers would need to manually tick the clock beyond the delay to trigger the API call.

This commit attempts to resolve these issues by adding a feature to the clock which allows it to advance on its own with the passage of time, just as clocks do without mocks installed. It also allows for some breathing time so any unmocked micro and macrotasks are given space to execute as well.

This feature would also address both #1725 and #1932. `asyncTick` can be accomplished by enabling the auto tick feature and then waiting for a promise with a timout to be resolved
(`await new Promise(resolve => setTimeout(resolve, 20))`) where `setTimeout` is captured by the mock clock and flushed while the code is waiting for the promise to resolve.

resolves #1725
resolves #1932

All credit goes to @stephenfarrar for this.

## Motivation and Context

resolves #1725
resolves #1932

## How Has This Been Tested?

Following the guide for testing in `CONTRIBUTING.md`, this has been testing with the specs in chrome, edge, safari, and firefox, as well as node.

## Types of changes

~From my understanding, changing the public interfaces of the `DelayedFunctionScheduler` and the `Clock` is a breaking change?~

## Checklist:

### DIFF
--- a/spec/core/ClockSpec.js
+++ b/spec/core/ClockSpec.js
@@ -687,6 +687,142 @@ describe('Clock (acceptance)', function() {
     expect(recurring1.calls.count()).toBe(4);
   });
 
+  describe('auto tick mode', () => {
+    let delayedFunctionScheduler;
+    let mockDate;
+    let clock;
+
+    beforeEach(() => {
+      delayedFunctionScheduler = new jasmineUnderTest.DelayedFunctionScheduler();
+      mockDate = {
+        install: function() {},
+        tick: function() {},
+        uninstall: function() {}
+      };
+      clock = new jasmineUnderTest.Clock(
+        // We use the real window for global or firefox is displeased when we try to call a real setTimeout on an object "that doesn't implement window".
+        typeof window !== 'undefined' ? window : { setTimeout: setTimeout },
+        function() {
+          return delayedFunctionScheduler;
+        },
+        mockDate
+      );
+      clock.install();
+      clock.autoTick();
+    });
+
+    afterEach(() => {
+      clock.uninstall();
+    });
+
+    it('can run setTimeouts/setIntervals asynchronously', function() {
+      const recurring = jasmine.createSpy('recurring'),
+        fn1 = jasmine.createSpy('fn1'),
+        fn2 = jasmine.createSpy('fn2'),
+        fn3 = jasmine.createSpy('fn3');
+
+      const intervalId = clock.setInterval(recurring, 50);
+      // In a microtask, add some timeouts.
+      Promise.resolve()
+        .then(function() {
+          return new Promise(function(resolve) {
+            clock.setTimeout(resolve, 25);
+          });
+        })
+        .then(function() {
+          fn1();
+          return new Promise(function(resolve) {
+            clock.setTimeout(resolve, 200);
+          });
+        })
+        .then(function() {
+          fn2();
+          return new Promise(function(resolve) {
+            clock.setTimeout(resolve, 100);
+          });
+        })
+        .then(function() {
+          fn3();
+        });
+
+      expect(recurring).not.toHaveBeenCalled();
+      expect(fn1).not.toHaveBeenCalled();
+      expect(fn2).not.toHaveBeenCalled();
+      expect(fn3).not.toHaveBeenCalled();
+
+      return new Promise(resolve => clock.setTimeout(resolve, 50))
+        .then(function() {
+          expect(recurring).toHaveBeenCalledTimes(1);
+          expect(fn1).toHaveBeenCalled();
+          expect(fn2).not.toHaveBeenCalled();
+          expect(fn3).not.toHaveBeenCalled();
+
+          return new Promise(resolve => clock.setTimeout(resolve, 175));
+        })
+        .then(function() {
+          expect(recurring).toHaveBeenCalledTimes(4);
+          expect(fn1).toHaveBeenCalled();
+          expect(fn2).toHaveBeenCalled();
+          expect(fn3).not.toHaveBeenCalled();
+
+          clock.clearInterval(intervalId);
+          return new Promise(resolve => clock.setTimeout(resolve, 100));
+        })
+        .then(function() {
+          expect(recurring).toHaveBeenCalledTimes(4);
+          expect(fn1).toHaveBeenCalled();
+          expect(fn2).toHaveBeenCalled();
+          expect(fn3).toHaveBeenCalled();
+        });
+    });
+
+    it('speeds up the execution of the timers in all browsers', async () => {
+      const startTimeMs = performance.now() / 1000;
+      await new Promise(resolve => clock.setTimeout(resolve, 5000));
+      await new Promise(resolve => clock.setTimeout(resolve, 5000));
+      await new Promise(resolve => clock.setTimeout(resolve, 5000));
+      await new Promise(resolve => clock.setTimeout(resolve, 5000));
+      const endTimeMs = performance.now() / 1000;
+      // Ensure we didn't take 20s to complete the awaits above and, in fact, can do it in a fraction of a second
+      expect(endTimeMs - startTimeMs).toBeLessThan(100);
+    });
+
+    it('avoids throttling in browsers other than Safari', async () => {
+      if (
+        typeof navigator !== 'undefined' &&
+        /^((?!chrome|android|firefox).)*safari/i.test(navigator.userAgent)
+      ) {
+        return;
+      }
+      // This test ensures the setTimeout loop isn't getting throttled by browsers
+      const promises = [];
+      // 2000 timers at ~4ms throttling = 8_000ms would time out if we weren't
+      // preventing the throttle with the MessageChannel trick.
+      for (let i = 0; i < 2000; i++) {
+        promises.push(new Promise(resolve => clock.setTimeout(resolve)));
+      }
+      const startTimeMs = performance.now() / 1000;
+      await Promise.all(promises);
+      const endTimeMs = performance.now() / 1000;
+      expect(endTimeMs - startTimeMs).toBeLessThan(1000);
+    });
+
+    it('is easy to test async functions with interleaved timers and microtasks', async () => {
+      async function blackBoxWithLotsOfAsyncStuff() {
+        await new Promise(r => clock.setTimeout(r, 10));
+        await Promise.resolve();
+        await Promise.resolve();
+        await new Promise(r => clock.setTimeout(r, 20));
+        await Promise.resolve();
+        await Promise.resolve();
+        await Promise.resolve();
+        return 'done';
+      }
+      const result = await blackBoxWithLotsOfAsyncStuff();
+      expect(result).toBe('done');
+    });
+  });
+
   it('can clear a previously set timeout', function() {
     const clearedFn = jasmine.createSpy('clearedFn'),
       delayedFunctionScheduler = new jasmineUnderTest.DelayedFunctionScheduler(),
--- a/src/core/Clock.js
+++ b/src/core/Clock.js
@@ -29,6 +29,15 @@ getJasmineRequireObj().Clock = function() {
     let installed = false;
     let delayedFunctionScheduler;
     let timer;
+    // Tracks how the clock ticking behaves.
+    // By default, the clock only ticks when the user manually calls a tick method.
+    // There is also an 'auto' mode which will advance the clock automatically to
+    // to the next task. Once enabled, there is currently no mechanism for users
+    // to disable the auto ticking.
+    let tickMode = {
+      mode: 'manual',
+      counter: 0
+    };
 
     this.FakeTimeout = FakeTimeout;
 
@@ -138,7 +147,94 @@ getJasmineRequireObj().Clock = function() {
       }
     };
 
+    /**
+     * Updates the clock to automatically advance time.
+     *
+     * With this mode, the clock advances to the first scheduled timer and fires it, in a loop.
+     * Between each timer, it will also break the event loop, allowing any scheduled promise
+callbacks to execute _before_ running the next one.
+     *
+     * This mode allows tests to be authored in a way that does not need to be aware of the
+     * mock clock. Consequently, tests which have been authored without a mock clock installed
+     * can one with auto tick enabled without any other updates to the test logic.
+     *
+     * In many cases, this can greatly improve test execution speed because asynchronous tasks
+     * will execute as quickly as possible rather than waiting real time to complete.
+     *
+     * Furthermore, tests can be authored in a consitent manner. They can always be written in an asynchronous style
+     * rather than having `tick` sprinkled throughout the tests with mock time in order to manually
+     * advance the clock.
+     *
+     * When auto tick is enabled, `tick` can still be used to synchronously advance the clock if necessary.
+     */
+    this.autoTick = function() {
+      if (tickMode.mode === 'auto') {
+        return;
+      }
+
+      tickMode = { mode: 'auto', counter: tickMode.counter + 1 };
+      advanceUntilModeChanges();
+    };
+
     return this;
+
+    // Advances the Clock's time until the mode changes.
+    //
+    // The time is advanced asynchronously, giving microtasks and events a chance
+    // to run before each timer runs.
+    //
+    // @function
+    // @return {!Promise<undefined>}
+    async function advanceUntilModeChanges() {
+      if (!installed) {
+        throw new Error(
+          'Mock clock is not installed, use jasmine.clock().install()'
+        );
+      }
+      const { counter } = tickMode;
+
+      while (true) {
+        await newMacrotask();
+
+        if (
+          tickMode.counter !== counter ||
+          !installed ||
+          delayedFunctionScheduler === null
+        ) {
+          return;
+        }
+
+        if (!delayedFunctionScheduler.isEmpty()) {
+          delayedFunctionScheduler.runNextQueuedFunction(function(millis) {
+            mockDate.tick(millis);
+          });
+        }
+      }
+    }
+
+    // Waits until a new macro task.
+    //
+    // Used with autoTick(), which is meant to act when the test is waiting, we need
+    // to insert ourselves in the macro task queue.
+    //
+    // @return {!Promise<undefined>}
+    async function newMacrotask() {
+      // MessageChannel ensures that setTimeout is not throttled to 4ms.
+      // https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#reasons_for_delays_longer_than_specified
+      // https://stackblitz.com/edit/stackblitz-starters-qtlpcc
+      // Note: This trick does not work in Safari, which will still throttle the setTimeout
+      const channel = new MessageChannel();
+      await new Promise(resolve => {
+        channel.port1.onmessage = resolve;
+        channel.port2.postMessage(undefined);
+      });
+      channel.port1.close();
+      channel.port2.close();
+      // setTimeout ensures that we interleave with other setTimeouts.
+      await new Promise(resolve => {
+        realTimingFunctions.setTimeout.call(global, resolve);
+      });
+    }
 
     function originalTimingFunctionsIntact() {
       return (
--- a/src/core/DelayedFunctionScheduler.js
+++ b/src/core/DelayedFunctionScheduler.js
@@ -87,6 +87,37 @@ getJasmineRequireObj().DelayedFunctionScheduler = function(j$) {
       }
     };
 
+    // Returns whether there are any scheduled functions.
+    // Returns true if there are any scheduled functions, otherwise false.
+    this.isEmpty = function() {
+      return this.scheduledFunctions_.length === 0;
+    };
+
+    // Runs the next timeout in the queue, advancing the clock.
+    this.runNextQueuedFunction = function(tickDate) {
+      if (this.scheduledLookup_.length === 0) {
+        return;
+      }
+
+      const newCurrentTime = this.scheduledLookup_[0];
+      if (newCurrentTime >= this.currentTime_) {
+        tickDate(newCurrentTime - this.currentTime_);
+        this.currentTime_ = newCurrentTime;
+      }
+
+      const funcsAtTime = this.scheduledFunctions_[this.currentTime_];
+      const fn = funcsAtTime.shift();
+      if (funcsAtTime.length === 0) {
+        delete this.scheduledFunctions_[this.currentTime_];
+        this.scheduledLookup_.splice(0, 1);
+      }
+
+      if (fn.recurring) {
+        this.reschedule_(fn);
+      }
+      fn.funcToCall.apply(null, fn.params || []);
+    };
+
     return this;
   }
 
--- a/spec/core/DelayedFunctionSchedulerSpec.js
+++ b/spec/core/DelayedFunctionSchedulerSpec.js
@@ -264,6 +264,42 @@ describe('DelayedFunctionScheduler', function() {
     expect(fn).toHaveBeenCalled();
   });
 
+  it('runs the next scheduled funtion', function() {
+    const scheduler = new jasmineUnderTest.DelayedFunctionScheduler();
+    const fn = jasmine.createSpy('fn');
+    const tickSpy = jasmine.createSpy('tick');
+
+    scheduler.scheduleFunction(fn, 10, [], false, 'foo');
+
+    expect(fn).not.toHaveBeenCalled();
+
+    scheduler.runNextQueuedFunction(tickSpy);
+
+    expect(fn).toHaveBeenCalled();
+    expect(tickSpy).toHaveBeenCalledWith(10);
+  });
+
+  it('runs the only a single scheduled funtion in a time slot', function() {
+    const scheduler = new jasmineUnderTest.DelayedFunctionScheduler();
+    const fn1 = jasmine.createSpy('fn');
+    const fn2 = jasmine.createSpy('fn2');
+    const tickSpy = jasmine.createSpy('tick');
+
+    scheduler.scheduleFunction(fn1, 10, [], false, 'foo1');
+    scheduler.scheduleFunction(fn2, 10, [], false, 'foo2');
+
+    scheduler.runNextQueuedFunction(tickSpy);
+
+    expect(fn1).toHaveBeenCalled();
+    expect(fn2).not.toHaveBeenCalled();
+    expect(tickSpy).toHaveBeenCalledWith(10);
+
+    tickSpy.calls.reset();
+    scheduler.runNextQueuedFunction(tickSpy);
+    expect(fn2).toHaveBeenCalled();
+    expect(tickSpy).toHaveBeenCalledWith(0);
+  });
+
   it('updates the mockDate per scheduled time', function() {
     const scheduler = new jasmineUnderTest.DelayedFunctionScheduler(),
       tickDate = jasmine.createSpy('tickDate');


src/core/Clock.js
```diff
@@ -138,8 +142,88 @@ getJasmineRequireObj().Clock = function() {
...
+        ) {
+          return;
+        }
+
+        if (!delayedFunctionScheduler.isEmpty()) {
+          delayedFunctionScheduler.runNextQueuedFunction(function(millis) {
+            mockDate.tick(millis);
+          });
+        } else {
+          await Promise.resolve();
```
stephenfarrar, line 199: Just wondering about this: is it necessary?

After this line, the very next thing that will happen is going back to the top of the while loop and awaiting newMacrotask. So this could be redundant.

sgravrock:
> From my understanding, changing the public interfaces of the DelayedFunctionScheduler and the Clock is a breaking change?

Not necessarily. `DelayedFunctionScheduler` is internal, so any changes to it are fair game for a minor release as long as they preserve the behavior of the [public APIs](https://jasmine.github.io/api/edge/global). Additive changes to `clock` can also go in a minor release. At a glance, I'm not seeing anything here that looks like it would break existing code.

src/core/Clock.js
atscott: Agreed, I don't think it is. Removed.

spec/core/ClockSpec.js
```diff
@@ -687,6 +687,136 @@ describe('Clock (acceptance)', function() {
...
+        .then(function() {
+          fn3();
+        });
+
+      expect(recurring).not.toHaveBeenCalled();
+      expect(fn1).not.toHaveBeenCalled();
+      expect(fn2).not.toHaveBeenCalled();
+      expect(fn3).not.toHaveBeenCalled();
+
+      clock.setAutoTickMode(true);
```
ewinslow, line 753: For the sake of providing a stable yet evolvable API, I would really recommend making this function take an enum instead of Boolean. It's hard to know if some future subtle change will be needed to the tick mode. Making this multiple choice rather than true false can allow for experimentation and smooth upgrades.

atscott: I'm certainly open to making that change. I don't know of any enum-like inputs in jasmine today so this might be a bike-shedding question for the maintainers for how best to do this.

// cc @sgravrock

sgravrock:
I think I understand the problems this is meant to solve, but it's not clear to me how a user would use it to solve those problems. Could you provide an example spec that shows realistic usage? Or better yet, one example spec per problem this is meant to solve.

My best guess is that the intended usage is something like this:

```
beforeEach(function() {
   jasmine.clock().install();
});

it('uses the new async tick feature', async function() {
   jasmine.clock().install();
   doThingsThatRequireTheMockClock();

   try {
      const promise = doSomethingThatNeedsRealAsync();
      jasmine.clock().setAutoTickMode(true);
      await promise;
   } finally {
      jasmine.clock().setAutoTickMode(false);
   }

   doMoreThingsThatRequireTheMockClock();
});
```

But I might not have that right.

src/core/DelayedFunctionScheduler.js
```diff
@@ -87,6 +87,42 @@ getJasmineRequireObj().DelayedFunctionScheduler = function(j$) {
       }
     };
 
+    /**
```
sgravrock, line 90: Please remove the jsdoc comments from everything that isn't meant to be part of the public interface. Explanatory comments on private methods etc are fine as long as you avoid jsdoc syntax. All jsdoc comments semi-automatically become part of the published API reference documentation.

atscott: For these two, I actually think they need to be in the public interface, right? Assuming developers can implement their own `DelayedFunctionScheduler` independently from the `Clock` implementation, these functions do need to be available to the `Clock` in order to use the auto ticking behavior.

atscott:
> I think I understand the problems this is meant to solve, but it's not clear to me how a user would use it to solve those problems. Could you provide an example spec that shows realistic usage? Or better yet, one example spec per problem this is meant to solve.

I'm not sure how to go about doing this in the end. Generally this would end up making the jasmine test suite a lot more complicated than I think it should be. The use-cases would include a lot of "business"-type logic or async browser APIs (such as resize observer) that would be more easily understood to a reader of a spec as a simple `Promise.resolve`. I'm afraid adding real world examples would make the tests harder to understand and debug rather than easier. 

An API like an `asyncTick` is needed because without it, mock clocks don't have the ability to drain the microtask queue between the macrotask timers that they capture. It is especially difficult to get the correct behavior when microtasks are interleaved between the timer functions. As mentioned in the commit message, testers are forced go back and forth between manually ticking the mock clock and then flush microtasks somehow (more difficult without having a handle on the real `setTimeout`). 

The auto ticking API goes a bit further while being able to accomplish the same correctness attributes of an `asyncTick` (`asyncTick(100)` is equivalent to `new Promise(r => setTimeout(100, r)` with auto ticking). When you enable automatic ticking, you can effectively forget about the mock clock altogether. You no longer need to think about what needs to be flushed and what doesn't because time advances on its own, just like it does without a mock clock. 

> My best guess is that the intended usage is something like this:
> ...

Generally, my expected usage would instead be to "enable and forget" rather than micro-managing when to enable and disable the auto-tick behavior. That would be more like what you would do with something like `tickAsync` (i.e. choose to use async tick at a particular spot when needed). Using auto ticking would look more like:

```
beforeEach(function() {
   jasmine.clock().install();
   jasmine.clock().setAutoTickMode(true);
});

afterEach(function() {
   jasmine.clock().uninstall();
});

it('uses the new async tick feature', async function() {
   doThingsThatRequireTheMockClockAndMicrotasks();
   await someCondition();

   doMoreThingsThatRequireTheMockClockAndMicrotasks();
   await otherCondition();
});
```

Without auto ticking in the test above, you have to ensure that the thing you're awaiting either (a) only has microtask async things and does not use any timer APIs, or (b) knows about the mock clock and can tick until the conditions for resolving the promise become true. In the second scenario, testing libraries or harnesses that poll or wait for conditions to be true need to be told how to advance timers or they may never reach the expected conditions. This is the case for the popular [Testing Library](https://testing-library.com/docs/user-event/options#advancetimers) and its [async methods](https://testing-library.com/docs/dom-testing-library/api-async), but also applies to any other async test harnesses which wrap async behaviors that include timers. As a tester, you don't always know or control what the testing APIs do, so it can be dangerous or difficult to wait for those promises to resolve.

I've updated the tests in two ways to hopefully cover the questions/comments you have: first, auto tick is enabled in the `beforeEach` like I would expect in practice. Second, I added a test scenario to have some example "black box" helper function that includes a "random" number of timeouts and promises. Without auto ticking, getting to a point where the result is available would be quite difficult: `tick(30)` wouldn't work, nor does `tick(30); await Promise.resolve();`. You would have to instead know the implementation detail of the black box and alternate ticking enough time with resolving enough promises to flush the microtask queue between the timers.

atscott:
As one example that seems reasonable for an app to test with fake timers, consider an input meant to be some instant search API. The search is debounced (using a timer), the results are awaited (a Promise, and the mocked response might be just a Promise.resolve() or simulated delay) before the results are updated. In addition, maybe the results page blurs and shows a progress bar when the fetch takes longer than 200ms to give feedback to the user that something's happening. On top of that, this state is guaranteed to be visible for some amount of time (maybe also 200ms) to avoid the "flicker" of the UI when something is shown and then removed immediately. 

All of these things together can lead to a _really_ difficult situation to test when mock timers are involved. You can't just dispatch an input event to the input and wait for the results to display. You have to find a way to tick your way through all the steps required to get to the end state.

spec/core/ClockSpec.js
atscott: I've updated the implementation to use a string union. Since this is only used for tests (no bundle size concerns), I felt that'd be more ergonomic than an enum/object alternative.

sgravrock:
> I'm not sure how to go about doing this in the end. Generally this would end up making the jasmine test suite a lot more complicated than I think it should be.

Sorry, I didn't mean to suggest that you should include those tests in Jasmine's suite. I was looking for something to help me understand how the new API is supposed to be used so I can evaluate the API's design, document it, support it later on, etc.

I think I understand how to use this in a couple of cases:

* Running the entire spec with auto tick mode
* Doing manual ticking at the start, then switching to auto tick mode for the remainder of the spec
 
But I'm not sure how to write a spec that switches back out of auto tick mode. Seems like I'd need to schedule a call to `jasmine.clock().setTickMode("manual")` at just the right point, right? 

Any chance you could provide a link to a gist or repo containing a fully worked, "realistic-ish" example spec that uses the entire API? I think that would help.

src/core/DelayedFunctionScheduler.js
sgravrock: Users can't provide their own `DelayedFunctionScheduler`. (Well, it's technically possible to construct your own `Clock`, but [the documentation explicitly says not to](https://jasmine.github.io/api/5.3/Clock.html)). Even if it didn't, `DelayedFunctionScheduler` is an undocumented private API and while Jasmine doesn't go out of our way to block users from accessing it, we also don't put any effort into preserving compatibility with user code that does.

With a couple of edge-case exceptions, anything not mentioned on https://jasmine.github.io is a private API. Promoting private APIs to public is something I don't do without a compelling reason, because it limits our ability to evolve Jasmine. You're benefiting from that policy right now: If `DelayedFunctionScheduler` and the `Clock` constructor were already public, I'd have to decline this PR because it would break compatibility with user code that supplies its own `DelayedFunctionScheduler`.

sgravrock:
Ah, now that I re-read your earlier comment, I think I misunderstood your intent:

> Generally, my expected usage would instead be to "enable and forget" rather than micro-managing when to enable and disable the auto-tick behavior.

If the idea is that auto tick mode should be in effect for the entire duration of the spec, then maybe the API should consist of just an `autoTick` or `automaticallyAdvance` method on `Clock` that takes no parameters? That would avoid the risk of creating an impression, like I had after my first look at this, that users not only could disable auto tick mode but were expected to do so.

sgravrock:
I assume you've lost interest in this, so I'll eventually close it if I don't hear anything more. If I'm wrong about that, I'm still willing to move forward. But only if we can come to a consensus about the API.

My thinking at this point is that an `autoTick` or `automaticallyAdvance` method on `Clock` that takes no parameters covers all the use cases we've discussed, and should be quite a bit easier for users to understand and use than `setTickMode`. However, if nobody says "yes, I think that's the right API, and I would use it", I'm going to take that as a no. I'm not necessarily opposed to something more complicated but I'd need to see solid justification for it and a clear explanation of how to use all aspects of it.

atscott:
Hi Steve, I’m definitely still interested but I’m on paternity leave until mid-January. Looking forward to working on this with you.

atscott:
> If the idea is that auto tick mode should be in effect for the entire duration of the spec, then maybe the API should consist of just an `autoTick` or `automaticallyAdvance` method on Clock that takes no parameters? That would avoid the risk of creating an impression, like I had after my first look at this, that users not only can disable auto tick mode but are expected to do so.

That's certainly an option, though I think the big concern for me here is that if there's a use-case for wanting to write a test that does not use automatic advancement, then there's definitely also a use-case for having it enabled at the top level `describe` and then disabling it for a given test. It's not exactly an expectation that you _should_ disable it, but that we're expecting some people might want to for specific use-cases.

You might want to disable auto advancing whenever your test does something that is truly async, such as making a real RPC, asking webdriver to take a screenshot, or synthesize a real click event, cross-frame communication, or waiting for a CSS animation to finish. The reason to pause it might be that you're trying to test or screenshot something at a given point and don't want the timers to flush while you're waiting for something truly async. You might do one of these things for a single (or a few) tests in the suite, but still want to work in a world of auto-advancing for everything else.

It'd be unfortunate to have things only flip one way. If you want to have manual ticking by default and enable auto advance for some tests, you're in a happy place. But if you want things the other way, it wouldn't be possible with just an "autoTick" method.


------


Separately, the initial design did have just an autoTick(true/false) method. @ewinslow [suggested](https://github.com/jasmine/jasmine/pull/2042#discussion_r1758111871) something more future-proof (and I think rightfully so). I could imagine another tick mode similar to what jest/sinon has which automatically advances 1ms every 1ms, for example (sinon advances 20ms every 20ms). You could use enable this mode to put your test in a "live"/"playground" state where things are neither frozen (manual ticks like today) or fast-forwarded as quickly as possible (the auto tick proposal). In this mode, you could interact with the UI in the test for easier debugging.

sgravrock:
In Jasmine's context, I disagree that what you're proposing future-proofing. It might very well paint us into a corner that we can only get out of with a breaking change. Breaking changes can be catastrophically expensive. I estimate that the Karma debacle alone left roughly half of Jasmine's user base stranded on 4.x. And Karma is far from the only abandonware that people use Jasmine with. 

So if we ship something narrow that we know is useful, like `clock.autoTick()`, we can iterate on it by adding more methods later if the need arises. But we can't iterate by changing or removing APIs after they ship. I'm not going to do a major release of jasmine-core just to fix problems with this feature. (I'm going to try very hard to avoid doing a major release of jasmine-core at all, for as long as I can.) Everyone's going to be stuck with whatever ships. That makes speculative APIs risky.

I think there are two key questions to think about in a situation like this:

1. Would shipping a narrower API first and then expanding in a non-breaking way later work?
2. How confident are we that the broader API successfully meets a real user need?

I think the answer to the first question is yes. Maybe at the cost of a little bit of redundancy, but that's acceptable. The answer to the second question is no: For my part at least, I'm not even a little bit confident. Unless I've missed something, there isn't a single piece of working example code that disables auto tick, and there isn't a single person saying they actually have a problem that it would solve.

That would be cause for concern in any case, but it's especially a problem in this case because I haven't been able to figure out how to write a spec that benefits from disabling auto tick. It seems to me that in all but one case (more about that one in a moment) the key would be to arrange for the call to `clock.setTickMode('manual’)` to happen at just the right time. How do you do that? How do you document it so that a user with poor to average async skill can do it without opening an issue asking for help?

> there's definitely also a use-case for having it enabled at the top level describe and then disabling it for a given test.

I disagree. I don’t think it’s good style to do setup in outer describes and undo it in inner ones. (I personally have a hard time working on test suites that do that, and I know I'm not alone in that.) Users are free to write suites that way, but [we’ve declined to build in support for that style in the past](https://jasmine.github.io/pages/faq.html#let) and I think that was the right call.

> You might want to disable auto advancing whenever your test does something that is truly async, such as making a real RPC, asking webdriver to take a screenshot, or synthesize a real click event, cross-frame communication, or waiting for a CSS animation to finish. The reason to pause it might be that you're trying to test or screenshot something at a given point and don't want the timers to flush while you're waiting for something truly async. 

Those are all plausible scenarios, and it’s plausible that the proposed API might solve those problems. If I could hear from real people who have some of those problems and see working code that uses the proposed API to solve them, I'd be more likely to go along with this.

sgravrock:
If I'm right to think that request is coming from the Angular team, then I should also point out that I don't intend to do any further backports of new features to jasmine-core 4.x. That likely limits the number of Angular users who would have access to this feature.

sgravrock:
I wrote:

> If I could hear from real people who have some of those problems and see working code that uses the proposed API to solve them, I'd be more likely to go along with this.

If you feel strongly that these use cases are important but don't know anyone who needs them right now, another option is to create quality documentation, including:

* A tutorial that shows how to use all aspects of the new API. Ideally this would allow a user who doesn't have a deep understanding of async and has never heard of things like microtasks to use the new API without asking for help.
* A fully worked, run-able example that shows how to use all aspects of the API to solve concrete real-world testing problems. [An example project checked into the docs repo](https://github.com/jasmine/jasmine.github.io/tree/master/examples) might be the best option. Inline code snippets in the tutorial could work if it's quite small and self-contained. I'd specifically want to see something that solves a problem by turning auto-tick back off in the middle of a test, since that's the aspect that I'm most skeptical of. This should be framework-agnostic: no Karma or Angular dependency, please.

What I'm asking you to do, one way or another, is to put in the work to show that the broader API is both useful and usable and to minimize the chance of it creating a support burden.

atscott:
> So if we ship something narrow that we know is useful, like clock.autoTick(), we can iterate on it by adding more methods later if the need arises
...
> If you feel strongly that these use cases are important [for disabling auto ticking]...

Ultimately, while I do think there would be some cases where it would be useful to disable the auto ticking within a test when it had already been disabled, I don't think it's so common or important that I would push hard for it from the start. I agree that the simple API covers the vast majority of use-cases and can be iterated on if necessary. If a developer needs to disable the feature for a test, they could always reorganize the suite in a way that never enables it for that one case. 

> If I'm right to think that this request is coming from the Angular team, then I should also point out that I don't intend to do any further backports of new features to jasmine-core 4.x. That likely limits the number of Angular users who would have access to this feature.

All good. We do think there's a lot of value in being able to write tests against an auto ticking mock clock. Even if this feature isn't available in older versions of Jasmine, the fact that it's out there in the world can help developers start thinking about how it might be useful for their own use-cases. They could always use a different mock clock implementation if it's not available in the built-in jasmine versions.


----

If you're happy with the simple API (i.e. `clock.autoTick()`), I can push up that change to this PR.

sgravrock:
> They could always use a different mock clock implementation if it's not available in the built-in jasmine versions.

True. The rest of Jasmine doesn't depend on the mock clock, so there's no reason somebody couldn't use a different one.

> If you're happy with the simple API (i.e. clock.autoTick()), I can push up that change to this PR.

Sounds good. Thank you!

atscott:
Okay, I've updated the public API to just have `clock.autoTick` with some additional documentation of why the API might be useful. The internals are still implemented in a way that could support an API for disabling it after it was enabled.

spec/core/ClockSpec.js
```diff
@@ -687,6 +687,123 @@ describe('Clock (acceptance)', function() {
...
+    it('can execute hundreds of timers "quickly"', async () => {
+      // This test ensures the setTimeout loop isn't getting throttled by browsers
+      const promises = [];
+      for (let i = 0; i < 1000; i++) {
+        promises.push(new Promise(resolve => clock.setTimeout(resolve)));
+      }
+      const startTimeMs = performance.now() / 1000;
+      await Promise.all(promises);
+      const endTimeMs = performance.now() / 1000;
+      expect(endTimeMs - startTimeMs).toBeLessThan(1000);
```
sgravrock, line 788: On Safari (all supported versions), this test fails with a timeout. When I upped the timeout to 10 seconds and added some logging, the test passed and I saw that the value of `endTimeMs - startTimeMs` was 5.476. So I think there's an incorrect unit conversion, but also Safari is actually doing the throttling that the test is checking for.

```
Failures:
1) Clock (acceptance) auto tick mode can execute hundreds of timers "quickly"
  Message:
    Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in http://localhost:5555/__jasmine__/jasmine.js (line 8204)
  Stack:
    <Jasmine>
```

ewinslow:
I would like to provide more context with a common use-case we have for changing the tick mode during a test.

We have a custom tool we call "pausepoints" that is meant to stop progress in the middle of a running test at a certain point so that the test author can interact with whatever UI has been rendered at that point. I was skeptical of this feature at first because "why invent another `debugger`" but trying things out the issue quickly become apparent: `debugger;` doesn't leave the UI interactable which impedes certain investigation, and it makes auto-reload/HMR impossible since all the scripts are frozen.

```ts
it('renders some UI', async () => {
  // enable auto-tick and render
  // simulate user actions

  await pausepoint(); // clock should be able to exit auto-tick mode here, then restore it afterwards

  // the rest of the rest, more interactions, expects, etc
});
```

When a pause point is triggered, we uninstall the fake clock if one is installed, so that time can advance normally while the test author investigates/interacts. At some point the test author triggers testing to continue (via some UI we provide, or a console command). We then re-install the fake clock that was there before. Importantly, the continuation should be in the same tick mode as before.

Looking forward to your thoughts.

ewinslow:
I will also +1 the "You might want to disable auto advancing whenever your test does something that is truly async, such as making a real RPC, asking webdriver to take a screenshot". If there's anything on the screen that is derived from the current time, being able to pause the clock and set the current time to a known instant could is valuable for reducing flakes. The workaround right now is to just ignore the style of that part of the screen, at the cost of slightly reduced coverage.

sgravrock:
That's interesting. If I've understood correctly, your current implementation is something like this:

```
async function pausepoint() {
   const clock = jasmine.clock();
   clock.uninstall();
   await new Promise(function(resolve) {
      //... hook up resolve so something else can cause it to be called later
   });
   clock.install();
}
```

and you're envisioning something like this:

```
async function pausepoint() {
   const clock = jasmine.clock();
   const wasAutoTicking = clock.isAutoTicking();
   clock.uninstall();
   await new Promise(function(resolve) {
      //... hook up resolve so something else can cause it to be called later
   });
   clock.install();
   clock.setAutoTick(wasAutoTicking);
}
```

Is that about right? (I suspect that I'm missing something: the clock throws away scheduled fns when it's uninstalled, so reinstalling the existing one shouldn't have the effect that I think you're describing.)

The big potential difference I can see is that a spec that installs the mock clock is by default in full control of `setTimeout`/`setInterval` execution, but that control goes away once you enable auto tick. I can imagine this working well in some cases but in others I can imagine it being hard to figure out the right place/time to call `pausepoint()`.

Would you be willing to try that out using the code in this PR (modified to add the missing bits that you need) and see if it proves workable for debugging your actual specs?

ewinslow:
You've got the spirit of it, yea. I'll find some time to try this out and report findings.

spec/core/ClockSpec.js
atscott: We weren't able to find a way to prevent the throttling behavior in Safari. I feel like it's still valuable to have the throttle prevention in place to improve performance in other browsers so I updated this test to omit Safari and added another one to run in all browsers that verifies timers complete much quicker than waiting real time. Please let me know if you feel differently.

atscott:
Hi @sgravrock, we've discussed this some more and feel like the simple `autoTick()` API is sufficient. This should work well enough with `pausepoints`, with minor edge cases for tests that have assertions on intermediate state _after_ the `pausepoint` since it'll essentially flush them while waiting for the `pausepoint` promise to resolve.

src/core/Clock.js
```diff
@@ -138,8 +147,93 @@ getJasmineRequireObj().Clock = function() {
...
+    //
+    // Used with autoTick(), which is meant to act when the test is waiting, we need
+    // to insert ourselves in the macro task queue.
+    //
+    // @return {!Promise<undefined>}
+    async function newMacrotask() {
+      // MessageChannel ensures that setTimeout is not throttled to 4ms.
+      // https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#reasons_for_delays_longer_than_specified
+      // https://stackblitz.com/edit/stackblitz-starters-qtlpcc
+      // Note: This trick does not work in Safari, which will still throttle the setTimeout
```
sgravrock, line 225: That's not particularly surprising. We've seen similar throttling issues in Safari elsewhere.

What's the impact of this on tests that use the new feature? Does it make autoTick useless in Safari, or (I'm guessing) just a bit slower? If you were going to put a caveat about this in the docs and release notes, what would you say?

atscott: I played with this a bunch and cannot conclusively say when Safari chooses to throttle the timeouts in this feature. When I add back the test which attempts to advance through 2000 timers and also log the `performance.now` difference between the start and end of the macrotask, sometimes I get through the test with all 0s and other times it's closer to the throttled ~4ms. Firefox and Chrome are always able to do it with ~0-1ms and NodeJS is consistently ~1ms. 

> Does it make autoTick useless in Safari, or (I'm guessing) just a bit slower

So based on the above, this appears to be "just a bit slower" even in the worst case. It mostly depends how you consider 4ms vs 0-1ms: is it slightly slower (3ms) or 4 times slower or worse. Even in a suite with thousands of timers, it's still going to be decently fast, _especially_ compared to running the timers without the mock clock at all.

> If you were going to put a caveat about this in the docs and release notes, what would you say?

I might note that Safari may throttle the timers in the autoTick mode to approximately 4ms apart so a suite with 1000 timers would use about 4 seconds of pure idling time, just waiting for the auto tick to advance.

atscott: Sounds good! If [Scheduler.yield](https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/yield) eventually makes its way out of experimental and to other browsers, we should be able to update the implementation to use that and avoid any throttling.

sgravrock:
Merged. Thanks for the contribution and for your cooperation.

atscott:
@sgravrock Awesome! Our repo is on the latest version of jasmine so I'm looking forward to the next release when this is available 🥳

spec/core/ClockSpec.js
sgravrock: @atscott It looks like something might be amiss in Node too. This one test takes over 2.5 seconds to run in Node. For comparison, the entire rest of the jasmine-core test suite takes about a second. The test should fail, but it passes because it still has the incorrect unit conversion bug I mentioned earlier.

sgravrock: It's actually impossible for this test to fail on CI. It only fails if the elapsed time is at least 1000 **seconds**, but Circle CI will kill the entire CI job before then.

atscott: Oops, sorry I missed the unit conversion bug! I can send a PR to fix that.

On the time in Node: It doesn't appear that Node is throttling the timeouts, but that it just takes that long in general, even for the very first `setTimeout(0)` to run. I can reduce the number of timers the test is advancing through. It currently does 2000 timers which is certainly overkill for what it's trying to achieve (though it did help with identifying this unit conversion bug). For reference, the rest of the suite only appears to execute 62 total scheduled functions through the `DelayedFunctionScheduler`.

sgravrock: > I can send a PR to fix that.

I'd appreciate that.

>  I can reduce the number of timers the test is advancing through.

Sure. But I think it also might be worth disabling the test on Node as well.

There are a couple of different perspectives to look at this test from:

* The test verifies a particular implementation detail (setTimeout throttling)
* The test verifies a user-visible behavior (you can do lots of setTimeouts very quickly in auto tick mode)

From the first perspective, it makes sense to keep running the test in Node because setTimeout throttling works. From the second perspective, it doesn't make any more sense to run the test in Node than it does in Safari. I think the tie-breaker in favor of disabling it in Node is that one test for a niche feature shouldn't take an appreciable portion of the total test suite's run time. The 0.11 seconds or so that the test takes in e.g. Firefox on my machine is already a bit much: one test like that isn't a problem, but two or three more would make a noticeable impact.

So I think it makes sense to fix the conversion bug, reduce the number of iterations, and skip the test on platforms where it's known not to work.

atscott: I think it’s fair to disable it for Node. My intent for this test was the first case: to verify an implementation detail that we do the best we can to make execution as fast as possible, regardless of how that is implemented internally (eg we can change to scheduler.yield if we wanted). Since that implementation detail doesn’t work for safari and makes no difference in Node (not that it doesn’t work, just that there’s never any throttling), we might as well skip it for those environments. 

Without the test, there’s a risk of someone negatively impacting performance by refactoring and removing that implementation detail for whatever reason: code simplification, misunderstanding the purpose, thinking that it’s not necessary anymore for whatever reason, etc.

sgravrock:
@atscott I just noticed something I missed in my earlier reviews. `autoTick` is synchronous, but it calls the asynchronous `advanceUntilModeChanges`. That would mean that the user code immediately after the call to `autoTick` could run before all of the effects of `autoTick`, right?

It looks to me like it should be more like this:

```
this.autoTick = async function() {
  if (tickMode.mode === 'auto') {
    return;
  }

  tickMode = { mode: 'auto', counter: tickMode.counter + 1 };
  await advanceUntilModeChanges();
};
```

and callers should await the result of `autoTick`.

Does that seem right to you? If so, that creates some doubt in my mind about how thoroughly this has been tested.

atscott:
`advanceUntilModeChanges` is an infinite loop and not meant to be awaited (it’d resolve only when the mode changes again). It’s async only because it’s easier to write with async/await but I can definitely see how that can lead to confusion about whether the returned promise should be awaited. I could refactor it to keep an inner async/await function and make it a void return if that makes it more clear.

atscott:
> In typical usage, would user code poll for some side effect of the timers that `advanceUntilModeChanges` triggers?

Yes, exactly. This would allow writing unit test code in a way that's similar to how it would look in an e2e test:

```
...
await user.click('Log in');
await user.see('Welcome, atscott!');
```

Since the polling would also be included in the mock clock, the autoTick would auto-advance to the poll timer automatically and the test code would resume once the conditions are satisfied.

This feature allows you to get the the benefits of mock clocks (faster time, disposing timers at the end of the test so they don't leak to future tests, etc) without making the time advancement or clock tick implementation part of the test spec. In most situations, you could add this to any existing async test that doesn't have a mock clock and not have to change anything else.

sgravrock:
This has been released in 5.7.0.

atscott:
Just wanted to circle back and share some results of integrating this into the Angular Router tests (it took a while because we had some tooling updates necessary to update to the latest jasmine version): https://github.com/angular/angular/pull/62776

I think this does a great job demonstrating how easy this feature is to integrate into existing suites to get the benefits of mock clocks with minimal toil.

atscott:
> I think you mentioned somewhere that this helps with the move to zoneless as well?

Yes, indeed it does. Typically full zoneless would also include removing `fakeAsync` in the tests. This was done in the Router code, by moving all `fakeAsync` uses to real timeouts (https://github.com/angular/angular/pull/61078). 

I had to adjust a bunch of timers to be a lot smaller. This change also resulted in some flakiness until the timeouts were adjusted since real timeouts introduce some ordering problems (e.g. if there are 2 timeouts, `timeout(5)` and `timeout(1)` and `timeout(0)` takes 10 ms to execute, then the `timeout(5)` is really set to run immediately rather than "4 ms later"). Adjusting them to have bigger differences of course impacts overall execution time. 

`useAutoTick` makes the migration off of `fakeAsync` a lot easier because the timings are more deterministic and still execute in a fake, accelerated nature. Automated migrations can much more easily move `tick(#)` to `await timeout(#)` when there is an auto ticking feature that can be enabled post-migration if necessary.

Issue #2015, opened 2023-08-24
Use globalThis instead of jasmineGlobal


0o001:
### Your question

Wouldn't it be better to use globalThis instead of the code here?

### Example code that demonstrates the problem

```JavaScript
https://github.com/jasmine/jasmine/blob/main/lib/jasmine-core/jasmine.js#L27
```

### jasmine-core version

latest

### Node.js or browser version

node 18, 20

### Operating System

All

sgravrock:
I'd certainly use globalThis if I were writing that code from scratch today. It might be worth making the switch if we ever have to change how the global object is detected for other reasons. But I don't think it's worth switching to globalThis just for the sake of switching to globalThis.

It would allow us to turn this:

```
  if (
    typeof module !== 'undefined' &&
    module.exports &&
    typeof exports !== 'undefined'
  ) {
    if (typeof global !== 'undefined') {
      jasmineGlobal = global;
    } else {
      jasmineGlobal = {};
    }
    jasmineRequire = exports;
  } else {
    if (
      typeof window !== 'undefined' &&
      typeof window.toString === 'function' &&
      window.toString() === '[object GjsGlobal]'
    ) {
      jasmineGlobal = window;
    }
    jasmineRequire = jasmineGlobal.jasmineRequire = {};
  }
```

into something more like this:

```
 if (
    typeof module !== 'undefined' &&
    module.exports &&
    typeof exports !== 'undefined'
  ) {
    jasmineRequire = exports;
  } else {
    jasmineRequire = globalThis.jasmineRequire = {};
  }
```

Which is nice enough, and would provide a small maintainability benefit... if that code needed any maintenance. But nobody's had to touch it since 2017. (2016, if you only count code that actually affects jasmineGlobal.) So the benefit is purely cosmetic. 

The old code didn't become any less correct when globalThis arrived. It's just no longer the tidiest way to do the job. That has to be weighed against the fact that the existing code has been working correctly across all supported environments for 7 years.

PR #2048, opened 2024-12-11
Use globalThis if everything else fails

jkrems:
## Types of changes

## Checklist:

Fixes https://github.com/jasmine/jasmine/issues/2015

### DIFF
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -44,6 +44,12 @@ var getJasmineRequireObj = (function(jasmineGlobal) {
     ) {
       jasmineGlobal = window;
     }
+    if (
+      typeof jasmineGlobal === 'undefined' &&
+      typeof globalThis !== 'undefined'
+    ) {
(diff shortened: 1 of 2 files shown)


sgravrock:
I know this is still marked as a draft, but, to potentially save you some time:

If the changes in this PR make a difference, then something is wrong with the behavior of `this` in your environment. Is it possible that something in your setup is wrapping Jasmine's code in a `"use strict"` context? That would break the global object detection, and also other things. I've only reviewed a small amount of Jasmine's code for strict mode compatibility, and I've already found [one thing that can't work in strict mode](https://github.com/jasmine/jasmine/blob/7683325d687d559d68dcc856aa99c88830aec801/src/core/DelayedFunctionScheduler.js#L1-L3).

If you'd like to move forward with this, please open an issue explaining why this is needed and include a minimal reproduction. I'm much more likely to say yes if you can show that there's a problem running an un-modified copy of Jasmine in a supported environment without any monkey patching.

PR #2051, opened 2025-01-11
toHaveNoOtherSpyInteractions implementation

Eradev:
## Description

* Added matcher `toHaveNoOtherSpyInteractions` to verify for any unverified calls.
* Modified `toHaveBeenCalled` matchers to verify call when passing.

## Motivation and Context

Fixing #1991

## How Has This Been Tested?

Ran `npm test`, `npm run serve` in the latest version of Chrome and Firefox, and `npm run cleanup`

## Types of changes

## Checklist:

### DIFF
--- a/src/core/CallTracker.js
+++ b/src/core/CallTracker.js
@@ -125,6 +125,16 @@ getJasmineRequireObj().CallTracker = function(j$) {
     this.saveArgumentsByValue = function() {
       opts.cloneArgs = true;
     };
+
+    /**
+     * Get the number of unverified invocations of this spy.
+     * @name Spy#calls#unverifiedCount
+     * @function
+     * @return {Integer}
+     */
+    this.unverifiedCount = function() {
+      return calls.reduce((count, call) => count + (call.verified ? 0 : 1), 0);
+    };
   }
 
   return CallTracker;
--- a/src/core/CallTracker.js
+++ b/src/core/CallTracker.js
@@ -126,12 +126,6 @@ getJasmineRequireObj().CallTracker = function(j$) {
       opts.cloneArgs = true;
     };
 
-    /**
-     * Get the number of unverified invocations of this spy.
-     * @name Spy#calls#unverifiedCount
-     * @function
-     * @return {Integer}
-     */
     this.unverifiedCount = function() {
       return calls.reduce((count, call) => count + (call.verified ? 0 : 1), 0);
     };
--- a/src/core/matchers/toHaveBeenCalledWith.js
+++ b/src/core/matchers/toHaveBeenCalledWith.js
@@ -44,6 +44,11 @@ getJasmineRequireObj().toHaveBeenCalledWith = function(j$) {
         }
 
         if (matchersUtil.contains(actual.calls.allArgs(), expectedArgs)) {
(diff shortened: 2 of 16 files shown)


src/core/matchers/toHaveBeenCalledWith.js
```diff
@@ -44,6 +44,11 @@ getJasmineRequireObj().toHaveBeenCalledWith = function(j$) {
         }
 
         if (matchersUtil.contains(actual.calls.allArgs(), expectedArgs)) {
+          actual.calls
+            .all()
+            .filter(call => matchersUtil.equals(call.args, expectedArgs))
```
Eradev, line 49: Any ideas why I can't use matchersUtil.equals here?

sgravrock: Each of the failing specs uses a mock `matchersUtil` that doesn't define `equals`. You'll need to update the mocks.

Eradev: Can you point me where they are in the project?

src/core/matchers/toHaveNoOtherSpyInteractions.js
```diff
@@ -0,0 +1,75 @@
...
+
+        if (!hasSpy) {
+          throw new Error(
+            getErrorMsg(
+              'Expected an object with spies, but object has no spies.'
+            )
+          );
+        }
+
+        result.message = result.pass
```
sgravrock, line 59: These messages don't match the style of the other built-in matchers, and the message for when there are other spy interactions doesn't necessarily provide enough information for the user to debug the problem. 

For the non-negated case, I'd rather see something like "Expected <name of spy object> to have no other spy interactions, but it had <reasonably well formatted summary of the calls that weren't matched, including the spy name and pretty-printed arguments>'.

For the negated case, something like "Expected <name of spy object> to have other spy interactions but it didn't" should work.

Let's avoid using "verified" and "unverified" in the failure messages since the matcher names don't use those words.

Code style note: I'd rather not have long or complex conditional expressions. I realize that this one is well within the bounds of common JS style these days, but common JS style these days is pretty awful for readability. In general, I think any of the following means that the ternary operator is no longer a good choice:

* There is any other conditional logic in the statement
* Any branch of the expression is too big to fit on a single line
* Any branch of the expression is nontrivial (this is subjective but I'd argue that the third branch of this one definitely qualifies)

sgravrock (commented):
Thanks for the PR.

This looks good for the most part. But I think the failure messages need some improvement, and that's going to drive some changes to the implementation.

src/core/matchers/toHaveNoOtherSpyInteractions.js
Eradev: How about this?

src/core/CallTracker.js
```diff
@@ -125,6 +125,16 @@ getJasmineRequireObj().CallTracker = function(j$) {
     this.saveArgumentsByValue = function() {
       opts.cloneArgs = true;
     };
+
+    /**
```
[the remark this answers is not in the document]
Eradev: Removed.

PR #2055, opened 2025-04-07
refactor(Clock): Fix throttling test unit conversion bug

atscott:
* fixes the throttling timeout test which incorrectly converted from milliseconds to seconds before making an assertion expecting milliseconds.
* reduces number of timeouts from 2000 to 100, which is still more than enough to observe throttling (or lackthereof).
* Omits Node from the test since it does not throttle timeouts


sgravrock:
Thanks for the quick fix.

PR #2057, opened 2025-04-30
fix(Clock): Ensure that uninstalling the clock also stops auto tick

atscott:
The autotick feature mistakenly does not account for the clock being a singleton and the re-installation of the clock causes the auto ticking exit conditions to become true again, before it has a chance to break.

### DIFF
--- a/spec/core/ClockSpec.js
+++ b/spec/core/ClockSpec.js
@@ -699,16 +699,19 @@ describe('Clock (acceptance)', function() {
         tick: function() {},
         uninstall: function() {}
       };
+      // window setTimeout to window to make firefox happy
+      const _setTimeout =
+        typeof window !== 'undefined' ? setTimeout.bind(window) : setTimeout;
+      // passing a fake global allows us to preserve the real timing functions for use in tests
+      const _global = { setTimeout: _setTimeout, setInterval: setInterval };
       clock = new jasmineUnderTest.Clock(
-        // We use the real window for global or firefox is displeased when we try to call a real setTimeout on an object "that doesn't implement window".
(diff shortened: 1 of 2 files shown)


atscott:
@sgravrock Sorry for missing this bit. The good news is that I was able to pretty easily migrate all of the Angular Router's tests from the ZoneJS-based fake clock (`fakeAsync`) to the jasmine mock clock with `autoTick`. In doing so, I discovered this bug and also a nice perf improvement for node (https://github.com/jasmine/jasmine/pull/2058) which reduced the test execution time from ~7s to ~5s with 3100 timers being advanced by the autoTick functionality.

spec/core/ClockSpec.js
```diff
@@ -776,6 +776,26 @@ describe('Clock (acceptance)', function() {
         });
     });
 
+    it('aborts auto ticking when uninstalled, even if installed again synchonrously', async () => {
```
sgravrock, line 779: This passes in Node but times out in all browsers. I think the problem is that [the clock is set up to overwrite the real timing functions in browsers](https://github.com/jasmine/jasmine/blob/c6b3e947e9673d50d7fd570b6e5cc84b9285b4b7/spec/core/ClockSpec.js#L704), so line 791 calls the mock clock's `setTimeout` instead of the real one.

The best way to fix it is probably to pass a fake global object in all environments, and use `setTimeout.bind(window)` to make Firefox happy. If it's necessary to pass the real global object to the clock, then it should be done unconditionally.

atscott: 👍 Done

PR #2069, opened 2025-07-14
fix(clock): Avoid generating timers with IDs that conflict with native

### LINKED ISSUE #2068: DelayedFunctionScheduler uses timer IDs that can conflict with native timers
### Steps to Reproduce

Native timers that were added before the clock is installed can conflict with timers added after the clock was installed. Clearing these native timers then clears the timers that were created by the `DelayedFunctionScheduler`.

```
    const nativeIntervalId = setInterval(function() {}, 100);
    
    const scheduler = new jasmineUnderTest.DelayedFunctionScheduler(),
      fn = jasmine.createSpy('fn');

    // Don't know the exact ID of the native one, but is 3-5 in firefox, safari, and chrome for this test
    for(let i = 0; i < nativeIntervalId; i++) {
      scheduler.scheduleFunction(function() {}, 0, [], false, i);
    }
    clearInterval(nativeIntervalId);
    scheduler.tick(1);
    
    expect(fn).toHaveBeenCalledTimes(nativeIntervalId);
```

### Expected Behavior

The scheduled function should not conflict with the native timer and should still execute when attempting to clear with the native ID. 

### Actual Behavior

The above will run the `fn` 0 times because the native timer will clear the scheduled one.

### Possible Solution

The patched `clearTimeout` should either ignore the call to `clearTimeout` if the id is for a native timer or call the native `clearTimeout` function (this was the choice in sinon: https://github.com/sinonjs/fake-timers/pull/407)

### Context

_No response_

### jasmine-core version

5.8

_No response_

### Node.js and/or browser version

all browsers

### Operating System

n/a

atscott:
This commit attempts to ensure that the timers created by jasmine mock clock do not conflict with the native timers. This also retains pre-existing behavior whereby a native scheduled function cannot be cleared if it was created prior to the mock clock being installed (unless the mock clock is uninstalled first).

Prior to this commit, attempting to clear a native timer would result in clearing a mocked scheduled function instead, in some scenarios where the IDs conflicted.

fixes #2068

### DIFF
--- a/spec/core/DelayedFunctionSchedulerSpec.js
+++ b/spec/core/DelayedFunctionSchedulerSpec.js
@@ -86,12 +86,13 @@ describe('DelayedFunctionScheduler', function() {
   it('increments scheduled fns ids unless one is passed', function() {
     const scheduler = new jasmineUnderTest.DelayedFunctionScheduler();
 
-    expect(scheduler.scheduleFunction(function() {}, 0)).toBe(1);
-    expect(scheduler.scheduleFunction(function() {}, 0)).toBe(2);
+    const initial = scheduler.scheduleFunction(function() {}, 0);
+    expect(scheduler.scheduleFunction(function() {}, 0)).toBe(initial + 1);
+    expect(scheduler.scheduleFunction(function() {}, 0)).toBe(initial + 2);
     expect(scheduler.scheduleFunction(function() {}, 0, [], false, 123)).toBe(
       123
     );
-    expect(scheduler.scheduleFunction(function() {}, 0)).toBe(3);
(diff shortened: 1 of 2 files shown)


sgravrock:
> This also retains pre-existing behavior whereby a native scheduled function cannot be cleared if it was created prior to the mock clock being installed (unless the mock clock is uninstalled first).

Yeah, I think that's the right call. I wouldn't want to forward to the real clearTimeout by default, for fear of changing the behavior of existing specs in hard-to-debug ways. Doing it on an opt-in basis like sinon does would be reasonable if there was demand for it, but as far as I know there hasn't been.

sgravrock:
Thanks for the PR.

PR #2092, opened 2026-01-30
Catch invalid path in url parsing

HolgerJeromin:
Before this, putting a filter in `path` instead of `spec`

http://localhost:8888/?hideDisabled=true&spec=&path=subPath

fails without a visual feedback

## Types of changes

## Checklist:

### DIFF
--- a/lib/jasmine-core/jasmine-html.js
+++ b/lib/jasmine-core/jasmine-html.js
@@ -1314,7 +1314,12 @@ jasmineRequire.HtmlSpecFilterV2 = function() {
       const params = this.#getFilterParams();
 
       if (params.path) {
-        return this.#matchesPath(spec, JSON.parse(params.path));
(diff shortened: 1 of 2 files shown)


sgravrock (changes requested):
> - [x] I have read the [**CONTRIBUTING**](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md) guide.

The contributing guide asks you to write specs and revert your changes to jasmine-html.js. The change to jasmine-html.js isn't a big deal, but the lack of tests is. I and other contributors rely on the tests to tell us whether or not our changes have broken Jasmine. That only works as long as we maintain a high standard of effective test coverage. I'm generally not wiling to merge PRs that lack tests, outside of a few edge cases like docs-only changes and things that are genuinely impractical to test.

PR #2110, opened 2026-06-29
Add private wrapper function for console.error() calls

amezin:
## Description

Add a private `consoleError()` function, that simply forwards all arguments to `console.error()`, and patch it in tests instead of the real console object.

## Motivation and Context

1. This is a fix for Jasmine's own tests on a runtime where console object is read-only (GJS: https://github.com/GNOME/gjs/blob/d7f5ba748123f7d4a3bd57095ae499a990cd1dd2/modules/esm/console.js#L657). The wrapper can be monkey-patched regardless of the "real" console object.

2. Repeated ESLint "no-console" rule suppressions are not necessary anymore.

## How Has This Been Tested?

I have a working port of Jasmine 7.0.0-pre.2 to GJS in the `main` branch here: https://github.com/ddterm/jasmine. Some tests are disabled because of missing APIs (`URL`, global errors), but most of the test suite passes - both on GJS, Node, browsers. This is one of the changes that was needed to fix the tests.

This PR was also tested separately on Node v26.0.4, Firefox 152.0.3, Chromium 149.

## Types of changes

## Checklist:

### DIFF
--- a/spec/core/DeprecatorSpec.js
+++ b/spec/core/DeprecatorSpec.js
@@ -1,8 +1,7 @@
-/* eslint no-console: 0 */
 describe('Deprecator', function() {
   describe('#deprecate', function() {
     beforeEach(function() {
-      spyOn(console, 'error');
+      spyOn(privateUnderTest, 'consoleError');
     });
 
     it('logs the mesage without context when the runnable is the top suite', function() {
@@ -14,7 +13,9 @@ describe('Deprecator', function() {
         omitStackTrace: true
       });
 
-      expect(console.error).toHaveBeenCalledWith('DEPRECATION: the message');
+      expect(privateUnderTest.consoleError).toHaveBeenCalledWith(
+        'DEPRECATION: the message'
+      );
     });
 
     it('logs the message in a descendant suite', function() {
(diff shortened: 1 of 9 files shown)


sgravrock:
I'm willing to merge this but I want to set expectations. Quoting from the FAQ re unsupported platforms:

> “Unsupported” means that:
>
>  * Jasmine is not tested against that browser or Node version prior to release.
>  * Contributed fixes may be accepted, but that doesn’t mean that browser or Node version will become supported.
> * Except for contributed fixes, no effort will be made to keep Jasmine working on that browser or Node version.

In this case, that means that there's a real risk of somebody breaking GJS by adding a raw `spyOn(console, 'error')` in the future. With that in mind, do you think this is worth upstreaming? Or would you prefer to maintain it in your fork?

amezin:
> With that in mind, do you think this is worth upstreaming? Or would you prefer to maintain it in your fork?

I'll have to maintain my fork anyway, but I want to reduce the diff as much as possible - it'll make maintenance easier for me.

This particular change touches lots of lines in multiple files -> high chance of conflicts and thus will likely be a pain to rebase again and again

sgravrock:
That makes sense. Merged.

PR #2113, opened 2026-07-01
Check whether `URL` class is present again

amezin:
## Description

Restore the check for presence of `URL` class.

## Motivation and Context

Jasmine had this check before. Then it was removed in v4.0.0, commit fe0a83b "Removed support for Internet Explorer".

Unfortunately, GJS has no built-in `URL` too (even though it's SpiderMonkey under the hood).

## How Has This Been Tested?

Ran the test suite on Node v26.4.0, Firefox 152.0.3, Chromium 149.

Together with other changes from `main` branch of https://github.com/ddterm/jasmine, tested on GJS 1.88.1 and 1.80.2

## Types of changes

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/jasmine/jasmine/blob/main/.github/CONTRIBUTING.md) guide.

### DIFF
--- a/spec/core/baseSpec.js
+++ b/spec/core/baseSpec.js
@@ -69,6 +69,7 @@ describe('base helpers', function() {
 
   describe('isURL', function() {
     it('returns true when the object is a URL', function() {
+      specHelpers.requireUrls();
       expect(privateUnderTest.isURL(new URL('http://localhost/'))).toBe(true);
     });
 
--- a/spec/core/matchers/matchersUtilSpec.js
+++ b/spec/core/matchers/matchersUtilSpec.js
@@ -662,6 +662,8 @@ describe('matchersUtil', function() {
     });
 
     it('passes when comparing two identical URLs', function() {
+      specHelpers.requireUrls();
+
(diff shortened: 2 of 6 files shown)


HolgerJeromin (approved):

sgravrock:
Thanks, but I think it's better for this to live in your fork. I'm thinking not just about this one check but about the broader question of what should happen when GJS turns out to need a piece of compatibility code that was removed from jasmine-core. Given how far GJS has historically lagged behind browsers and Node, I don't think this is the last time this will happen. (It might have even happened again this morning when I removed compatibility checks for BigInt64Array and BigUint64Array.) I don't want to end up in a pattern where cruft that I removed keeps coming back because of GJS.

sgravrock:
Also note that I just pushed a new `notApplicable` function and corresponding spec result status that's intended for cases like this where the spec is permanently disabled in certain configurations. I've been meaning to do that for quite some time, because jasmine-core's use of `pending` was getting out of hand. Your PR reminded me that the window to get that change in for 7.0 was closing. Sorry about the resulting rebase conflicts, but hopefully this will make things a little more manageable for everyone.

PR #2117, opened 2026-08-18
Fix jasmine-core reset in Node

### LINKED ISSUE #2116: `jasmine-core@7.0.1`: `reset()` throws immediately in Node
### Steps to Reproduce

Start in an empty directory:

```bash
mkdir jasmine-core-reset-repro
cd jasmine-core-reset-repro
npm init -y
npm install --save-dev jasmine-core@7.0.1
```

Run:

```bash
node - <<'NODE'
const jasmineCore = require('jasmine-core');

console.log('jasmine-core:', jasmineCore.version());

const before = jasmineCore.jasmine.getEnv();
jasmineCore.reset();
const after = jasmineCore.jasmine.getEnv();

console.log('environment replaced:', before !== after);
NODE
```

### Expected Behavior

`jasmineCore.reset()` should complete without throwing and reset all Jasmine state as documented, including:

- specs and suites;
- reporters;
- configuration;
- the current `Env`;
- exported and installed interface functions such as `describe`, `it`, and `afterEach`.

After the call, `jasmineCore.jasmine.getEnv()` should return the reset environment, and the public interface should be bound to that environment.

This is the state-clearing use case discussed in [#2094](https://github.com/jasmine/jasmine/issues/2094#issuecomment-3910812351).

### Actual Behavior

`reset()` throws on its first statement:

```text
jasmine-core 7.0.1
node_modules/jasmine-core/lib/jasmine-core.js:37
  private$.currentEnv_ = null;
                       ^

TypeError: Cannot set properties of undefined (setting 'currentEnv_')
    at Object.reset (node_modules/jasmine-core/lib/jasmine-core.js:37:24)
```

The published `lib/jasmine-core.js` expects `private$`:

```js
const {
  globals,
  installGlobals,
  version,
  private$
} = require('./jasmine-core/jasmine.js');

function reset() {
  private$.currentEnv_ = null;
  const env = jasmine.getEnv({ suppressLoadErrors: true });
  rebindInterface(env);
}
```

However, `lib/jasmine-core/jasmine.js` returns only:

```js
return {
  jasmine,
  globals,
  version,
  installGlobals(dest) {
    // ...
  }
};
```

It does not return `private$`, so `private$` is `undefined` and the first line of `reset()` throws.

There appear to be two additional unresolved bindings after that first failure:

- `jasmine` is not declared or imported by `lib/jasmine-core.js`;
- `rebindInterface` is not declared, imported, or exported anywhere in the published package.

Therefore, exporting `private$` alone would not make the current implementation functional.

### Possible Solution

_No response_

### Context

Long-lived Node processes may need to execute multiple independent Jasmine runs without retaining specs, suites, reporters, configuration, or globals from previous runs.

Jasmine 7's self-booting core means repeated imports return the same active core. The public `reset()` API appears intended to support clearing that state, but its published implementation currently cannot be called.

### jasmine-core version

7.0.1

_No response_

### Node.js and/or browser version

Node.js 26.7.0

### Operating System

Arch Linux

breezeFur:
## Summary

- fix `jasmine-core.reset()` so it can access the active core state
- recreate the environment and rebind exported and installed interface functions
- add an npm package regression test for the reset behavior

Fixes #2116

## Validation

- `node -e "...jasmineCore.reset()..."` reset smoke test
- `npm test` (1874 specs, 0 failures; lint and Prettier passed)

### DIFF
--- a/lib/jasmine-core.js
+++ b/lib/jasmine-core.js
@@ -30,13 +30,12 @@ const {
   globals,
   installGlobals,
   version,
-  private$
+  reset: resetCore
 } = require('./jasmine-core/jasmine.js');
 
 function reset() {
-  private$.currentEnv_ = null;
-  const env = jasmine.getEnv({ suppressLoadErrors: true });
-  rebindInterface(env);
+  resetCore();
+  Object.assign(module.exports, globals);
 }
 
 const rootPath = path.join(__dirname, 'jasmine-core');
--- a/lib/jasmine-core/jasmine.js
+++ b/lib/jasmine-core/jasmine.js
@@ -12387,25 +12387,47 @@ getJasmineRequireObj().version = function() {
   const jasmineRequire = getJasmineRequireObj();
 
   function bootJasmine(options) {
-    const jasmine = jasmineRequire.core(jasmineRequire).jasmine;
-    const env = jasmine.getEnv(options);
-    const jasmineInterface = jasmineRequire.interface(jasmine, env);
+    const core = jasmineRequire.core(jasmineRequire);
+    const jasmine = core.jasmine;
+    const private$ = core.private;
     const globals = {
       jasmine,
-      ...jasmineInterface
+      ...jasmineRequire.interface(jasmine, jasmine.getEnv(options))
     };
+    const installedDestinations = new Set();
+
+    function installGlobals(dest) {
+      dest = dest ?? globalThis;
+      installedDestinations.add(dest);
+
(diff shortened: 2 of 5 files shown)


sgravrock:
Thanks! I appreciate the quick PR. But I went with a different env instance management approach. It's fairly common for user/library code to close over globals, e.g.:

```
const underlyingIt = it;
it = function(name, fn) {
   doSomething();
   underlyingIt(name, fn);
};
```

and I wanted to cover that case as well as the more standard usage where globals aren't captured.

PR #2119, opened 2026-09-10
Show hidden differences in truncated string comparisons

### LINKED ISSUE #1870: Long String Truncation Omits Failure Cause in Message

fzlzjerry:
## Description

Show a bounded window around the first differing content when the usual pretty-printed prefixes hide a string mismatch. This covers `toEqual`, nested string differences, and `toBe` failure messages.

The implementation uses the existing `MAX_PRETTY_PRINT_CHARS` budget, marks omitted leading content, and avoids splitting UTF-16 surrogate pairs at either window boundary. Already-informative messages, other matcher predicates, equality decisions, injected pretty-printers, custom object formatters, and `jasmineToString` hooks keep their existing behavior. Tiny budgets that cannot accommodate a useful window retain the previous formatting behavior.

## Motivation and Context

Fixes #1870.

This addresses the budget/truncation interaction and coverage concerns raised on the closed #1871 rather than introducing its separate 80-character cutoff. On the original issue's 159,106/159,107-character inputs, the difference at offset 157,114 was absent from the message; it is now visible in the bounded excerpt.

## How Has This Been Tested?

- 37 new regression/control specs. Running them against the unchanged base produces 20 diagnostic failures; the other 17 controls and all existing specs pass. All 37 pass with the fix.
- `npm test` on Node 22.23.2 and 24.18.0: 1,911 specs, zero failures, plus ESLint and Prettier. Eight existing browser-only specs are not applicable in Node.
- Complete browser suites through the repository's browser runner, using local Chrome 153.0.8010.36 and Firefox 155.0.1: 2,023 specs and zero failures in each. Chrome has four existing Node-only non-applicable specs; Firefox has those four plus the existing stack-trace-format exclusion.
- Built the distribution, reran self-tests, then restored generated files and reran the exact submission tree. The extracted `npm pack` artifact also reproduces the corrected public `expect` behavior.
- A separate fixed-seed check covers 1,200 string/budget/nesting combinations: 1,200 pass versus 212 on the base. Five in-memory fault controls are detected by the regressions, including budget, both surrogate boundaries, and repeated formatter calls.

Finding a hidden difference adds a linear common-prefix scan on the failure path. For a 64k shared prefix, local sampling measured about 0.26 ms per diagnostic versus 0.03 ms before; the old output omits the difference, so these are not equivalent-output performance results. No general speedup is claimed.

## Types of changes

## Checklist:

No new public matcher API or configuration option is introduced. Generated distribution files are excluded as requested by the contribution guide.


sgravrock:
Checking for surrogate pairs is a good idea. But I see significant problems with this PR, a notable one being that the tests still pass with important parts of the implementation commented out. I also think the PR description and your comment on the original issue were LLM-generated, and the code in this PR is likely LLM output that didn't get thoroughly reviewed before you submitted it. Because of that, I'm not going to spend the time it would take to give this a detailed review.

